PR #19, opened 2024-05-21
Evaluate channel READ/WRITE entitlements locally on stream node

clemire:
For READ and WRITE permissions on the stream node, the stream node now fetches entitlements (including xchain entitlements), evaluates them, and caches them in positive / negative cache depending on result. This includes a ban check.  All other permissions are checked via the usual route.

Note: channel and space entitlements evaluated on the stream node are now evaluated through the same flow:
1. if owner, universally pass
2. if banned,reject
3. otherwise, evaluate entitlements

Other changes:
- unified wallet linking logic between xchain and stream node


core/node/config-template.yaml
```diff
@@ -103,4 +103,6 @@ network:
 # Debug feature flags.
 disableBaseChain: <DISABLE_BASE_CHAIN>
 
+chains: '31337:http://localhost:8545,31338:http://localhost:8546,84532:https://sepolia.base.org,11155111:https://ethereum-sepolia-rpc.publicnode.com'
```
clemire, line 106: This is needed for local and CI testing to be successful

core/xchain/entitlement/check_operation.go
```diff
@@ -41,6 +41,23 @@ func evaluateCheckOperation(
 	case MOCK:
 		defer infra.StoreExecutionTimeMetrics("evaluateMockOperation", infra.CONTRACT_CALLS_CATEGORY, time.Now())
 		return evaluateMockOperation(ctx, op)
+	case CheckNONE:
+		return false, fmt.Errorf("unknown operation")
+	}
+
+	// Sanity checks
+	log := dlog.FromCtx(ctx).With("function", "evaluateCheckOperation")
```
clemire, line 49: explicit errors instead of segfaults with these checks

core/sdk/src/spaceWithEntitlements.test.ts
```diff
@@ -354,150 +249,77 @@ describe('spaceWithEntitlements', () => {
         log('Done', Date.now() - doneStart)
     })
 
-    test('userEntitlementPass - join as root, linked wallet whitelisted', async () => {
-        const createAliceAndBobStart = Date.now()
+    // This test is commented out as the membership joinSpace does not check linked wallets
+    // against the user entitlement.
+    test.skip('userEntitlementPass - join as root, linked wallet whitelisted', async () => {
```
clemire, line 254: This test case checked the wrong condition - I fixed and commented out. Turns out wallet linking is not implemented in the space contract for space joins, will need to fix this one and below.

brianathere (approved):

core/node/config-template.yaml
sergekh2: what is the reason for using external blockchains for local tests?

core/node/auth/banning.go
```diff
@@ -0,0 +1,112 @@
...
+) (Banning, error) {
+	contract, err := baseContracts.NewBanning(spaceAddress, backend)
+	if err != nil {
+		return nil, err
+	}
+
+	return &banning{
+		contract:           contract,
+		spaceAddress:       spaceAddress,
+		bannedAddressCache: NewBannedAddressCache(time.Duration(cfg.NegativeEntitlementCacheTTLSeconds) * time.Second),
```
sergekh2, line 110: This value is never set in config and is always 0.

Default value used in other parts of the code is 2 seconds.

core/node/config-template.yaml
clemire: anvil doesn't behave the same way as external chains, which has exposed us to issues in the past because xchain incompatibilities weren't caught until deploying. However I'm not sure we're testing external chains on stream node, this is just the same config as what xchain is using for local testing, which I think makes sense - we still have two node config template files at this time, maybe worth refactoring.

PR #36, opened 2024-05-22
node: Read stream related settings from RiverRegistry

bas-vk:
First part to load configuration from the river registry config facet. This change includes:

- [X] write default configuration to the config facet when local env is started
- [X] adds a script that can generated configuration settings for the River Association
- [X] decodes configuration values from the chain when loaded initially or on a configuration update instead of on demand
- [X] tries to use the on-chain `stream.replicationFactor` config value, if that fails falls back to the value loaded by viper when the node starts
- [X] add `/debug/config` endpoint that dumps an overview over all settings the node has loaded

Wanted to get some initial feedback before I migrate all stream related settings in the node over.

### DIFF
--- a/core/node/crypto/config.go
+++ b/core/node/crypto/config.go
@@ -371,9 +371,7 @@ func (occ *onChainConfiguration) GetStreamMembershipLimit(streamType byte) (int,
 func (occ *onChainConfiguration) GetUint64OnBlock(blockNumber uint64, key ChainKey) (uint64, error) {
 	setting := occ.settings.getOnBlock(key, blockNumber)
 	if setting == nil {
-		return 0, RiverError(Err_NOT_FOUND, "Missing on-chain configuration setting").
-			Tag("key", key.Name()).
-			Func("GetUint64OnBlock")
+		return uint64(key.DefaultAsInt64()), nil
 	}
 	return setting.Uint64()
 }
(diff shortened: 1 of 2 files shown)


core/node/crypto/config.go
```diff
@@ -20,15 +20,44 @@ import (
...
+	StreamRecencyConstraintsAgeSecondsConfigKey  = newChainKeyImpl("stream.recencyConstraints.ageSeconds", uint64Type)
+	StreamRecencyConstraintsGenerationsConfigKey = newChainKeyImpl("stream.recencyConstraints.generations", uint64Type)
+	// StreamReplicationFactorConfigKey is the key for how often a stream is replicated over nodes
+	StreamReplicationFactorConfigKey                = newChainKeyImpl("stream.replicationFactor", uint64Type)
+	StreamDefaultMinEventsPerSnapshotConfigKey      = newChainKeyImpl("stream.defaultMinEventsPerSnapshot", uint64Type)
+	StreamMinEventsPerSnapshotUserInboxConfigKey    = newChainKeyImpl("stream.minEventsPerSnapshot.a1", uint64Type)
+	StreamMinEventsPerSnapshotUserSettingsConfigKey = newChainKeyImpl("stream.minEventsPerSnapshot.a5", uint64Type)
+	StreamMinEventsPerSnapshotUserConfigKey         = newChainKeyImpl("stream.minEventsPerSnapshot.a8", uint64Type)
+	StreamMinEventsPerSnapshotUserDeviceConfigKey   = newChainKeyImpl("stream.minEventsPerSnapshot.ad", uint64Type)
+	StreamCacheExpirationConfigKey                  = newChainKeyImpl("stream.cacheExpiration", uint64Type)
```
sergekh2, line 34: For all time values please add unit to the name: in normal config it's set as 5s or such, so it's obvious what the value is. Here it's set as uint, so without explicit probability of config bugs is going to be high.

(Also, nanos is rather extreme unit to operate with)

sergekh2 (approved):

PR #51, opened 2024-05-23
Add river and river devnet addresses

giuseppecrj:


jterzis:
Can you also add an `omega` folder under `packages/generated/deployments` with the addresses, chainId as we have for `gamma` ? A pre-install hook that runs this (https://github.com/HereNotThere/harmony/blob/main/river/packages/generated/scripts/make-config.js)on the app during build concatenates all addresses into a single deployments.json file that is used to hydrate the environment with web3 context so without that folder the app will not be aware of any of the mainnet addresses. cc @texuf for maybe a simpler way to create this directory structure.

texuf:
> Can you also add an `omega` folder under `packages/generated/deployments` with the addresses from base / river, chainId as we have for `gamma` (https://github.com/river-build/river/tree/main/packages/generated/deployments/gamma) ? A pre-install hook that runs this (https://github.com/HereNotThere/harmony/blob/main/river/packages/generated/scripts/make-config.js)on the app during build concatenates all addresses into a single deployments.json file that is used to hydrate the environment with web3 context so without that folder the app will not be aware of any of the mainnet addresses. cc @texuf for maybe a simpler way to create this directory structure.

yes this is correct, please put the addresses in the right places. Also, looks like there are things in addresses/gamma that shouldn't be there. G did you deploy contracts recently and put them in the wrong place?

texuf:
> yes this is correct, please put the addresses in the right places. Also, looks like there are things in addresses/gamma that shouldn't be there. G did you deploy contracts recently and put them in the wrong place?

@g can you follow the convention in `scripts/deploy-contracts.sh`?

giuseppecrj:
> > Can you also add an `omega` folder under `packages/generated/deployments` with the addresses from base / river, chainId as we have for `gamma` (https://github.com/river-build/river/tree/main/packages/generated/deployments/gamma) ? A pre-install hook that runs this (https://github.com/HereNotThere/harmony/blob/main/river/packages/generated/scripts/make-config.js)on the app during build concatenates all addresses into a single deployments.json file that is used to hydrate the environment with web3 context so without that folder the app will not be aware of any of the mainnet addresses. cc @texuf for maybe a simpler way to create this directory structure.
> 
> yes this is correct, please put the addresses in the right places. Also, looks like there are things in addresses/gamma that shouldn't be there. G did you deploy contracts recently and put them in the wrong place?

@jterzis Got it, yeah was unsure as to where to put the omega ones. @texuf not sure what addresses in gamma you're talking about?

jterzis (approved):

PR #55, opened 2024-05-24
Configure xchain, stream server to have default rpc urls for all supported chains for local testing, mainnet and sepolia

clemire:
Clients should be able to override these settings via a `CHAINS` environment variable. This will allow us to locally test all chains.

### DIFF
--- a/core/node/config-template.yaml
+++ b/core/node/config-template.yaml
@@ -102,3 +102,20 @@ network:
 
 # Debug feature flags.
 disableBaseChain: <DISABLE_BASE_CHAIN>
+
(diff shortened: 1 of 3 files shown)


brianathere (approved):

salzbrenner:
for the record, i think it would be simplest to have gamma and omega support the following (or whatever agreed upon mainnet chain ids):
- 1 ethereum
- 137 polygon
- 42161 arbitrum
- 10 optimism
- 8453 base

And then additionally have gamma support 84532 base-sepolia

And each rpc url should be our own alchemy url, not public rpc

PR #69, opened 2024-05-25
Add miniblockNum and eventNum to the forEachEvent iterator

texuf:
I want to add updatedAtEventNum for some items in the snapshot so that we can do proper cache invalidation. This refactor is required to keep things consistant when dealing with events past the snapshot


brianathere (approved):
Thank you!

brianathere (approved):

PR #70, opened 2024-05-25
Add updatedAtEventNum to the channel metadata

texuf:
I created a cache key in the client that depended on the updatedAtHash in the channel metadata, but that has was updating every time we had a snapshot, causing a flicker in the ui when we refreshed all channels. This will allow us to have proper cache invalidation.

I’ve split the protobuf into two new messages, one for the event and one for the snapshot. They share the same base properties and property numbers so i’m expecting this to smothly go in without any conflicts.


core/sdk/src/streamStateView_Space.ts
```diff
@@ -17,7 +18,7 @@ import { isDefaultChannelId, streamIdAsString } from './id'
 
 export type ParsedChannelProperties = {
     isDefault: boolean
-    updatedAtHash: string
+    updatedAtEventNum: bigint
 }
```
brianathere, line 22: Nice

brianathere (approved):

brianathere (approved):

PR #80, opened 2024-05-28
add modifier to EntitlementChecker to only allow registered operator to be checkers

giuseppecrj:
Testnet Proposal: https://testnets.llama.xyz/orgs/river/base-sepolia/actions/100
Mainnet Proposal: https://app.llama.xyz/orgs/river/base/actions/34


brianathere (approved):

contracts/test/base/registry/NodeOperator.t.sol
```diff
@@ -424,14 +439,14 @@ contract NodeOperatorFacetTest is
   //   address randomOperator
   // ) public givenOperatorIsRegistered(randomOperator) {
   //   vm.expectRevert(NodeOperator__InvalidAddress.selector);
-  //   nodeOperator.addSpaceDelegation(address(0), randomOperator);
+  //   nodenodeOperator.addSpaceDelegation(address(0), randomOperator);
```
clemire, line 442: accidental changes?

clemire (changes requested):
At a v high level, this looks fine to me - but I don't know what's going on in a lot of / most of this PR, so I'm not sure I'm the best person to approve it. If you can't get someone else, let me know if you want to jump on a call.

giuseppecrj:
> At a v high level, this looks fine to me - but I don't know what's going on in a lot of / most of this PR, so I'm not sure I'm the best person to approve it. If you can't get someone else, let me know if you want to jump on a call.
> 
> request for changes for a nit.

@clemire The main component that's either missing or I'm not able to asses is whether or not we're registering operators prior to calling `registerNode` on the entitlement checker contract. If not, any calls to registerNode will fail on tests.

cc. @brianathere

brianathere (approved):

core/xchain/create_multi.sh
```diff
@@ -59,6 +59,8 @@ BASE_DIR="./run_files/${RUN_ENV}"
 
 mkdir -p "${BASE_DIR}"
 
+source ../../contracts/.env.localhost
```
clemire, line 62: @brianathere I made the following changes to get xchain to properly register itself on startup, to mimic how stream node registers when the service is being launched (using the same wallet as the operator address), but was unsuccessful.

clemire: Right now xchain isn't launching, and it's failing on the registration step. This is why the skd tests for space admission that require xchain in the loop are failing.

core/xchain/launch_multi.sh
```diff
@@ -12,7 +12,11 @@ RUN_FILES_DIR="./run_files"
 # PID file for the script, stored in the base directory
 SCRIPT_PID_FILE="${RUN_FILES_DIR}/${RUN_ENV}/launch_multi.pid"
 
+source ../../contracts/.env.localhost
+
 BASE_DIR="${RUN_FILES_DIR}/${RUN_ENV}"
+BASE_REGISTRY_ADDRESS=$(jq -r '.address' ../../packages/generated/deployments/local_single/base/addresses/baseRegistry.json)
+OPERATOR_ADDRESS=$(cast wallet addr $LOCAL_PRIVATE_KEY)
```
texuf, line 19: you can't hard code `local_single` here - you need to rewpect the RUN_ENV

texuf (changes requested):
i can look through this in more detail after 9:00

core/xchain/launch_multi.sh
texuf: just change this to 
```
BASE_REGISTRY_ADDRESS=$(jq -r '.address' ../../packages/generated/deployments/${RIVER_ENV}/base/addresses/baseRegistry.json)
```

but it doesn't seem to fix it

texuf:
> I have added 2 steps that first register a node operator and that operator than registeres the xchain nodes. Not sure if this is the best place for it because running the script a second time will now give an error because it tries to register an operator/node for the second time. For CI that isn't an issue, for development it can be annoying.

yes, can you move these calls to the create_multi.sh script and fix the hard coded local_single (change to ${RIVER_ENV})?

texuf:
> > I have added 2 steps that first register a node operator and that operator than registeres the xchain nodes. Not sure if this is the best place for it because running the script a second time will now give an error because it tries to register an operator/node for the second time. For CI that isn't an issue, for development it can be annoying.
> 
> yes, can you move these calls to the create_multi.sh script and fix the hard coded local_single (change to ${RIVER_ENV})?

Actually there are several things that need to happen for this to work, i will update this pr

texuf (approved):

PR #81, opened 2024-05-28
Proxy batch delegation

giuseppecrj:


contracts/src/tokens/river/base/delegation/MainnetDelegationBase.sol
```diff
@@ -15,6 +15,39 @@ import {MainnetDelegationStorage} from "./MainnetDelegationStorage.sol";
...
+    ds.delegatorsByOperator[operator].add(delegator);
+    ds.delegationByDelegator[delegator] = Delegation(
+      operator,
+      quantity,
+      delegator,
+      block.timestamp
+    );
+
+    // Set the claimer
+    ds.claimerByDelegator[delegator] = claimer;
```
jterzis, line 47: Could we save gas by avoiding these 2 storage writes if the claimer == currentClaimer (`ds.claimerByDelegator`) ?

contracts/src/tokens/river/base/delegation/MainnetDelegation.sol
```diff
@@ -47,6 +47,25 @@ contract MainnetDelegation is
...
+  //                           Batch Delegation
+  // =============================================================
+  function setBatchDelegation(
+    address[] calldata delegators,
+    address[] calldata delegates,
+    address[] calldata claimers,
+    uint256[] calldata quantities
+  ) external onlyCrossDomainMessenger {
+    for (uint256 i = 0; i < delegators.length; i++) {
+      _replaceDelegation(
```
jterzis, line 60: What would happen if the arrays are not of equal length ? Maybe we should check and revert prior to entering the for loop if the array sizes differ from one another.

PR #97, opened 2024-06-03
Add hook to capture local id when sending messages

ttillberg:
Current use-case for this PR is to locally identify a message whose attachments are still processing.

### DIFF
--- a/core/sdk/src/client.ts
+++ b/core/sdk/src/client.ts
@@ -1124,7 +1124,7 @@ export class Client
     async sendChannelMessage(
         streamId: string,
         payload: ChannelMessage,
-        opts: SendChannelMessageOptions = {},
+        opts?: SendChannelMessageOptions,
     ): Promise<{ eventId: string }> {
         const stream = this.stream(streamId)
         check(stream !== undefined, 'stream not found')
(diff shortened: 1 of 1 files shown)


texuf (approved):

core/sdk/src/client.ts
```diff
@@ -1119,11 +1124,12 @@ export class Client
     async sendChannelMessage(
         streamId: string,
         payload: ChannelMessage,
-        opts: { beforeSendEventHook?: Promise<void> } = {},
+        opts: SendChannelMessageOptions = {},
     ): Promise<{ eventId: string }> {
         const stream = this.stream(streamId)
         check(stream !== undefined, 'stream not found')
         const localId = stream.view.appendLocalEvent(payload, 'sending', this)
+        opts?.onLocalEventAppended?.(localId)
```
brianathere, line 1132: opts isn't optional, so the ?. after opts shouldn't be required

brianathere (approved):
Approve with comment

core/sdk/src/client.ts
ttillberg: ok, I made the options optional instead, as they should be

brianathere (approved):

PR #102, opened 2024-06-04
Only allow raising commission if operator is in standby

pfives:


contracts/src/base/registry/facets/operator/NodeOperatorFacet.sol
```diff
@@ -160,6 +160,14 @@ contract NodeOperatorFacet is INodeOperator, OwnableBase, ERC721ABase, Facet {
       revert NodeOperator__NotRegistered();
     if (rateBps > 10000) revert NodeOperator__InvalidCommissionRate();
 
+    //only allow raising the commission if operator is in standby status
+    if (
+      rateBps > ds.commissionByOperator[msg.sender] &&
+      ds.statusByOperator[msg.sender] != NodeOperatorStatus.Standby
+    ) {
+      revert NodeOperator__InvalidCommissionRate();
```
jterzis, line 168: I think we should be throwing a distinct revert error here so as not to confused with https://github.com/river-build/river/pull/102/files#diff-d592fe7f19597478874616beb354457117445758958b5ccd0a4d681bc5788033L161

jterzis (approved):
approved,comment inline

PR #105, opened 2024-06-04
core: Refactor blockchain text context for more realistic tests

bas-vk:


core/node/crypto/blockchain.go
```diff
@@ -110,15 +110,25 @@ func NewBlockchainWithClient(
...
 		InitialBlockNum: initialBlockNum,
-		ChainMonitor:    chainMonitor,
+		ChainMonitor:    monitor,
 	}
 
+	go monitor.RunWithBlockPeriod(
+		ctx,
+		client,
+		initialBlockNum,
+		10*time.Millisecond,
```
sergekh2, line 128: this is not correct - should be setting from config.

core/node/crypto/blockchain.go
```diff
@@ -110,15 +110,25 @@ func NewBlockchainWithClient(
...
+		infra.NewMetrics("", ""),
```
sergekh2, line 129:
```suggestion
		metrics,
```

sergekh2 (approved):

PR #109, opened 2024-06-04
Update token struct used for test cases to use local base chain id

clemire:
These test cases broken when I moved entitlement evaluation to stream node for channels, I am updating this to use local base chain id so I can re-enable the tests in harmony.

### DIFF
--- a/core/web3/src/entitlement.ts
+++ b/core/web3/src/entitlement.ts
@@ -526,7 +526,7 @@ export function createExternalTokenStruct(addresses: Address[]) {
         return NoopRuleData
     }
     const defaultChain = addresses.map((address) => ({
-        chainId: 1n,
+        chainId: 31337n,
         address: address,
         type: CheckOperationType.ERC20 as const,
     }))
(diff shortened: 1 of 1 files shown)


brianathere (approved):

core/web3/src/entitlement.ts
```diff
@@ -526,7 +526,7 @@ export function createExternalTokenStruct(addresses: Address[]) {
         return NoopRuleData
     }
     const defaultChain = addresses.map((address) => ({
-        chainId: 1n,
+        chainId: 31337n,
```
tak-hntlabs, line 529: the path of entitlement.ts suggests that this is a public function that can be used for both test and as an SDK API. I think a better treatment is to pass an optional arg in `createExternalTokenStruct(..., opts?: TokenOptions)` where this can be set by a test to 31337.

clemire: That's a good point, Tak - these are currently only used for testing, but so is a great deal of river code and it's too much to move it all in this PR. I'll add optional overrides and refactor harmony usages.

PR #112, opened 2024-06-05
[HNT-6493] Gate key solicitations in channels on channel read permission

clemire:


core/node/rules/can_add_event.go
```diff
@@ -404,9 +405,13 @@ func (params *aeParams) canAddMemberPayload(payload *StreamEvent_MemberPayload)
 			params:      params,
 			fulfillment: content.KeyFulfillment,
 		}
-		return aeBuilder().
+		ruleBuilderAE := aeBuilder().
 			checkOneOf(params.creatorIsMember).
 			check(ru.validKeyFulfillment)
+		if _, err := params.streamView.(events.ChannelStreamView).GetChannelInception(); err == nil {
+			ruleBuilderAE = ruleBuilderAE.requireChainAuth(params.channelMessageReadEntitlements)
+		}
```
texuf, line 413: can you use the same check that we use above? 
```
		} else if shared.ValidChannelStreamId(ru.params.streamView.StreamId()) {
```

texuf: Just so we're clear... this already requires the user to be a member of the channel
- how is this check different from the membership requirement?
- how is it not redundant? especially if we get the scrubber working

clemire: 9:52 AM
I thought that last week we agreed that the stream node itself should not only check that the user is a member, but also thta they have channel read permissions, before accepting a key fulfillment request.





But now that I think about it maybe I misunderstood our conversation, and that even if a malicious node inserts a bad join stream for a channel, a non-malicious node will not accept this event into the minipool, is that right?
so if the channel stream join event went through this node's validation then we wouldn't really need to check read permissions again.

Trying to understand your perspective here. Is that where you're coming from?

texuf: no i'm remembering that we might want to proactively kick a user out of a stream if they fail permissions checks.
does this set us up for that?

clemire: Oh that's true... I wasn't thinking about implementing that here, and I think it might require some architectural changes. let's discuss?

texuf: Can you use the same check we use above?

else if shared.ValidChannelStreamId(ru.params.streamView.StreamId())

core/node/rpc/add_event.go
```diff
@@ -80,6 +80,51 @@ func (s *Service) addParsedEvent(
 	if chainAuthArgs != nil {
 		err := s.chainAuth.IsEntitled(ctx, s.config, chainAuthArgs)
 		if err != nil {
+			log := dlog.FromCtx(ctx).With("function", "addParsedEvent")
+			log.Info(
```
clemire, line 84: leaving this noisy for any potentially needed gamma debugging after shipping

core/node/default_config.yaml
```diff
@@ -22,6 +22,8 @@ baseChain:
     chainId: 31337
     networkUrl: 'http://127.0.0.1:8545'
     blockTimeMs: 2000
+    # TODO: make sure this does not get propagated to production
+    positiveEntitlementCacheTTLSeconds: 5 # Short TTL for testing
```
sergekh2, line 26: this config is used as default config in the docker image, and this will propagate it to production. 
use config-template.yaml for CI/local runs

core/node/rules/process_entitlement_loss.go
```diff
@@ -0,0 +1,60 @@
...
+func processMembershipEntitlementLoss(
+	ctx context.Context,
+	event *events.ParsedEvent,
+	payload *StreamEvent_MemberPayload,
+	streamView events.StreamView,
+) (entitlementLossEvent *PropogateEntitlementLossEvent, err error) {
+	switch payload.MemberPayload.Content.(type) {
+	case *MemberPayload_KeySolicitation_:
+		if shared.ValidChannelStreamId(streamView.StreamId()) {
+			userStreamId, err := shared.UserStreamIdFromBytes(event.Event.CreatorAddress)
```
texuf, line 44: the creator address isn't guaranteed to be the user id that got rejected.

texuf (changes requested):

core/node/default_config.yaml
brianathere: Please us yq as in the create_multi.sh for setting the correct values for testing.

core/node/rules/process_entitlement_loss.go
clemire: For a key solicitation, you can send it on behalf of someone else?

In that case, how would I derive the correct user id?

clemire: discussed in slack - key solicitation events are not created by the node, this is ok

core/node/default_config.yaml
clemire: @sergekh2 to be clear, values in config-template do NOT make it into gamma or production?

packages/sdk/src/channelsWithEntitlements.test.ts
```diff
@@ -58,6 +62,7 @@ async function setupChannelWithCustomRole(
         alice: alicesWallet.address,
         bob: bobsWallet.address,
         carol: carolsWallet.address,
+        dave: alicesWallet.address,
```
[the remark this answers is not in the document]
clemire: ah, that's kruft, I'll remove

core/node/rpc/add_event.go
```diff
@@ -78,9 +78,54 @@ func (s *Service) addParsedEvent(
 	}
 
 	if chainAuthArgs != nil {
-		err := s.chainAuth.IsEntitled(ctx, s.config, chainAuthArgs)
-		if err != nil {
-			return err
+		chainAuthErr := s.chainAuth.IsEntitled(ctx, s.config, chainAuthArgs)
+		if chainAuthErr != nil {
```
texuf, line 82: how do we know this error is an actual permission denied and not a timeout or network failure?

core/node/default_config.yaml
```diff
@@ -22,6 +22,8 @@ baseChain:
...
 riverChain:
```
texuf, line 27: is this resolved? you just said the opposite thing serge said and then resolved the convo.
also please don't commit todos in this way

core/node/rules/process_entitlement_loss.go
```diff
@@ -0,0 +1,77 @@
...
+			if !isMember {
+				log.Info("user already removed from channel", "user", event.Event.CreatorAddress, "channel", streamView.StreamId())
+				return nil, nil
+			}
+
+			userStreamId, err := shared.UserStreamIdFromBytes(event.Event.CreatorAddress)
+			if err != nil {
+				return nil, err
+			}
+			initiatorId, err := shared.AddressHex(event.Event.CreatorAddress)
```
texuf, line 61: initiator id should be the node address

texuf: eh never mind sorry

core/node/default_config.yaml
clemire: My mistake - I thought I reverted this file. This file should be unchanged and config-template.yaml changes should stand.

clemire: I think I said the same thing that Serge said, though - correct me if I'm wrong: default_config is propagated to gamma/omega and config-template is only used for local testing.

texuf: i was just confused about it being resolved and not reverted. thank you.

texuf (approved):

brianathere (commented):
Draft, right?

clemire:
> Draft, right?

No, not a draft - I'm trying to get the failing test to pass.

PR #126, opened 2024-06-07
update space implementation proxy on mainnet

giuseppecrj:
Proposals:

* Update proxy implementation to latest: https://app.llama.xyz/orgs/river/base/actions/35
* Add proxy to governance so we can diamond cuts instead of complete replacements: https://app.llama.xyz/orgs/river/base/actions/36


contracts/scripts/deployments/DeploySpace.s.sol
```diff
@@ -39,6 +39,9 @@ import {DeployMembershipReferral} from "contracts/scripts/deployments/facets/Dep
 import {DeployMultiInit} from "contracts/scripts/deployments/DeployMultiInit.s.sol";
 
 contract DeploySpace is DiamondDeployer {
+  address internal constant GOVERNANCE_ADDRESS =
+    0x63217D4c321CC02Ed306cB3843309184D347667B;
```
jterzis, line 43: Is this the same contract address on base sepolia as on base mainnet ?

giuseppecrj: That's correct, both are the same

PR #148, opened 2024-06-11
Address gas estimate too low in tests when multiple test NFT are being minted in one block by incrementing the nonce properly

brianathere:
…

### DIFF
--- a/packages/web3/src/TestGatingNFT.ts
+++ b/packages/web3/src/TestGatingNFT.ts
@@ -189,14 +189,16 @@ export async function publicMint(nftName: string, toAddress: `0x${string}`) {
     try {
         const currentNonce = await client.getTransactionCount({ address: address })
 
-        const estimatedGas = await client.estimateContractGas({
-            address: contractAddress,
-            abi: MockERC721a.abi,
-            functionName: 'mint',
-            args: [toAddress, 1n],
-            account: address,
-            nonce: currentNonce,
-        })
+        const estimatedGas = (
+            await client.estimateContractGas({
+                address: contractAddress,
+                abi: MockERC721a.abi,
+                functionName: 'mint',
+                args: [toAddress, 1n],
+                account: address,
+                nonce: currentNonce,
+            })
+        ).valueOf()
 
         nftReceipt = await client.writeContract({
             address: contractAddress,
@@ -205,7 +207,7 @@ export async function publicMint(nftName: string, toAddress: `0x${string}`) {
             args: [toAddress, 1n],
             account: address,
             nonce: currentNonce,
-            gasPrice: estimatedGas + BigInt(1e9), // 1 Gwei for cover any increases between estimation and execution
+            gasPrice: (estimatedGas + BigInt(1e9)).valueOf(), // 1 Gwei for cover any increases between estimation and execution
         })
     } finally {
         publicMintMutex.unlock()
--- a/packages/web3/src/TestGatingNFT.ts
+++ b/packages/web3/src/TestGatingNFT.ts
@@ -6,7 +6,6 @@ import MockERC721a from './MockERC721A'
 import { keccak256 } from 'viem/utils'
 
 import { dlogger } from '@river-build/dlog'
-import { log } from 'console'
 
 const logger = dlogger('csb:TestGatingNFT')
 
--- a/packages/web3/src/TestGatingNFT.ts
+++ b/packages/web3/src/TestGatingNFT.ts
@@ -3,7 +3,9 @@ import { foundry } from 'viem/chains'
 
 import MockERC721a from './MockERC721A'
 
-import { keccak256 } from 'viem/utils'
+import { keccak256, parseEther } from 'viem/utils'
+import { privateKeyToAccount, generatePrivateKey } from 'viem/accounts'
+
 import { dlogger } from '@river-build/dlog'
 
 const logger = dlogger('csb:TestGatingNFT')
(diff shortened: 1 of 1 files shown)


texuf (approved):
1) why is the test client re-using the same private key between transactions?
2) this won't prevent conflicts that are happening between node processes. I think by default the tests run 4 right?

texuf:
did this cause the 
```
    TypeError: Do not know how to serialize a BigInt
        at stringify (<anonymous>)
```
or is that something else?

brianathere:
> did this cause the
> 
> ```
>     TypeError: Do not know how to serialize a BigInt
>         at stringify (<anonymous>)
> ```
> 
> or is that something else?

I believe this change caused this, I'm now debugging locally.

brianathere:
> 1. why is the test client re-using the same private key between transactions?
> 2. this won't prevent conflicts that are happening between node processes. I think by default the tests run 4 right?

I believe the client uses the default anvil test wallets when running, so it creating a test client uses the same wallet.

Excellent point about multi node, that is still a risk, but a smaller one.

Early on I had each node process creating a set of wallets to avoid that, using worker to generate them in parallel, but I believe you removed that and we defaulted back to the test wallets. I can look at re-introducing that, or using some form of nodeJS interprocess sharing to fix this, but for now I'll accept this risk.

This change addresses the coordinated race where a single test wants to mint multiple NFT all at once to test a multi NFT scenario.

texuf:
> > 1. why is the test client re-using the same private key between transactions?
> > 2. this won't prevent conflicts that are happening between node processes. I think by default the tests run 4 right?
> 
> I believe the client uses the default anvil test wallets when running, so it creating a test client uses the same wallet.
> 
> Excellent point about multi node, that is still a risk, but a smaller one.
> 
> Early on I had each node process creating a set of wallets to avoid that, using worker to generate them in parallel, but I believe you removed that and we defaulted back to the test wallets. I can look at re-introducing that, or using some form of nodeJS interprocess sharing to fix this, but for now I'll accept this risk.
> 
> This change addresses the coordinated race where a single test wants to mint multiple NFT all at once to test a multi NFT scenario.

Instead of doing all this mutex stuff you can pass a privateKey to the "account" property of the createTextClient https://viem.sh/docs/clients/test.html#account-optional (I looked for this a month ago and couldn't find it) This convo helped.

brianathere:
> > > 1. why is the test client re-using the same private key between transactions?
> > > 2. this won't prevent conflicts that are happening between node processes. I think by default the tests run 4 right?
> > 
> > 
> > I believe the client uses the default anvil test wallets when running, so it creating a test client uses the same wallet.
> > Excellent point about multi node, that is still a risk, but a smaller one.
> > Early on I had each node process creating a set of wallets to avoid that, using worker to generate them in parallel, but I believe you removed that and we defaulted back to the test wallets. I can look at re-introducing that, or using some form of nodeJS interprocess sharing to fix this, but for now I'll accept this risk.
> > This change addresses the coordinated race where a single test wants to mint multiple NFT all at once to test a multi NFT scenario.
> 
> Instead of doing all this mutex stuff you can pass a privateKey to the "account" property of the createTestClient https://viem.sh/docs/clients/test.html#account-optional (I looked for this a month ago and couldn't find it) This convo helped.

Creating private keys is expensive, though there's probably a way to make it fast. That's why I was doing it in parallel early on, it definitely goes faster with more CPU.

texuf:
> > > > 1. why is the test client re-using the same private key between transactions?
> > > > 2. this won't prevent conflicts that are happening between node processes. I think by default the tests run 4 right?
> > > 
> > > 
> > > I believe the client uses the default anvil test wallets when running, so it creating a test client uses the same wallet.
> > > Excellent point about multi node, that is still a risk, but a smaller one.
> > > Early on I had each node process creating a set of wallets to avoid that, using worker to generate them in parallel, but I believe you removed that and we defaulted back to the test wallets. I can look at re-introducing that, or using some form of nodeJS interprocess sharing to fix this, but for now I'll accept this risk.
> > > This change addresses the coordinated race where a single test wants to mint multiple NFT all at once to test a multi NFT scenario.
> > 
> > 
> > Instead of doing all this mutex stuff you can pass a privateKey to the "account" property of the createTestClient https://viem.sh/docs/clients/test.html#account-optional (I looked for this a month ago and couldn't find it) This convo helped.
> 
> Creating private keys is expensive, though there's probably a way to make it fast. That's why I was doing it in parallel early on, it definitely goes faster with more CPU.

It’s test code that isn’t called very often and we’re already making multiple new ethers Wallet calls per user anyway, one more will not be a performance issue, and it will work across multiple processes

brianathere:
> > > > > 1. why is the test client re-using the same private key between transactions?
> 
> > > > > 2. this won't prevent conflicts that are happening between node processes. I think by default the tests run 4 right?
> 
> > > > 
> 
> > > > 
> 
> > > > I believe the client uses the default anvil test wallets when running, so it creating a test client uses the same wallet.
> 
> > > > Excellent point about multi node, that is still a risk, but a smaller one.
> 
> > > > Early on I had each node process creating a set of wallets to avoid that, using worker to generate them in parallel, but I believe you removed that and we defaulted back to the test wallets. I can look at re-introducing that, or using some form of nodeJS interprocess sharing to fix this, but for now I'll accept this risk.
> 
> > > > This change addresses the coordinated race where a single test wants to mint multiple NFT all at once to test a multi NFT scenario.
> 
> > > 
> 
> > > 
> 
> > > Instead of doing all this mutex stuff you can pass a privateKey to the "account" property of the createTestClient https://viem.sh/docs/clients/test.html#account-optional (I looked for this a month ago and couldn't find it) This convo helped.
> 
> > 
> 
> > Creating private keys is expensive, though there's probably a way to make it fast. That's why I was doing it in parallel early on, it definitely goes faster with more CPU.
> 
> 
> 
> It’s test code that isn’t called very often and we’re already making multiple new ethers Wallet calls per user anyway, one more will not be a performance issue, and it will work across multiple processes

After finding more nonce issues, I'm convinced.

It kills me though how slow our tests are

PR #166, opened 2024-06-12
Refactor IsEntitled to return bool, error pair

clemire:
Allow caller to distinguish between not entitled and evaluation errors.


texuf (approved):
nice!

PR #169, opened 2024-06-12
update deployment to allow for 4337

giuseppecrj:


contracts/test/utils/TestUtils.sol
```diff
@@ -114,7 +114,8 @@ contract TestUtils is Test {
   }
 
   function isAnvil() internal view returns (bool) {
-    return block.chainid == 31337 || block.chainid == 31338;
+    return
+      block.chainid == 1337 || block.chainid == 31337 || block.chainid == 31338;
```
pfives, line 118: would be nice to have these as constants so we know which actual chains you're looking for here

salzbrenner:
@giuseppecrj so what should i be doing? Run the stackup repo and then run deploy-contracts.sh? Cos I did that and it still has anvil specific commands that throw errors i.e. `automine`

texuf (changes requested):
these chain id changes look wrong.
when we launch anvil, we tell it what chain id will be.

salzbrenner:
i've been trying to get the devnet working outside of river/harmony w/ https://github.com/stackup-wallet/erc-4337-examples
- can spin up devnet fine
- get errors w/ the examples. i reached out to stackup team for guidance

texuf (approved):
If we run our base chain as a geth chain and or river chain as an anvil chain, I guess this will work, but I still think we should just set the geth chain id to the existing one, otherwise it breaks a lot of other code

giuseppecrj:
> If we run our base chain as a geth chain and or river chain as an anvil chain, I guess this will work, but I still think we should just set the geth chain id to the existing one, otherwise it breaks a lot of other code

The stackup docker file is confusing, I updated the chain id but it would still default to their 1337

brianathere:
> > @giuseppecrj so what should i be doing? Run the stackup repo and then run deploy-contracts.sh? Cos I did that and it still has anvil specific commands that throw errors i.e. `automine`
> 
> Yeah, the automine call doesn't exist on anvil, I also had to deploy our contracts one by one

Blocks should be produced on every transaction while in dev mode.

https://geth.ethereum.org/docs/developers/dapp-developer/dev-mode

salzbrenner:
> If we run our base chain as a geth chain and or river chain as an anvil chain, I guess this will work, but I still think we should just set the geth chain id to the existing one, otherwise it breaks a lot of other code

I tried doing this too w/out success. You can set the --networkId to whatever, but that doesn't set the chain id. To modifiy the chain id, you have to init geth with a genesis block json configuration. And I simply could not figure out how to both init with a genesis config AND run geth in dev mode. Searching online I could not conclude that this was possible or not.

texuf:
> > If we run our base chain as a geth chain and or river chain as an anvil chain, I guess this will work, but I still think we should just set the geth chain id to the existing one, otherwise it breaks a lot of other code
> 
> I tried doing this too w/out success. You can set the --networkId to whatever, but that doesn't set the chain id. To modifiy the chain id, you have to init geth with a genesis block json configuration. And I simply could not figure out how to both init with a genesis config AND run geth in dev mode. Searching online I could not conclude that this was possible or not.

We will just need to be sure that when we write the deployment files to generaged/deployments/<local_single|local_multi|...etc> that it writes the right chain id - it might just work

PR #173, opened 2024-06-14
Add stop method to RPC transport to shutdown interval logger

brianathere:
DRAFT - Another change will be needed on the harmony side to shutdown the unauthenticated client as well.

This logger keeps running after shutdown without this. At least two loggers are running in my client.


texuf (approved):

texuf:
looks good when it builds. I wish we could stop any inflight requests too.

brianathere:
> looks good when it builds. I wish we could stop any inflight requests too.

Excellent idea, I'll add that as a follow up

PR #175, opened 2024-06-14
Stress test updates

texuf:
refactors for ergonomics and usablility
updates for test output
add ci test that runs node js code


texuf:
i'll pull out the node changes

PR #184, opened 2024-06-17
Add support to subscribe to historical and future chain events

bas-vk:
This change adds support to subscribe to events at a specific block instead of only the current block the chain monitor is on. This solves the issue when the node starts, determine the current head block and orders these various components (stream cache, node registry, ...) to load the state from this block. After the initial state is loaded these components subscribe to events such as node added, stream allocated, ... to update their internal state. In the mean time the chain monitor advances to new blocks when the chain progresses. Events in intermediate blocks between the block the state is loaded from and the block the subscription is made on will be missed.

This change solves this issue by allowing event subscriptions to pass in a block number. The flow is:
1. determine `start block` at boot
2. order components to load state from this `start block`
3. components will register at the chain monitor for new events they are interested in on `start block` + 1

It adds several tests for subscribing to events on past and future blocks.

### DIFF
--- a/core/node/crypto/chain_monitor.go
+++ b/core/node/crypto/chain_monitor.go
@@ -35,7 +35,7 @@ type (
 		// OnContractEvent matches all events created by the contract on the given address.
 		OnContractEvent(addr common.Address, cb OnChainEventCallback)
 		// OnContractWithTopicsEvent matches events created by the contract on the given
-		OnContractWithTopicsEvent(addr common.Address, topics [][]common.Hash, cb OnChainEventCallback)
+		OnContractWithTopicsEvent(from BlockNumber, addr common.Address, topics [][]common.Hash, cb OnChainEventCallback)
 		// OnStopped calls cb after the chain monitor stopped monitoring the chain
 		OnStopped(cb OnChainMonitorStoppedCallback)
 	}
@@ -131,13 +131,14 @@ func (ecm *chainMonitor) OnContractEvent(addr common.Address, cb OnChainEventCal
 }
 
 func (ecm *chainMonitor) OnContractWithTopicsEvent(
+	from BlockNumber,
 	addr common.Address,
 	topics [][]common.Hash,
 	cb OnChainEventCallback,
 ) {
 	ecm.muBuilder.Lock()
 	defer ecm.muBuilder.Unlock()
-	ecm.builder.OnContractWithTopicsEvent(addr, topics, cb)
+	ecm.builder.OnContractWithTopicsEvent(from, addr, topics, cb)
 }
 
 func (ecm *chainMonitor) OnStopped(cb OnChainMonitorStoppedCallback) {
--- a/core/node/crypto/chain_monitor.go
+++ b/core/node/crypto/chain_monitor.go
@@ -35,12 +35,7 @@ type (
 		// OnContractEvent matches all events created by the contract on the given address.
 		OnContractEvent(from BlockNumber, addr common.Address, cb OnChainEventCallback)
 		// OnContractWithTopicsEvent matches events created by the contract on the given
-		OnContractWithTopicsEvent(
-			from BlockNumber,
-			addr common.Address,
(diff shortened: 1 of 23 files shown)


texuf (approved):
- can you please add a description to your pr describing 1) what this new code is doing 2) how you test this code, or alternatively, under what conditions is this code run in the wild
- there is a lot of formatting churn in this pr. it would be great if you could do that in a separate pr. Do we have a script for auto-formatting everything?  Maybe we can run that on a chron

sergekh2 (approved):

core/node/crypto/chain_monitor.go
```diff
@@ -255,19 +278,22 @@ func (ecm *chainMonitor) RunWithBlockPeriod(
...
 
+			ecm.mu.Lock()
+			fromBlock := ecm.fromBlock
+			fromBlockVersion := ecm.fromBlockVersion
+			ecm.mu.Unlock()
+
 			// ensure that the search range isn't too big because RPC providers
 			// often have limitations on the block range and/or response size.
-			if head.Number.Uint64()-fromBlock.Uint64() > 25 {
+			if head.Number.Cmp(fromBlock) > 0 && head.Number.Uint64()-fromBlock.Uint64() > 25 {
```
clemire, line 288: nit: is this condition redundant? header block # is already > fromBlock if it's at least fromBlock + 25, yes?

core/node/crypto/chain_monitor.go
```diff
@@ -314,10 +340,10 @@ func (ecm *chainMonitor) RunWithBlockPeriod(
 			}
 
 			callbacksExecuted.Wait()
-			ecm.muBuilder.Unlock()
 
 			// from and toBlocks are inclusive, start at the next block on next iteration
-			fromBlock = new(big.Int).Add(query.ToBlock, one)
+			ecm.setFromBlockVersioned(fromBlockVersion, new(big.Int).Add(query.ToBlock, one))
```
clemire, line 345: why versioned updates here? Do we expect this method to be entered by multiple threads? And also arent updates already write protected by the mutex?

core/node/crypto/chain_monitor_test.go
```diff
@@ -223,3 +224,377 @@ func TestChainMonitorEvents(t *testing.T) {
...
+	}
+
+	for i := 0; i < N; i++ {
+		tc.Commit(ctx)
+	}
+
+	receipt = <-pendingTx.Wait()
+	require.Equal(crypto.TransactionResultSuccess, receipt.Status)
+
+	// ensure that historicalContractWithTopicsEventCallback received old NodeAdded events
```
clemire, line 364: comment

core/node/crypto/chain_monitor_test.go
```diff
@@ -223,3 +224,377 @@ func TestChainMonitorEvents(t *testing.T) {
...
+	for {
+		blockNum := tc.BlockNum(ctx)
+		if blockNum.AsUint64() > lastRegisteredNodeEvent.BlockNumber {
+			break
+		}
+		time.Sleep(10 * time.Millisecond)
+	}
+
+	// register a callback for the NodeAdded event on a future block.
+	// Ensure that historicalContractAllEventsCallback and historicalContractEventsCallback receive all node added
```
clemire, line 323: comment: historicalContractAllEventsCallback, etc

clemire (changes requested):
nits, questions

core/node/crypto/chain_monitor.go
bas-vk: 2 reasons:
- in theory it is possible to subscribe on events from future blocks. In that case fromBlock > head and therefore the subtraction underflows (go doesn't panic on int over/underflows)
- head is retrieved externally from an rpc provider. If for whatever reason the rpc node reverted a chain segment head can become < fromBlock -> same problem with the subtraction underflow.

sergekh2 (approved):

clemire (approved):

PR #187, opened 2024-06-17
Run ./fmt.sh

sergekh2:


texuf:
i beat you and also i ran for xchain https://github.com/river-build/river/pull/186

PR #199, opened 2024-06-18
If setEntitlement is called a second time, updating the ruleData, first clear the existing ruleData

brianathere:


contracts/test/crosschain/RuleEntitlement.t.sol
```diff
@@ -115,6 +121,17 @@ contract RuleEntitlementTest is TestUtils, IEntitlementBase {
     assertEq(ruleOperations.length, 3);
   }
 
+  function test_setRuleEntitlementTwice() external givenRuleEntitlementIsSet {
```
jterzis, line 124: May be a good idea to also test that the previous entitlementSet was deleted along with the final cardinality of the set.

jterzis (approved):

giuseppecrj (approved):

giuseppecrj:
@brianathere please do not merge this without a proposal

PR #200, opened 2024-06-18
playground: app to use the future react sdk [HNT-7021]

miguel-nascimento:
I'm calling it `playground` instead of sample app, looks more fun :P (lmk if I should go back and use `sample-app`)

It uses Tailwind CSS as the styling option and radix components, that can be copied and pasted from https://ui.shadcn.com/ very quickly. 

- Uses the same `tsconfig` as other river packages
- `/components` has a demo of the components that we already have


packages/eslint-config/package.json
```diff
@@ -1,5 +1,5 @@
 {
-    "name": "@harmony/eslint-config",
+    "name": "@river/eslint-config",
```
texuf, line 2: Our package names should be river-build everywhere. I will make a task to rename the rest

packages/playground/package.json
```diff
@@ -0,0 +1,52 @@
+{
+    "name": "playground",
```
texuf, line 2: Can we make this @river-build/playground

packages/tsconfig.base.json
```diff
@@ -3,7 +3,7 @@
         /* Visit https://aka.ms/tsconfig to read more about this file */
 
         /* Projects */
-        "incremental": false /* Save .tsbuildinfo files to allow for incremental compilation of projects. */,
+        "incremental": true /* Save .tsbuildinfo files to allow for incremental compilation of projects. */,
```
texuf, line 6: Instead of changing this in the base tsconfig can you set it in the playground one?

packages/prettier-config/package.json
```diff
@@ -1,6 +1,6 @@
 {
-    "name": "@harmony/prettier-config",
-    "version": "workspace:^",
+    "name": "@river/prettier-config",
+    "version": "1.0.0",
```
texuf, line 3: @River-build

packages/tsconfig.base.json
miguel-nascimento: sure!

packages/eslint-config/package.json
miguel-nascimento: okay, I renamed the stuff that I changed from @river to @river-build

miguel-nascimento:
huh now I dont think its my fault the CI

texuf (approved):

PR #206, opened 2024-06-19
Hoist up fmt.sh script

sergekh2:


sergekh2:
(didn't merge main for stacked PRs, so diff is incorrect)

PR #208, opened 2024-06-19
Unite stream node and xchain node into the single binary

sergekh2:


core/archive_node.supervisord.conf
```diff
@@ -7,7 +7,7 @@ logfile=/dev/stdout
 logfile_maxbytes=0
 
 [program:archive_node]
-command=/usr/bin/stream_node archive
+command=/usr/bin/river_node archive
```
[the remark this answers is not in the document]
sergekh2: we neither have code, build, nor run chain nodes ourselves though, I don't see how it can be confusing. this name better reflects multi-functional aspect of the binary.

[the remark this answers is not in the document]
texuf: We should pick a name and stick with it. While I agree that we don't have any code for the chain nodes, everyone that we talk to, all of our documentation, and everyone that will join in the future, will see "river node" and think "block chain". I think we should call it the stream node - just in this pr you have run_stream_cmd, run_xchain_cmd, but you've got `river_node` and `xchain` folders... these should match up right?

core/river_node/main.go
bas-vk, line 1: Can we move `core/cmd` and `core/config` into the `river_node` because these are not real top level packages.

And I would suggest to rename `core/node` to `core/stream`.

core/cmd/run_stream_cmd.go
```diff
@@ -99,7 +98,7 @@ func runMetricsAndProfiler(ctx context.Context, cfg *config.Config) error {
 	return nil
 }
 
-func runServer(cfg *config.Config) error {
+func runStreamMode(cfg *config.Config) error {
```
bas-vk, line 101: Personally I would like to see the `cmd` package only deal with registering (sub)commands and arguments, loading configuration on start and call the appropriate function in the `xchain` or `node` package. That means this function should be moved to the `node` package and the caller should only have to do `stream.RunNode(cfg)` (if we rename the `node` package to `stream`)

core/cmd/run_xchain_cmd.go
```diff
@@ -8,11 +8,9 @@ import (
 	"syscall"
 
 	"github.com/river-build/river/core/xchain/server"
-
-	"github.com/spf13/cobra"
 )
 
-func run() error {
+func runXChain() error {
```
bas-vk, line 13: Same as for `runStreamMode` I would like to see this function being moved to the `xchain` package and that the run xchain command directly calls `xchain.Run(cmdConfig)`.

core/river_node/main.go
sergekh2: lets get together and figure out what module/package structure we want it to be

core/cmd/run_stream_cmd.go
sergekh2: there is some extra setup in this function before calling Execute, it needs to be futher refactored, in this PR I'm moving files around mostly, so I don't want to introduce too drastic change.

core/cmd/run_xchain_cmd.go
sergekh2: same as above

PR #214, opened 2024-06-19
Update issue templates

jterzis:


.github/ISSUE_TEMPLATE/ask-a-question.md
```diff
@@ -0,0 +1,9 @@
+---
+name: Ask a question
+about: Something is not clear
+title: ""
+labels: ""
+assignees: ""
+---
+
+This should only be used in very rare cases e.g. if you are not 100% sure if something is a bug or asking a question that leads to improving the documentation. For governance related questions use https://gov.river.build/ forum.
```
tak-hntlabs, line 9: this comment makes me wonder if we should even have this template. Can we add this later if forum isn't sufficient and there is a strong signal from the community to create a ticket to "ask a question".

jterzis: Yeah i thought it might not be needed but the governance forum is more focused on gov related questions rather than code, technical questions which I do believe should be colocated with the github repo in the Issues queue.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,35 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: ""
+labels: ""
+assignees: ""
```
tak-hntlabs, line 6: Should the md file call out "required" vs "optional" values?

tak-hntlabs: what's the intent for a bug filer to set 'assignees'. Is it to get the assignees' attention?

.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -0,0 +1,19 @@
+---
+name: Feature request
+about: Suggest an idea for this project
+title: ""
+labels: ""
+assignees: ""
+---
+
+**Is your feature request related to a problem? Please describe.**
+A clear and concise description of what the problem is. Ex. I'm always frustrated when [...]
```
tak-hntlabs, line 10: i like this clarity - based on a strong feeling.

tak-hntlabs (approved):
overall - lgtm.

.github/ISSUE_TEMPLATE/ask-a-question.md
tak-hntlabs: are we going to have a dev forum on `towns`? Like how we use team-engineering for discussing technical questions, and get fast resolution from any knowledgeable dev. I feel like filing a 'ticket' to ask a dev question is inefficient and slow to get a respond. 
If this is an RFC, then it makes sense to me to have it in github.

.github/ISSUE_TEMPLATE/ask-a-question.md
```diff
@@ -0,0 +1,9 @@
+---
+name: Ask a question
+about: Something is not clear
+title: ""
+labels: ""
+assignees: ""
```
bas-vk, line 6: I believe labels is an array and assignees a list.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,35 @@
...
+
+**Smartphone (please complete the following information):**
+
+- Device: [e.g. iPhone6]
+- OS: [e.g. iOS8.1]
+- Browser [e.g. stock browser, safari]
+- Version [e.g. 22]
+
+**Additional context**
+Add any other context about the problem here.
```
bas-vk, line 35: Do we expect xchain/stream node operators also to use this form to report bugs? In that case it is convenient to explicitly ask for relevant logs.

.github/ISSUE_TEMPLATE/ask-a-question.md
jterzis: Both can be comma-delimited strings; https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms. Ill add appropriate default label, `question`, `bug`, `feature`, for each template.

.github/ISSUE_TEMPLATE/ask-a-question.md
jterzis: Good point - I believe there's a lot of overlap between RFC and feature request, but we may want to introduce an RFC template soon. We can divert questions to the dev town once we've established one. For now, my opinion is that asking a question from the same place that you are observing the codebase is simple enough bar the additional energy needed to click a button to file a question.

.github/ISSUE_TEMPLATE/bug_report.md
jterzis: yes definitely, great suggestion. added.

PR #216, opened 2024-06-19
add generateDelegate function to signerContext

wil-hnt:


packages/sdk/src/signerContext.ts
```diff
@@ -101,6 +101,24 @@ export async function makeSignerContext(
     }
 }
 
+export async function generateDelegate(
```
texuf, line 104: can you call it `makeSignerDelegate` to match the other api?

texuf (approved):
approved with one ask

PR #217, opened 2024-06-19
Remove log lines for polling interval updates that clog test outputs.

clemire:


texuf (approved):

texuf:
looks good after lint fixes

PR #218, opened 2024-06-19
react-sdk: init [HNT-7062]

miguel-nascimento:
Got some stuff from the wagmi monorepo `react` package.

They're some stuff missing, like the lack of auto import from vitest in `.test.ts` files, but _for now_ I don't think it will be a problem (just import expect, test from vitest and you're good to go)

Later we can improve the monorepo, adding the `test:build` to CI to confirm if the package matches what the current JS ecosystem expects.

### DIFF
--- a/packages/react/README.md
+++ b/packages/react/README.md
@@ -1,0 +1,25 @@
+# @river-build/react
+
+React Hooks for River SDK.
+
+# Installation
+
+in the future:
+
+```sh
+yarn add @river-build/react
+```
+
+# Usage
+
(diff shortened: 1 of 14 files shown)


packages/react/README.md
```diff
@@ -0,0 +1,25 @@
+# @river-build/react
```
texuf, line 1: sorry to be picky, but can we call it `@river-build/react-sdk`?

miguel-nascimento: sure :))

texuf (approved):

PR #219, opened 2024-06-20
Xchain support: Fetch channel read/write entitlements and evaluate locally on the client

clemire:


packages/sdk/src/util.test.ts
```diff
@@ -423,6 +428,15 @@ export async function expectUserCanJoin(
     wallet: ethers.Wallet,
 ) {
     const joinStart = Date.now()
+
+    // Check that the local evaluation of the user's entitlements for joining the space
+    // passes.
+    const entitledWallet = await spaceDapp.getEntitledWalletForJoiningSpace(spaceId, address, [
+        'http://127.0.0.1:8545',
+        'http://127.0.0.1:8546',
+    ])
```
texuf, line 437: are these our two anvils?

texuf: i see these are hard coded twice, should we be using getXchainSupportedRpcUrlsForTesting here?

texuf (approved):
one comment, looks good

packages/sdk/src/util.test.ts
clemire: Yes, looks like I missed a spot - thank you!

PR #220, opened 2024-06-20
[db] parse sslmode

drew-u410:
* The `sslmode` is hard-coded and if you attempt to use a non-`disable` value, the service will not start.
* If not specified, returns `disable` just as current functionality expects.

### DIFF
--- a/core/node/storage/pg_storage.go
+++ b/core/node/storage/pg_storage.go
@@ -1296,6 +1296,20 @@ func (s *PostgresEventStore) createSchemaTx(ctx context.Context, tx pgx.Tx) erro
 	return nil
 }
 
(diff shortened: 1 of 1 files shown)


brianathere (commented):
Can you give more context? Why would you like to disable SSL?

drew-u410:
> Can you give more context? Why would you like to disable SSL?

I do not want to disable SSL.  I am trying to enable it.

If you pass `require` (or any argument that isn't `disable`) to `sslmode` as part of `DATABASE__EXTRA` - the service will not start due to the current implementation that hard-codes `disable`.

sergekh2 (approved):

jterzis (approved):

PR #239, opened 2024-06-21
Migrates prepay facet from space factory to space

giuseppecrj:
Context:
* The current prepay facet implementation doesn't take into account the reduction of prepaid memberships once minted, we end up with inaccurate accounting mechanism. 
* An example is a town owner prepays for 2 memberships and potential members mint those memberships, I still hold the accounting of 2 memberships in our prepay storage, which can lead to exploits.
* It makes more sense to move this logic to the space itself since the proper reduction of prepaid memberships can be done within the `_issueToken` functionality of our membership minting process

Proposals:
* https://testnets.llama.xyz/orgs/river/base-sepolia/actions/126
* https://testnets.llama.xyz/orgs/river/base-sepolia/actions/125


contracts/src/spaces/facets/prepay/PrepayFacet.sol
```diff
@@ -0,0 +1,61 @@
...
+
+  function prepayMembership(
+    uint256 supply
+  ) external payable nonReentrant onlyOwner {
+    if (supply == 0) revert Prepay__InvalidSupplyAmount();
+
+    MembershipStorage.Layout storage ds = MembershipStorage.layout();
+    IPlatformRequirements platform = IPlatformRequirements(ds.spaceFactory);
+
+    uint256 cost = supply * platform.getMembershipFee();
```
pfives, line 31: why not using calculateMembershipPrepayFee function defined below? if not why keep that?

pfives (approved):

contracts/src/spaces/facets/prepay/PrepayFacet.sol
giuseppecrj: good catch! will update.

contracts/src/spaces/facets/membership/MembershipFacet.sol
```diff
@@ -147,8 +150,22 @@ contract MembershipFacet is
       (address, address)
     );
 
-    // allocate protocol and membership fees
-    uint256 membershipPrice = _getMembershipPrice(_totalSupply());
+    uint256 totalSupply = _totalSupply();
+    uint256 membershipPrice;
+
+    uint256 freeAllocation = _getMembershipFreeAllocation();
+    uint256 prepaidSupply = _getPrepaidSupply();
```
salzbrenner, line 157: can you remind me what is the difference between freeAllocation and prepaidSupply?

contracts/src/spaces/facets/prepay/PrepayBase.sol
```diff
@@ -0,0 +1,27 @@
...
+  function _addPrepay(uint256 supply) internal {
+    PrepayStorage.Layout storage ds = PrepayStorage.layout();
+    ds.supply += supply;
+    emit Prepay__Prepaid(supply);
+  }
+
+  function _reducePrepay(uint256 supply) internal {
+    PrepayStorage.Layout storage ds = PrepayStorage.layout();
+    ds.supply -= supply;
+  }
```
salzbrenner, line 22: is there any theoretical situation where supply could become < 0?

contracts/src/spaces/facets/membership/MembershipFacet.sol
giuseppecrj: free allocation - value defined by DAO of free memberships you get for a new town
prepaid supply - what a space owner pays for on behalf of their members

contracts/src/spaces/facets/prepay/PrepayBase.sol
giuseppecrj: no, solidity would revert if an underflow happens

salzbrenner (approved):
I'll do a follow up PR to fix the SpaceDapp. The types aren't broken but the implementation is

jterzis (approved):

PR #244, opened 2024-06-21
Delete the debug app

texuf:
We’ve got the playground now


miguel-nascimento (approved):
nice! we may want to merge the playground after/before this one https://github.com/river-build/river/pull/200

PR #245, opened 2024-06-21
Fix for accidental native assets can get stuck in the space contract during free allocation or prepaid positive assertions

giuseppecrj:


contracts/src/spaces/facets/membership/MembershipFacet.sol
```diff
@@ -147,15 +150,32 @@ contract MembershipFacet is
...
+    // get token id
     uint256 tokenId = _nextTokenId();
 
     if (membershipPrice > 0) {
       uint256 userValue = _getCapturedValue(transactionId);
 
       if (userValue == 0) revert Membership__InsufficientPayment();
-      if (membershipPrice > userValue) revert Membership__InsufficientPayment();
+      if (userValue != membershipPrice)
+        revert Membership__InsufficientPayment();
```
jterzis, line 178: This error is misleading when `userValue > membershipPrice`. Could we have a new error `Membership__OverPayment()` to distinguish between the two cases more clearly or refund `userValue - membershipPrice` at the end if `userValue > membershipPrice` ?

PR #248, opened 2024-06-21
Artifacts from first npm publish

texuf:
i used changeset, honestly wasn't expecting it to work, this is what fell out


texuf:
i used lerna instead

PR #249, opened 2024-06-21
react-sdk: first draft

miguel-nascimento:
Missing for now:
- actions / mutations (need spec)
- defaultOptions (can be super useful for actions, its a good place to add toasts)
- example usage in playground
- tests


packages/react-sdk/src/RiverContext.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+export const RiverContext = createContext<RiverContext | undefined>(undefined)
+
+type RiverProviderProps = {
+    signerContext: SignerContext
+    syncAgent: SyncAgent
+    children?: React.ReactNode
+}
+
+export const RiverProvider = (props: RiverProviderProps) => {
```
ttillberg, line 18: it's practical having provider and context in the same package but I've experienced annoyances with this pattern in regards to Vite & HMR causing full page reloads on changes. This won't occur unless you touch files related to the context of course but just wanted to flag this since we start fresh might as well avoid this by splitting them into two files.

core/scripts/generate_react_sdk_index.sh
```diff
@@ -0,0 +1,41 @@
...
+  done
+
+  echo "writing $dir/index.ts"
+
+  # Append the sorted export statements to index.ts
+  printf "%s\n" "${exports[@]}" >> index.ts
+
+  # Change back to the original directory to continue the loop
+  cd - || exit
+done
```
texuf, line 41: is this file different than generate_index.sh?

core/scripts/generate_sdk_index.sh
```diff
@@ -18,7 +18,7 @@ for dir in "${dirs[@]}"; do
 
   # Add a fun comment at the top
   echo "/**************************************************************************" > index.ts
-  echo " * This file can be auto generated by 🏕️ scripts/generate_sdk_index.sh 🏕️ *" >> index.ts
+  echo " * This file can be auto generated by 🏕️ scripts/core/generate_sdk_index.sh 🏕️ *" >> index.ts
```
texuf, line 21: yipes this file got lost! we should move to to the root ./scripts directory

core/scripts/generate_sdk_index.sh
```diff
@@ -5,7 +5,7 @@
 cd -P -- "$(dirname -- "${BASH_SOURCE[0]}")"
 
 # Array of directories to process
-declare -a dirs=("../sdk/src")
+declare -a dirs=("../../packages/sdk/src")
```
texuf, line 8: this is an array, we can just add more folders here.

packages/react-sdk/src/RiverContext.tsx
```diff
@@ -0,0 +1,25 @@
...
+    syncAgent: SyncAgent
+}
+
+export const RiverContext = createContext<RiverContext | undefined>(undefined)
+
+type RiverProviderProps = {
+    signerContext: SignerContext
+    syncAgent: SyncAgent
+    children?: React.ReactNode
+}
```
texuf, line 16: if we require a syncAgent here we don't need the signerContext too, the syncAgent contains the signerContext

packages/react-sdk/src/RiverContext.tsx
texuf: 1. can we call this component RiverContextProvider?
2. Theo you just mean RiverContext.ts contains the RiverContext type and `createContext`, And RiverContextProvider.ts contains the component?

ttillberg: 1, 2. Yes exactly

core/scripts/generate_sdk_index.sh
miguel-nascimento: thank you!

packages/react-sdk/src/RiverContext.tsx
miguel-nascimento: Thanks for the feedback and the context! I think I managed to address it

packages/react-sdk/src/RiverContext.tsx
miguel-nascimento: If that's the case, I think that we only need one context: `<RiverSyncContext syncAgent={syncAgent} />` 
Is there any other case that we would like to have a more granular approach?

packages/react-sdk/src/useConnected.ts
```diff
@@ -0,0 +1,7 @@
+import { useRiverSync } from './internals/useRiverSync'
+
+export const useConnection = () => {
+    const river = useRiverSync()
+    // TODO: check the case that there's a sync agent but it's disconnected by any reason
+    return { isConnected: !!river?.syncAgent }
+}
```
texuf, line 7: can we delete this for now in favor of `isConnected` on useConnectRiver - you could call it useRiverConnection if you want

packages/sdk/src/riverConfig.ts
```diff
@@ -47,14 +51,14 @@ function makeWeb3Deployment(environmentId: string): Web3Deployment {
...
-    check(isDefined(process.env.RIVER_CHAIN_RPC_URL), 'RIVER_CHAIN_RPC_URL is not defined')
-    check(isDefined(process.env.RIVER_REGISTRY_ADDRESS), 'RIVER_REGISTRY_ADDRESS is not defined')
+    check(isDefined(process?.env?.BASE_CHAIN_ID), 'BASE_CHAIN_ID is not defined')
+    check(isDefined(process?.env?.BASE_CHAIN_RPC_URL), 'BASE_CHAIN_RPC_URL is not defined')
+    check(isDefined(process?.env?.SPACE_FACTORY_ADDRESS), 'SPACE_FACTORY_ADDRESS is not defined')
+    check(isDefined(process?.env?.BASE_REGISTRY_ADDRESS), 'BASE_REGISTRY_ADDRESS is not defined')
+    check(isDefined(process?.env?.SPACE_OWNER_ADDRESS), 'SPACE_OWNER_ADDRESS is not defined')
+    check(isDefined(process?.env?.RIVER_CHAIN_ID), 'RIVER_CHAIN_ID is not defined')
+    check(isDefined(process?.env?.RIVER_CHAIN_RPC_URL), 'RIVER_CHAIN_RPC_URL is not defined')
+    check(isDefined(process?.env?.RIVER_REGISTRY_ADDRESS), 'RIVER_REGISTRY_ADDRESS is not defined')
```
texuf, line 61: can you pop this into its own commit?

texuf (approved):

packages/sdk/src/riverConfig.ts
miguel-nascimento: https://github.com/river-build/river/pull/279

PR #253, opened 2024-06-23
Delete dead code

texuf:
This function is not used


texuf:
> Did we replace this with something else? Just checking that it didn't accidentally fall off

no idea. maybe Tak knows. This is why i put it in a separate diff.

texuf:
> used in clients/web/lib/src/hooks/use-towns-client-listener.ts to let resume sync immediately when the app is brought to the foreground.



> used in clients/web/lib/src/hooks/use-towns-client-listener.ts to let resume sync immediately when the app is brought to the foreground.

you are right. sorry i missed that.

PR #262, opened 2024-06-24
Add a pause to fix flaky getstreamex forwarding test.

clemire:
I think inserting a short sleep will fix the flaky getstreamex forwarding test, but if it happens again I will disable it.

### DIFF
--- a/core/node/rpc/service_test.go
+++ b/core/node/rpc/service_test.go
@@ -970,19 +970,32 @@ func TestForwardingWithRetries(t *testing.T) {
 			require.Equal(t, streamId[:], resp.Msg.Stream.NextSyncCookie.StreamId)
 		},
 		"GetStreamEx": func(t *testing.T, ctx context.Context, client protocolconnect.StreamServiceClient, streamId StreamId) {
-			resp, err := client.GetStreamEx(ctx, connect.NewRequest(&protocol.GetStreamExRequest{
-				StreamId: streamId[:],
-			}))
-			require.NoError(t, err)
+			// Note: the GetStreamEx implementation bypasses the stream cache, which fetches miniblocks from the
+			// registry if none are yet present in the local cache. The stream creation flow returns when a quorum of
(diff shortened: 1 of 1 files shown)


core/node/rpc/service_test.go
```diff
@@ -1016,6 +1017,13 @@ func TestForwardingWithRetries(t *testing.T) {
 				require.NoError(t, err)
 			}
 
+			// Note: The GetStreamEx implementation bypasses the stream cache, which fetches miniblocks from the
+			// registry if none are present in the local cache, and hits storage directly. This means that sometimes
+			// quorum will be achieved and the createUser call will return when a node may not have received the miniblock,
+			// and we could occasionally end up with a stream response that does not contain the genesis miniblock.
+			// This is why we sleep for a bit before making the stream requests.
+			time.Sleep(2 * time.Second)
```
sergekh2, line 1025: Adding sleep in test quite often makes flake less probably, but doesn't eliminate it completely. But also it slows down test unnecessarily. 

I was under the impression that make_miniblock above should be consistent. 

If this is not the case, would using Eventually with a bit large timeout work out better? https://pkg.go.dev/github.com/stretchr/testify/require#Assertions.Eventually

clemire: Let's discuss? To make sure I understand the code.

I think the debug make miniblock method will only make a miniblock and commit it to storage on either a single node, but it won't necessarily require the miniblocks to be replicated synchronously to all nodes fora stream. It calls SetStreamLastMiniblock on a node, which emits an event... but there's no guarantee the event will be processed by all nodes for a stream before the call returns. These tests are set up with a replication factor of 5.

sergekh2: I see, thanks for clarifying!

In this case it's better to use `require.EventuallyXxx` than `time.Sleep` since `Eventually` will succeed once the condition is true (after tick delay), so test is going to complete faster. With `Sleep` test is always going to be slow.

texuf (approved):

PR #270, opened 2024-06-25
SyncAgent - wire up user memberships

texuf:
I think this is the pattern i’m going to stick with for now
The sync agent wraps the old client
The views act like useEffects, they register with the client connection and pass back an uninitialization function
This should allow me to easily move code over from the lib while getting persistence for free.
There’s still a lot of work but I think it’s possible to make a lot of progress this week.


miguel-nascimento (approved):
Nice to see the auth status here!

PR #277, opened 2024-06-25
Sort node urls before returning in RiverRegistryDapp

texuf:
this will help me do equality


miguel-nascimento (approved):
nice! But does anyone's know why we're doing `.join(',')` instead of returning an array?

PR #280, opened 2024-06-25
new view function added to get base operators

mechanical-turk:
gamma has a few token owners on the base NodeOperatorFacet, who are not on ds.operators. This is no longer possible and omega has no such token owners. this means that using the base NodeOperatorFacet token owners as a proxy to base operators is prone to creating confusion. this is why gamma is reporting a high number of base operators with status = Exiting, since their status is set to 0 by default. this PR introduces `INodeOperator.getOperators` to standardize how we obtain actual node operator addresses on base.

Next steps:
 - ✅ Gamma proposal 
- Deploy to gamma
- ✅ Update BaseRegistry.getOperators() to use the interface directly, instead of using the token owner workaround
- Update the `river_network.total_operators_on_base` metric to exclude operators of status 0 = Exiting.
- Keep fetching nodes on base for all operators, regardless of their status. This way we can be aware of nodes on base whose operators have status = Exiting for real - instead of those who appear as Exiting just because they are token owners who are not on ds.operators.
- Deploy to omega


giuseppecrj (approved):

giuseppecrj:
Proposal: https://testnets.llama.xyz/orgs/river/base-sepolia/actions/132

PR #283, opened 2024-06-25
Do not use context.Background()

sergekh2:


core/node/crypto/config.go
```diff
@@ -399,9 +399,9 @@ func (occ *onChainConfiguration) All() (*AllSettings, error) {
 	return &all, nil
 }
 
-func (ocs *onChainSettings) Remove(key chainKeyImpl, activeOnBlockNumber uint64) {
+func (ocs *onChainSettings) Remove(ctx context.Context, key chainKeyImpl, activeOnBlockNumber uint64) {
 	var (
-		log   = dlog.FromCtx(context.Background()) // lint:ignore context.Background() is fine here
+		log   = dlog.FromCtx(ctx)
 		keyID = key.ID()
```
texuf, line 405: We had a linter for this :)

PR #287, opened 2024-06-26
SyncAgent - create space, user streams, observable dock

texuf:
slowly copying code over from the stress client, leaving places for the app view logic to go

### DIFF
--- a/packages/sdk/src/observable/persistedObservable.ts
+++ b/packages/sdk/src/observable/persistedObservable.ts
@@ -59,10 +59,10 @@ export class PersistedObservable<T extends Identifiable>
             this.data.id,
             this.loadPriority,
             (data?: T) => {
-                super.value = { status: 'loaded', data: data ?? this.data }
+                super.setValue({ status: 'loaded', data: data ?? this.data })
             },
             (error: Error) => {
-                super.value = { status: 'error', data: this.data, error }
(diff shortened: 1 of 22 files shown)


packages/sdk/src/sync-agent/syncAgent.ts
```diff
@@ -33,18 +39,25 @@ export class SyncAgent {
     user: User
     //spaces: Spaces
 
+    // flattened observables - just pointers to the observable objects in the models
+    observables: {
```
miguel-nascimento, line 43: thank you! 😊

packages/sdk/src/observable/persistedObservable.ts
```diff
@@ -14,10 +14,7 @@ export type PersistedModel<T> =
     | { status: 'saving'; data: T }
     | { status: 'saved'; data: T }
 
-interface Storable {
-    tableName: string
-    load(): void
-}
+interface Storable {}
```
miguel-nascimento, line 17: do we still need it?

miguel-nascimento (approved):

packages/sdk/src/observable/persistedObservable.ts
texuf: probably not, but i don't know what to put in the decorator otherwise.

PR #288, opened 2024-06-26
fix SpaceDapp

salzbrenner:
build fix for g's PR


texuf:
when we did this for the channel metadata we supported a string or json, see `parseChannelMetadataJSON` - will this api change break existing spaces?

tak-hntlabs (approved):

PR #292, opened 2024-06-26
Only generate single contract bindings for ts

sergekh2:
+Remove versioning support from contract shims


salzbrenner:
I have never had to use multiple versions, but it does seem like this could break something in the future. @giuseppecrj @tak-hntlabs wdyt?

tak-hntlabs:
> I have never had to use multiple versions, but it does seem like this could break something in the future. @giuseppecrj @tak-hntlabs wdyt?

Serge, G and I are going to catch up tomorrow.

texuf:
> > I have never had to use multiple versions, but it does seem like this could break something in the future. @giuseppecrj @tak-hntlabs wdyt?
> 
> Serge, G and I are going to catch up tomorrow.

The idea (correct me if I'm wrong) is that we will not break the current api of our diamond contracts, we will only add new apis and maybe tweak the underlying behavior of existing apis. 

If we ever export new ABIs and our code stops compiling, this should be a red flag.

If we add new features, we should add the ability to query the current contract and get back a feature set (Brian is going to build this right? :) )

texuf (approved):

salzbrenner:
> If we ever export new ABIs and our code stops compiling, this should be a red flag.

Doesn't happen as often but still happens

sergekh2:
we discussed it in the online meeting, in agreement to pick it up

PR #293, opened 2024-06-26
Pinning and Unpinning of channel messages

texuf:


brianathere (approved):

texuf:
@clemire fyi in case someone comes around and wants better entitlements on this message type

core/node/rules/can_add_event.go
```diff
@@ -1101,6 +1121,47 @@ func (ru *aeMembershipRules) getPermissionForMembershipOp() (auth.Permission, st
...
+		return false, err
+	}
+	computedHash := crypto.RiverHash(eventBytes)
+
+	if !bytes.Equal(ru.pin.EventId, computedHash[:]) {
+		return false, RiverError(Err_INVALID_ARGUMENT, "invalid message hash")
+	}
+
+	// cast as channel view state
+	view := ru.params.streamView.(events.ChannelStreamView)
```
clemire, line 1145: any validation here on this cast?

core/node/rules/can_add_event.go
```diff
@@ -198,8 +204,22 @@ func (params *aeParams) canAddChannelPayload(payload *StreamEvent_ChannelPayload
...
+			params: params,
+			pin:    content.Pin,
+		}
+		return aeBuilder().
+			check(params.creatorIsMember).
+			check(pinRuls.validChannelPin).
+			requireChainAuth(params.channelMessageWriteEntitlements)
+	case *ChannelPayload_Unpin_:
+		return aeBuilder().
+			check(params.creatorIsMember).
```
clemire, line 221: @texuf did you want to check that the message is currently pinned? So we don't accept multiple unpin events for the same message, or for a non-existent pin

texuf: It’s probably a good idea. If you want to add it go ahead.

PR #295, opened 2024-06-27
SyncAgent - better tests, better db initialization

texuf:
cleaning up the chaos before it spreads


miguel-nascimento (approved):
nice job, tests are way more readable now

PR #305, opened 2024-06-27
add BASE_EXECUTION_CLIENT and PATH_TO_FUND_MULTI vars

salzbrenner:
This allows for passing in some needed vars for customizing the build process from harmony so that everything works running a geth node alongside other EIP-4337 infra

### DIFF
--- a/scripts/deploy-contracts.sh
+++ b/scripts/deploy-contracts.sh
@@ -8,6 +8,7 @@ export BASE_CHAIN_ID="${BASE_CHAIN_ID:-31337}"
 export RIVER_CHAIN_ID="${RIVER_CHAIN_ID:-31338}"
 
 SKIP_CHAIN_WAIT="${SKIP_CHAIN_WAIT:-false}"
+BASE_EXECUTION_CLIENT="${BASE_EXECUTION_CLIENT:-''}"
 BASE_ANVIL_SOURCE_DIR=${BASE_ANVIL_SOURCE_DIR:-"base_anvil"}
 RIVER_ANVIL_SOURCE_DIR=${RIVER_ANVIL_SOURCE_DIR:-"river_anvil"}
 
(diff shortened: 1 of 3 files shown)


scripts/deploy-contracts.sh
```diff
@@ -39,7 +40,10 @@ fi
 # make deploy-base-anvil type=contract contract=DeployAccountFactory
 RIVER_BLOCK_TIME="${RIVER_BLOCK_TIME:-1}"
 
-cast rpc evm_setAutomine true --rpc-url $BASE_ANVIL_RPC_URL
+# Alternative nodes i.e. geth
+if [ "${BASE_EXECUTION_CLIENT}" != "geth_dev" ]; then
```
tak-hntlabs, line 44: What does `geth_dev` mean? I assume it means local development. But then it's sending rpc to anvil if not equal. shouldn't it be:
`"${BASE_EXECUTION_CLIENT}" == "geth_dev"` ?

scripts/deploy-contracts.sh
```diff
@@ -59,7 +63,9 @@ make deploy-base-anvil type=contract contract=DeployCustomEntitlementExample
 # River Registry
 make deploy-river-anvil type=contract contract=DeployRiverRegistry
 
-cast rpc evm_setIntervalMining $RIVER_BLOCK_TIME --rpc-url $BASE_ANVIL_RPC_URL
+if [ "${BASE_EXECUTION_CLIENT}" != "geth_dev" ]; then
```
tak-hntlabs, line 66: same question.

brianathere (approved):

clemire (approved):

scripts/deploy-contracts.sh
salzbrenner: Yes, geth_dev means using a local geth node in dev mode, which does not support evm_setAutomine, so I want to skip it.
Alternatively, I could try/catch these cast commands

tak-hntlabs: I see. can you add this as a comment?

tak-hntlabs (approved):

PR #306, opened 2024-06-27
react-sdk: improve return type of useObservable

miguel-nascimento:
Previously, the `data` field of the useObservable could be of the type `PersistedModel<T>`

We dont want this, we want `data: T, status: PersistedModel<T>['status']`

This PR address this.


packages/react-sdk/src/useObservable.ts
```diff
@@ -2,31 +2,45 @@
...
+          fireImmediately?: boolean
+          onUpdate?: (data: UnwrappedData) => void
+          onError?: (error: Error) => void
+          onSaved?: (data: UnwrappedData) => void
+      }
+    : {
+          fireImmediately?: boolean
+          onUpdate?: (data: Data) => void
+          onError?: (error: Error) => void
+          onSaved?: (data: Data) => void
```
miguel-nascimento, line 16: in the future, I want to remove those functions if the Observable is not a persisted one.
You won't hit `onSaved` if you're not saving anything, the types should tell this.

packages/react-sdk/src/useObservable.ts
```diff
@@ -83,30 +97,31 @@ export function useObservable<T>(
...
-            error: status === 'error' ? value.error : undefined,
-            status,
-            isLoading: status === 'loading',
-            isError: status === 'error',
-            isSaving: status === 'saving',
-            isLoaded: status === 'loaded',
-            isSaved: status === 'saved',
-        }
-    }, [value]) satisfies ObservableReturn<T>
+    }, [value]) as ObservableValue<T>
```
miguel-nascimento, line 124: Unfortunately, I hit some issues with the TS compiler to understand that
```ts
if (isPersisted(value) {
   // means that T extends PersistedModel<T>
} else {
  // means that T dont extend PersistedModel<T>
}
```
that's why im using `as` here. Would love to remove, maybe TS 5.5 is smart enough?

packages/react-sdk/src/useObservable.ts
```diff
@@ -2,31 +2,45 @@
...
 
-export type ObservableConfig<T> = {
-    fireImmediately?: boolean
-    onUpdate?: (data: T) => void
-    onError?: (error: Error) => void
-    onSaved?: (data: T) => void
-}
-
-type ObservableReturn<T> = {
-    data: T | undefined
```
miguel-nascimento, line 13: Is there any case where the first data will be an undefined one and not of the type T? cc @texuf

miguel-nascimento (commented):
Explaining some stuff

miguel-nascimento (commented):
explaining why need to throw if sync agent is not defined

texuf (approved):

PR #311, opened 2024-06-28
EIP712 wallet-link fixes for xchain GO tests

tak-hntlabs:
@g,
If all the xchain tests passes 100%, can merge into your branch. Thanks for help with the Mock.


clemire (approved):
Looks like you remove all encryption checks from the mock wallet link?

giuseppecrj (approved):

tak-hntlabs:
> Looks like you remove all encryption checks from the mock wallet link?

yes because the WalletLink contract expects additional metadata from a separate EIP712 contract in addition to the bytes packing and signing. After discussing with G, we felt that it doesn't add any extra test coverage because the integration tests covered WalletLinking already. It's not productive to add more contract deployment + reimplementation of eip712 byte packing and signing in GO for xchain testing.

brianathere:
This change was made so the content a user has to sign in MetaMask when linking a wallet is user friendly. Otherwise it was a scary block of content that users reported confusing.

PR #315, opened 2024-06-28
Client-side entitlement caching

clemire:
Added a wrapped TTL cache (with unit tests) and using it in the SpacDapp to cache entitlements and entitlement evaluations. For local dev, the positive ttl is shortened to 5s for ease of testing.


packages/web3/src/EntitlementCache.ts
```diff
@@ -0,0 +1,61 @@
...
+            return { value: negativeCacheResult, cacheHit: true }
+        }
+
+        const positiveCacheResult = this.positiveCache.get(key)
+        if (positiveCacheResult !== undefined) {
+            return { value: positiveCacheResult, cacheHit: true }
+        }
+
+        const value = await onCacheMiss(keyable)
+        if (!!value == false) {
```
brianathere, line 53: Won't this collapse into the same ring when you want it to be an error or a minimum number?

brianathere (approved):

packages/web3/src/EntitlementCache.ts
clemire: I suppose yes, I think I misunderstood what the go code was doing and thought this was behaving the same when it was not.

salzbrenner (approved):

packages/web3/src/EntitlementCache.ts
sergekh2, line 53: Lets not use this pattern:
in JS this leads to all manner of side effects, when values evaluate to false. 
Instead, please change onCacheMiss to return either CacheResult or similar type and check boolean explicitly.

packages/web3/src/EntitlementCache.ts
```diff
@@ -0,0 +1,64 @@
+import TTLCache from '@isaacs/ttlcache'
+
+export interface Keyable {
+    toKey(): string
+}
+
+export interface CacheResult<V> {
+    value: V
+    cacheHit: boolean
+    isPositive: () => boolean
```
sergekh2, line 10: but why there is a need for the function, would a `boolean` be good enough?

sergekh2 (approved):

PR #320, opened 2024-06-30
Refactor cache tests to use common helpers. Unify ChainMonitor start.

sergekh2:
This should fix node registration race in some go tests and make them more stable.

### DIFF
--- a/core/node/crypto/blockchain.go
+++ b/core/node/crypto/blockchain.go
@@ -2,13 +2,15 @@ package crypto
 
 import (
 	"context"
+	"math/big"
+	"time"
+
 	"github.com/ethereum/go-ethereum"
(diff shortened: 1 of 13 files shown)


core/node/crypto/blockchain.go
```diff
@@ -48,6 +50,7 @@ type Blockchain struct {
 	Config          *config.ChainConfig
 	InitialBlockNum BlockNumber
 	ChainMonitor    ChainMonitor
+	Metrics         infra.MetricsFactory
```
clemire, line 53: Serge, did you ever set this value on Blockchain initialization?

sergekh2: yes, in line 122 below

PR #326, opened 2024-07-01
react-sdk: auth status, space, channels, actions

miguel-nascimento:
This PR adds

- [x] choose environment
- [x] fund wallet
- [x] log in
- [x] create space
- [x] render spaces
- [x] render space channels
- [x] render timelines
- [x] send message in channel

Also changes the implementation of `useObservable` to start using `useSyncExternalStore`


packages/react-sdk/src/useSendMessage.ts
```diff
@@ -0,0 +1,18 @@
...
+import { useCurrentChannel } from './useChannel'
+
+export const useSendMessage = () => {
+    const channel = useCurrentChannel()
+    const [isSending, setSending] = useState(false)
+    const send = useCallback(
+        async (message: string) => {
+            setSending(true)
+            return channel.sendMessage(message).finally(() => setSending(false))
+        },
```
texuf, line 13: In the same way we wrap the observables, we could probably have one hook that wraps all function calls  we will need to add error handling to everything,

packages/react-sdk/src/useSpace.ts
```diff
@@ -0,0 +1,31 @@
+'use client'
+
+import { useContext, useMemo } from 'react'
+import { useSyncAgent } from './useSyncAgent'
+import { SpaceContext } from './internals/SpaceContext'
+
+export const useSpace = (spaceId: string) => {
+    const sync = useSyncAgent()
+    const space = useMemo(() => sync.spaces.getSpace(spaceId), [sync, spaceId])
+    return space
```
texuf, line 10: This should be return useObservable(space) right?

packages/react-sdk/src/useSpaceList.ts
```diff
@@ -0,0 +1,11 @@
...
+
+import { useRiver } from './useRiver'
+
+/*
+ * Returns a list with all space ids.
+ */
+export const useSpaceList = () => {
+    const { data, ...rest } = useRiver((s) => s.spaces)
+    return { spaceIds: data.spaceIds, ...rest }
+}
```
texuf, line 11: This is confusing to me. I don’t like the name, and it looks like it’s just returning the same object

packages/react-sdk/src/useRiverAuthStatus.ts
```diff
@@ -0,0 +1,18 @@
...
+    const { data: status } = useRiver((s) => s.riverAuthStatus)
+    return {
+        status,
+        isInitializing: status === AuthStatus.Initializing,
+        isEvaluatingCredentials: status === AuthStatus.EvaluatingCredentials,
+        isCredentialed: status === AuthStatus.Credentialed,
+        isConnectingToRiver: status === AuthStatus.ConnectingToRiver,
+        isConnectedToRiver: status === AuthStatus.ConnectedToRiver,
+        isDisconnected: status === AuthStatus.Disconnected,
+        isError: status === AuthStatus.Error,
```
texuf, line 16: What do these booleans really add? Why wouldn’t the end user use the enum?

packages/react-sdk/src/useChannel.ts
```diff
@@ -0,0 +1,38 @@
...
+import { useCurrentSpaceId, useSpace } from './useSpace'
+import { ChannelContext } from './internals/ChannelContext'
+
+/**
+ *  Views a channel by its channelId and spaceId.
+ */
+export const useChannel = (channelId: string, spaceId: string) => {
+    const space = useSpace(spaceId)
+    const channel = useMemo(() => space.getChannel(channelId), [space, channelId])
+    return channel
```
texuf, line 13: This should be return useObservable(channel) right?

texuf (commented):
Looking good!!!

packages/react-sdk/src/useSendMessage.ts
miguel-nascimento: 100% - nice catch!

packages/react-sdk/src/useSpace.ts
miguel-nascimento: Yeah now I'm seeing this piece of code with different eyes -- This should wrap the space in a useObservable, any function like `createChannel` and so on will be done using a different hook, like `useCreateChannel`/ `useRiverAction`

packages/react-sdk/src/useRiverAuthStatus.ts
miguel-nascimento: I find it useful when building UI, ill start using it on playground to see if the current DX looks good

packages/react-sdk/src/useChannel.ts
miguel-nascimento: Yeah, same from the useSpace discussion: https://github.com/river-build/river/pull/326#discussion_r1662653577

packages/react-sdk/src/useSpaceList.ts
miguel-nascimento: maybe `useAllSpaces`?

packages/react-sdk/src/useAllSpaces.ts
```diff
@@ -1,11 +1,6 @@
-'use client'
-
 import { useRiver } from './useRiver'
 
-/*
- * Returns a list with all space ids.
- */
-export const useSpaceList = () => {
+export const useAllSpaces = () => {
```
texuf, line 3: Is this supposed to be the spaces that the user is joined? Then I suggest we pick one of the  following naming conventions: “useMySpaces”,  “useJoinedSpaces”, or “useUserSpaces” 

there are a bunch of things where this applies, we should pick something and stick to it. I personally like the “my” moniker, since it applies to lots of things like my settings, my channels, my device keys, my inbox, etc

packages/react-sdk/src/internals/useAction.ts
```diff
@@ -0,0 +1,54 @@
...
+    const [error, setError] = useState<Error | undefined>()
+    const [data, setData] = useState<Data | undefined>()
+
+    const sync = useSyncAgent()
+
+    const action = useCallback(
+        async (...args: MultipleParams<Params>) => {
+            setStatus('loading')
+            try {
+                const data = await fn(sync)(...args)
```
miguel-nascimento, line 26: The idea of a wrapper hook is a good one, but the implementation is tricky 😬

When we pass 
```ts
useAction(sync => sync.spaces.createSpace) // <-- the action from sync agent
```

`createSpace` is a _method_ on the Spaces **class**. Which means this is using `this` everywhere.

When we do 

```ts
block: {
    await fn(sync)(...args)
    //  javascript`this` here is "block", not the originally bind class anymore.
}
```

How we fix? 
Patching the context. Inject the right `this` via [`fn.call`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/call) / [`fn.apply`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/apply)

We need a way to rebuild the `sync.spaces.createSpace` path from `createSpace` method, so we can patch it with the right `this` context (in this example, `sync.spaces`)

This could be achieved by searching all the keys in the sync class, but due to circular references this isn't ideal.

I'll pause the `useAction` internal wrapper hook for now, unless anyone come with a good solution 


this sucks and im mad at it

miguel-nascimento (commented):
javascript was a bad idea

packages/react-sdk/src/useSpaceList.ts
texuf: Is this supposed to be the spaces that the user is joined? Then I suggest we pick one of the following naming conventions: “useMySpaces”, “useJoinedSpaces”, or “useUserSpaces”

there are a bunch of things where this applies, we should pick something and stick to it. I personally like the “my” moniker, since it applies to lots of things like my settings, my channels, my device keys, my inbox, etc

packages/react-sdk/src/useAllSpaces.ts
miguel-nascimento: I like the `useMySomething`, although I think the ecosystem is pending towards the `useUserSomething` -- will do a quick research, thanks for your feedback!!

packages/react-sdk/src/internals/useAction.ts
texuf: Check out useWithCatch in harmony repo, I did a similar thing https://github.com/HereNotThere/harmony/blob/4f18f9865a4cfb4fce08f5de487924205fc17385/clients/web/lib/src/hooks/use-towns-client.ts#L368

miguel-nascimento: In that case, you have a flatten structure: `<class>.<set-action>`, so it's safe to just apply the class as `this`. Here we have `<class>.<namespace>.<set-action>`.

Now we need to know the namespace of the action to provide the correct `this` context, but the useAction hook only knows `<class>` and `<set-action>`. It's missing information to recreate the full path

miguel-nascimento: Aha, since it's an internal hook only, I think a possible solution would be to pass the namespace and then the action
```ts
useAction(sync.channel, "createChannel")
// first you pass the namespace, the second one will list all the possible actions in this namespace
```
This way we have the right context for the action

packages/playground/package.json
```diff
@@ -28,8 +29,8 @@
         "react-router-dom": "^6.4.2",
         "tailwind-merge": "^2.3.0",
         "tailwindcss-animate": "^1.0.7",
-        "viem": "^1.18.2",
-        "wagmi": "^1.4.12",
+        "viem": "2.x",
+        "wagmi": "^2.10.9",
```
texuf, line 33: the latest wagmi has some breaking changes, are we able to go with two different versions?

miguel-nascimento: I would like to use our playground with the latest version of all the stuff.
Unfortunately, there are some difficulties around here.

> are we able to go with two different versions?

Since harmony has river inside, `syncpack` won't allow.

> wagmi has some breaking changes

I might be wrong here, but harmony auth library already supports wagmi 2.x, and we have a few uses of standalone wagmi 1.x, so maybe it's worth (and easy) to upgrade?

My purpose with this is that:
1. I don't want our users to check some example code in the playground and see outdated code 😅
2. Ensure that the sdk works with a good DX along all the new stuff happening in the ecosystem.

miguel-nascimento:
waiting #461 & #460

packages/sdk/src/store/store.ts
```diff
@@ -139,7 +139,10 @@ export class Store {
         log('+enqueue load', tableName, id, loadPriority)
         this.checkTableName(tableName)
         check(this.transactionGroup !== undefined, 'transaction not started')
-        check(!this.isLoaded(tableName, id), `model already loaded table: ${tableName} id: ${id}`)
+        if (this.isLoaded(tableName, id)) {
+            log(`model already loaded table: ${tableName} id: ${id}`)
+            return
+        }
```
texuf, line 145: can you revert this change? I know we need to do something for hot reloading, but i'd like to figure out what exactly is going on, because this breaks my model of how things should work.

texuf (approved):
approved with one request

PR #334, opened 2024-07-02
created a helper function to get all members of a space via ownerOf

mechanical-turk:
Created a helper function to get all member addresess of a given space. See comments.

### DIFF
--- a/packages/web3/src/v3/Space.ts
+++ b/packages/web3/src/v3/Space.ts
@@ -439,4 +439,22 @@ export class Space {
             ruleData: entitlementDetails.ruleData,
         }
     }
+
+    /**
+     * This function is potentially expensive and should be used with caution.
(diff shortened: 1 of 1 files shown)


brianathere (approved):

packages/web3/src/v3/Space.ts
```diff
@@ -439,4 +439,22 @@ export class Space {
...
+        if (untilTokenId === undefined) {
+            untilTokenId = await this.Membership.read.totalSupply()
+        }
+
+        untilTokenId = Number(untilTokenId)
+
+        const tokenIds = Array.from({ length: untilTokenId }, (_, i) => i)
+        const promises = tokenIds.map((tokenId) => this.Membership.read.ownerOf(tokenId))
+        return Promise.all(promises)
+    }
```
texuf, line 459: I guess the main issue is that you’ll get throttled by infura. We should add some pagination to the contract.

PR #335, opened 2024-07-02
Kerem/metrics discovery

mechanical-turk:
This is a dockerized workflow that combines:

- a river node discovery script
- a prometheus server that dynamically scrapes river nodes using said script
- a datadog agent that collects metrics from the prometheus server and sends them to datadog

See README.

Next steps: i ran this on my local machine for a while and observed `omega` metrics showing up on datadog. Will go over the data with @sergekh2 tomorrow. If everything looks good, we'll move on to deploying this new infrastructure to production.


packages/generated/scripts/make-config.sh
```diff
@@ -1,6 +1,8 @@
-#!/bin/bash
+#!/bin/sh
 set -euo pipefail
-cd -P -- "$(dirname -- "${BASH_SOURCE[0]}")"
+
+cd "$(dirname "$0")"
+
 cd ..
 
 node ./scripts/make-config.js
```
mechanical-turk, line 8: @sergekh2 i needed to add these changes to make this script work on alpine. please let me know if i'm missing anything here

packages/metrics-discovery/datadog/prometheus.d/conf.yaml
```diff
@@ -0,0 +1,10 @@
+init_config: {}
+
+instances:
+  - prometheus_url: "http://prometheus:9090/federate?match%5B%5D=%7Bjob%3D%22river-node%22%7D"
+    max_returned_metrics: 9999
+    namespace: "river-node"
+    metrics:
```
jterzis, line 7: it could be helpful to include "*process*" metrics too which correlate to the health of the stream node

packages/metrics-discovery/.eslintrc.json
```diff
@@ -0,0 +1,68 @@
...
+    "plugins": ["@typescript-eslint", "import"],
+    "ignorePatterns": ["dist/**", ".turbo/**", "node_modules/**", "jest.config.ts"],
+    "rules": {
+        "import/no-extraneous-dependencies": [
+            "error",
+            {
+                "packageDir": "."
+            }
+        ],
+        "require-await": "off",
```
jterzis, line 23: let's change this to "error" as we shouldn't run async functions that are not awaited

packages/metrics-discovery/src/metrics-discovery.ts
```diff
@@ -0,0 +1,53 @@
+import type { NodeStructOutput } from '@river-build/generated/dev/typings/INodeRegistry'
+import { RiverRegistry, getWeb3Deployment } from '@river-build/web3'
+import { ethers } from 'ethers'
+
+const NODE_STATUS = {
```
jterzis, line 5: Can we not read the node_status enum from abi file in web3 instead of copying it here with ellisions ?

packages/metrics-discovery/src/metrics-discovery.ts
```diff
@@ -0,0 +1,53 @@
...
+
+    public static init(config: { riverRpcURL: string; env: string }) {
+        const deployment = getWeb3Deployment(config.env)
+        const provider = new ethers.providers.JsonRpcProvider(config.riverRpcURL)
+        const riverRegistry = new RiverRegistry(deployment.river, provider)
+        return new MetricsDiscovery(riverRegistry, config.env)
+    }
+
+    public static isTargeted(node: NodeStructOutput) {
+        return node.status === NODE_STATUS.ACTIVE || node.status === NODE_STATUS.DEPARTING
```
jterzis, line 24: I think we should include Failed nodes in the set given stream replication is not yet implemented so the network failure case exists whereby an operational node -> failed and the streams it hosts are broken.  I understand that a Failed node may not be running to emit metrics or may be running in a partially failed state or may be running correctly and be mislabelled by the operator as Failed. Doesn't hurt to attempt scraping from these nodes in my view.

packages/metrics-discovery/src/metrics-discovery.ts
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    private async getTargetNodes() {
+        console.info('Getting target nodes')
+        const allNodes = await this.riverRegistry.nodeRegistry.read.getAllNodes()
```
jterzis, line 29: Just a future thought - As the number of nodes increases it may be cheaper to call `getAllNodes(nodeStatus)` if we are filtering on a subset of nodes, which we are.

packages/metrics-discovery/README.md
```diff
@@ -0,0 +1,43 @@
...
+- a river node discovery script
+- a prometheus server that dynamically scrapes river nodes using said script
+- a datadog agent that collects metrics from the prometheus server and sends them to datadog
+
+## Instructions
+
+For local development, create a .envrc file under package root and add the following:
+
+```sh
+export DD_API_KEY=
```
jterzis, line 14: need a post-step or at least I needed:

```
source .envrc
```

packages/generated/scripts/make-config.sh
```diff
@@ -1,6 +1,8 @@
-#!/bin/bash
+#!/bin/sh
```
brianathere, line 1: Not bash?

packages/metrics-discovery/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM node:lts-alpine3.20
+
+COPY . /river
+WORKDIR /river
+
+RUN yarn install
+RUN yarn build:metrics-discovery
+RUN corepack enable
```
brianathere, line 8: I thought you'd need to enable corepack before using yarn

packages/metrics-discovery/src/index.ts
```diff
@@ -0,0 +1,27 @@
...
+        const targets = await metricsDiscovery.getPrometheusTargets()
+        console.info('Writing prometheus targets...', targets)
+        fs.writeFileSync(PROMETHEUS_TARGETS_FILE, targets)
+        console.info(`Prometheus targets written to: ${PROMETHEUS_TARGETS_FILE}`)
+        console.info(`Sleeping for ${SLEEP_DURATION_MS} ms...`)
+        await sleep(SLEEP_DURATION_MS)
+    }
+}
+
+run().catch(console.error)
```
brianathere, line 27: Maybe better to use a watcher like fs.watch or chokadir:


import { promisify } from 'util';
import fs from 'fs';
import chokidar from 'chokidar';
import { MetricsDiscovery } from './metrics-discovery';
import { envVarsSchema } from './env-vars';

const writeFile = promisify(fs.writeFile);

const PROMETHEUS_TARGETS_FILE = './prometheus/etc/config/targets.json';

interface EnvVars {
    RIVER_RPC_URL: string;
    ENV: string;
}

const run = async () => {
    const envVars: EnvVars = envVarsSchema.parse(process.env);
    const metricsDiscovery = MetricsDiscovery.init({
        riverRpcURL: envVars.RIVER_RPC_URL,
        env: envVars.ENV,
    });

    const updateTargets = async () => {
        try {
            console.info('Getting prometheus targets...');
            const targets = await metricsDiscovery.getPrometheusTargets();
            console.info('Writing prometheus targets...', targets);
            await writeFile(PROMETHEUS_TARGETS_FILE, targets);
            console.info(`Prometheus targets written to: ${PROMETHEUS_TARGETS_FILE}`);
        } catch (error) {
            console.error('Error during metrics discovery or file write:', error);
        }
    };

    await updateTargets();

    const watcher = chokidar.watch(PROMETHEUS_TARGETS_FILE);

    watcher.on('change', async () => {
        console.info(`File change detected: ${PROMETHEUS_TARGETS_FILE}`);
        await updateTargets();
    });

    console.info('Watching for changes...');
};

run().catch(error => {
    console.error('Unhandled error in run function:', error);
    process.exit(1);
});

brianathere (approved):

packages/metrics-discovery/src/metrics-discovery.ts
mechanical-turk: @jterzis that was my original intention too, but i couldn't find them. do you know where they are?

packages/metrics-discovery/src/metrics-discovery.ts
mechanical-turk: good feedback 👍 adding failed nodes

packages/metrics-discovery/src/metrics-discovery.ts
mechanical-turk: are you talking about riverRegistry.getAllNodes(status)? that one is basically the same idea as here: fetches all nodes from the contract and filters them at the client side. filtering by status doesn't exist on the contract level. will make a note to keep this in mind as the number of nodes grow though 👍

packages/generated/scripts/make-config.sh
mechanical-turk: couldn't get it to build on alpine without sh. but i think i need to rethink my workarounds here, because now CI is failing

packages/metrics-discovery/Dockerfile
mechanical-turk: i think it worked because yarn came built in with the image. and if you're curious about why i'm even running 1corepack enable1 if not to install the package managers in the first place - it's because some error message told me to :) 

i'm hoisting it up though 👍

packages/metrics-discovery/src/index.ts
mechanical-turk: @brianathere the metrics discovery script is write-only with respect to targets.json. it doesn't need to watch for anything. the only responsibility here is to fetch all nodes and write `targets.json`. prometheus then picks up the changes automatically from `targets.json` and updates its collectors.

good idea to make fs.write async though 👍

packages/generated/scripts/make-config.sh
mechanical-turk: edit: I'm deleting this whole file. seems like the only purpose it serves is to cd into `packages/generated` before running `node ./scripts/make-config.js`. the same outcome can be achieved by running the node script directly from `package.json`. i'm also updating the `scripts/deploy-contracts.sh` script to run this new package.json script

jterzis (approved):

PR #337, opened 2024-07-03
Switch on-chain settings to use struct

sergekh2:
Settings are changed very rarely, but read all the time.
Use immutable structure that can be shared makes it easier and faster to access settings.


core/node/crypto/config.go
```diff
@@ -3,624 +3,492 @@ package crypto
...
+	input := make(map[string]any)
+	for _, blockNum := range blockNums {
+		for _, v := range occ.loadedSettingMap[blockNum] {
+			name, ok := occ.keyHashToName[v.key]
+			if !ok {
+				log.Warn("Unknown setting key", "key", v.key, "value", v.value, "block", blockNum)
+				continue
+			}
+			if v.deleted {
+				input[name] = occ.defaultsMap[name]
```
bas-vk, line 229: What happens in the following scenario:

block 0: `stream.media.maxChunkCount` = 50 (default)
block 5: `stream.media.maxChunkCount` = 100
block 10: `stream.media.maxChunkCount` = 200

200 gives problems and the config change on block 10 is reverted/deleted. It should fall back to 100 but this falls back to the default value of 50?

core/node/crypto/config.go
```diff
@@ -3,624 +3,492 @@ package crypto
...
+				}
+			}
+		} else if from.Kind() == reflect.Slice && from.Type().Elem().Kind() == reflect.Uint8 {
+			if to.Kind() == reflect.Int64 || to.Kind() == reflect.Int {
+				v, err := ABIDecodeInt64(from.Bytes())
+				if err == nil {
+					return v, nil
+				}
+				log.Error("failed to decode int64", "err", err, "bytes", from.Bytes())
+			} else if to.Kind() == reflect.Uint64 {
```
bas-vk, line 474: Missing `reflect.Uint`?

core/node/crypto/config.go
```diff
@@ -3,624 +3,492 @@ package crypto
...
+			// Preprocess durations based on name suffix.
+			mapValue, ok := from.Interface().(map[string]interface{})
+			if ok {
+				var badKeys []string
+				for key, value := range mapValue {
+					ms := strings.HasSuffix(key, "Ms")
+					sec := strings.HasSuffix(key, "Seconds")
+					bb, ok := value.([]byte)
+					if (ms || sec) && ok {
+						vv, err := ABIDecodeUint64(bb)
```
bas-vk, line 448: Durations can be negative. Decoding them to int64 is more accurate.

core/node/crypto/config.go
sergekh2: great point!
I looked at the contract code, I need to change processing to match expectations there.

sergekh2: now code matches proposed contract changes.

PR #339, opened 2024-07-03
Add a way to distinguish between setting deletion events

sergekh2:
There is no way to distinguish between `deleteConfiguration` and `deleteConfigurationOnBlock` on block 0 by emitted events. This adds special value to `deleteConfiguration` event.


contracts/src/river/registry/facets/config/RiverConfig.sol
```diff
@@ -78,7 +78,7 @@ contract RiverConfig is IRiverConfig, RegistryModifiers, OwnableBase, Facet {
 
     ds.configurationKeys.remove(key);
 
-    emit ConfigurationChanged(key, 0, "", true);
+    emit ConfigurationChanged(key, 0, "ALL", true);
```
giuseppecrj, line 81: Make "ALL" an enum

bas-vk: The meaning of this field is to communicate the value to anyone that is interested in config updates. For set config events this holds the abi encoded value for the key. So we can't change it to an enum.

I'm also not really keen on using `ALL`. It means that this field either contains a abi encoded value in case of a set config, `ALL` in case all values for the key are deleted or `""` in case a setting for a key on a particular block is deleted.

Maybe we can add a new event `AllConfigurationDeleted(key)` to indicate all settings for a particular key are deleted?

sergekh2: I want to keep ABI changes to the reasonable minimum.

Another idea - I changed it to use max_uint64 in the block number.

PR #347, opened 2024-07-06
optimize river protocol contracts

giuseppecrj:


contracts/src/river/registry/facets/stream/StreamRegistry.sol
```diff
@@ -31,7 +31,8 @@ contract StreamRegistry is IStreamRegistry, RegistryModifiers {
       revert(RiverRegistryErrors.ALREADY_EXISTS);
 
     // verify that the nodes stream is placed on are in the registry
-    for (uint256 i = 0; i < nodes.length; ++i) {
+    uint256 nodeCount = nodes.length;
```
brianathere, line 34: Makes me sad that solc doesn't do invariant hoisting like a real compiler would. :-(

brianathere (approved):

pfives (approved):

PR #361, opened 2024-07-08
Update addEvent rule evaluation to emit multiple chainAuthArgs for supporting multiple permissions

clemire:
This is a prerequisite for react_reply, which requires allowing channel message writes if the user has WRITE or REACT_REPLY.


core/node/rpc/add_event.go
```diff
@@ -77,11 +78,20 @@ func (s *Service) addParsedEvent(
 		return err
 	}
 
-	if chainAuthArgs != nil {
-		isEntitled, err := s.chainAuth.IsEntitled(ctx, s.config, chainAuthArgs)
-		if err != nil {
-			return err
+	if len(chainAuthArgsList) > 0 {
+		var isEntitled bool = false
```
bas-vk, line 82: just do `isEntitled := false`

core/node/rpc/add_event.go
```diff
@@ -93,11 +103,19 @@ func (s *Service) addParsedEvent(
...
 				}
 			}
+			var chainAuthArgsListStringBuilder strings.Builder
+			chainAuthArgsListStringBuilder.WriteString("[\n")
+			for _, chainAuthArgs := range chainAuthArgsList {
+				chainAuthArgsListStringBuilder.WriteString("  ")
+				chainAuthArgsListStringBuilder.WriteString(chainAuthArgs.String())
+				chainAuthArgsListStringBuilder.WriteString(",\n")
+			}
+			chainAuthArgsListStringBuilder.WriteString("]")
```
bas-vk, line 113: Maybe there is a good reason for but this looks a bit fishy. Why not leave the formatting to the logger?

core/node/rules/rule_builder_ae.go
```diff
@@ -127,12 +136,12 @@ func (re *ruleBuilderAEImpl) run() (bool, *auth.ChainAuthArgs, *AddEventSideEffe
 	if err != nil {
 		return false, nil, nil, err
 	}
-	if len(re.checks) == 0 && chainAuthArgs == nil && requiredParentEvent == nil {
+	if len(re.checks) == 0 && chainAuthArgsList == nil && requiredParentEvent == nil {
```
bas-vk, line 139: You probably want to check on `len(chainAuthArgsList) == 0`? If not this check also passed with an empty slice.

bas-vk (commented):
I'm not familiar enough with the entitlement checks to review the functional changes so I only looked at the code.

core/node/rpc/add_event.go
clemire: You know, I never did figure out how toString works with go - I think you're right, there's already a string method on ChainAuthArgs so this is probably unnecessary.

core/node/rules/rule_builder_ae.go
clemire: Yes, you're right - thank you for catching this!

brianathere (approved):

core/node/rpc/add_event.go
```diff
@@ -77,11 +77,20 @@ func (s *Service) addParsedEvent(
 		return err
 	}
 
-	if chainAuthArgs != nil {
-		isEntitled, err := s.chainAuth.IsEntitled(ctx, s.config, chainAuthArgs)
-		if err != nil {
-			return err
+	if len(chainAuthArgsList) > 0 {
+		isEntitled := false
+		var err error
```
texuf, line 82: This looks good. Random nit - hard to see on mobile, is there a reason you predicated this var err here? Looks like you return it immediately if it gets set

texuf: *pre-declared

PR #362, opened 2024-07-08
core: add metrics for in-flight grpc requests and open streams

bas-vk:
Add prom metrics for grpc calls:

- `grpc_unary_inflight` count the number of in progress unary calls
- `grpc_open_client_streams` number of open streams where node is receiving from a remote server
- `grpc_open_server_streams` number of open streams where node is streaming to client

### DIFF
--- a/core/node/rpc/metrics_interceptor.go
+++ b/core/node/rpc/metrics_interceptor.go
@@ -3,33 +3,82 @@ package rpc
 import (
 	"context"
 
+	"connectrpc.com/connect"
 	"github.com/prometheus/client_golang/prometheus"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
-
-	"connectrpc.com/connect"
 )
 
 type streamIdProvider interface {
 	GetStreamId() string
 }
 
-func (s *Service) NewMetricsInterceptor() connect.UnaryInterceptorFunc {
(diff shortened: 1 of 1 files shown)


sergekh2 (approved):

core/node/rpc/metrics_interceptor.go
```diff
@@ -3,33 +3,82 @@ package rpc
...
+	}
+}
+
+func (i *metricsInterceptor) WrapUnary(next connect.UnaryFunc) connect.UnaryFunc {
+	return func(
+		ctx context.Context,
+		req connect.AnyRequest,
+	) (connect.AnyResponse, error) {
+		var (
+			proc = req.Spec().Procedure
```
jterzis, line 37: proc name prepends a "/" which we do not need to pass as part of the label. Maybe remove this prior to creating the label or otherwise sanitize the label before adding it.

```
# HELP river_stream_grpc_unary_inflight gRPC unary calls in flight
# TYPE river_stream_grpc_unary_inflight gauge
river_stream_grpc_unary_inflight{proc="/river.StreamService/Info"} 0
```

bas-vk: It would change old behaviour that already measures rpc duration and uses the plain procedure name with the `/` prefix.

PR #363, opened 2024-07-08
Close idle connections in debug status request handling

brianathere:


sergekh2 (approved):

core/node/rpc/multi.go
```diff
@@ -218,18 +218,22 @@ func GetRiverNetworkStatus(
 	if err != nil {
 		return nil, err
 	}
+	defer http11client.CloseIdleConnections()
```
jterzis, line 221: With DisableKeepAlives: true by default for Http11Client, this line is now redundant and can be removed in favor of the transport setting.

jterzis (approved):

bas-vk (approved):

PR #366, opened 2024-07-09
Client checks for entitlements before channel posts

clemire:
The final step of react-reply "backend" support - the client self-filters invalid messages based on whether or not it has entitlements to post to the channel.

### DIFF
--- a/packages/sdk/src/util.test.ts
+++ b/packages/sdk/src/util.test.ts
@@ -10,10 +10,11 @@ import {
     SyncStreamsResponse,
     SyncOp,
 } from '@river-build/proto'
+import { Entitlements } from './sync-agent/entitlements/entitlements'
 import { PlainMessage } from '@bufbuild/protobuf'
 import { IStreamStateView } from './streamStateView'
 import { Client } from './client'
-import { makeBaseChainConfig, makeRiverChainConfig } from './riverConfig'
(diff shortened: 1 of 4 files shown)


packages/sdk/src/util.test.ts
```diff
@@ -206,14 +206,44 @@ export async function setupWalletsAndContexts() {
     const aliceSpaceDapp = createSpaceDapp(aliceProvider, baseConfig.chainConfig)
     const carolSpaceDapp = createSpaceDapp(carolProvider, baseConfig.chainConfig)
 
+    // Use this wrapper around the space dapp to help the client correctly evaluate
+    // on-chain entitlements for channel operations.
+    class SpaceDappChannelEntitlementsDelegate {
```
texuf, line 211: I added an entitlements class here https://github.com/river-build/river/blob/d731364051a4600c1bc4157d6c0cc25a6365d203/packages/sdk/src/sync-agent/entitlements/entitlements.ts#L5

can we use that instead? I want us to flush this out with the wallet linking code that needs to get ported from harmony

clemire: No problem, TY!

brianathere (approved):

clemire:
closes hnt-7728

PR #379, opened 2024-07-10
Publish npm packages nightly

jterzis:


brianathere (approved):

.github/workflows/Publish_npm_packages.yml
```diff
@@ -0,0 +1,139 @@
...
+                  }
+
+                  function make_pr_description() {
+                      git log origin/main..HEAD
+                  }
+
+                  if [[ "$(git status --porcelain)" != "" ]]; then
+                      echo "There are uncommitted changes. Please commit or stash them before running this script."
+                      exit 1
+                  elif [[ "$(parse_git_branch)" != "main" ]]; then
```
brianathere, line 55: You shouldn't fail this, or you should set the tag above to only run this on main.

texuf: 1. Why is this pr incrementing versions and changing build scripts at the same time?
2. Why is all this scripting in the yaml? @sergekh2 didnt we talk about not putting loads is scripting in yaml? Can we move this to a .sh file?
3. We need to pass credentials to lerna publish right? How are you handling that?
4. Nice work

jterzis: 1. reverted the commit that increments versions - i was just testing the script in the gh action worfklow
2. we can move this to an .sh file - it was easier to iterate inline at first. Note that the gh action file does need to be in yaml. 
3. yes we are passing creds to npm actually here: https://github.com/river-build/river/pull/379/files#diff-b18da9d93fa6ddfc7c8c164dedbf63e3ec3c240986964eceac8014a8aac6b0caR30. If you have a better way to authenticate, please let me know.
4. 🤝 👍

texuf: Ah nice. What’s the expiry for that auth? This script takes a while to run.

texuf:
> @jterzis in case you're wondering how other projects handle automatically bumping up their package versions and publishing npm packages, i can't recommend [changesets](https://github.com/changesets/changesets) enough. it can be configured as a github action with ~5 lines of code. lot's of other well known monorepos use it, e.g svelte, mobx, wagmi, graphiql.
> 
> the only tradeoff is that when we want our PRs to trigger a version bump and a publish, we need to write a changeset. this changeset file captures what kind of version bump this change needs (i.e major vs minor vs patch), as well as a description which later becomes a part of the changelog. everything else is taken care of, including bumping versions of dependencies of dependencies and so on.

@mechanical-turk i looked at changesets. The implementation seemed both amateurish and limited, lerna seemed to do exactly what we wanted, but maybe I missed something? Maybe I was looking at the docs to run it myself and all we needed was the actions. 

My concerns
- Lerna allows you to force publish all non-public packages
- lerna lets you publish all packages to the same number
- lerna lets us split the tagging and publishing, so we can run coming between
- lerna uses got tags so we can easily sync the came commit as the release

I actually fucked up the first published version using change sets any fixed it with Lerna   

that said, lerna was obviously built to be run by hand not in ci, our hotshot contractor Miguel told me to use changesets, and now your recommendation - I will not be offended to find out it really is better! If we can publish all packages to the same number even if one doesn’t have changes, and if we can easily add new packages, or if there’s a good reason not to do that - then great. Let’s use changesets.

texuf:
> > @jterzis in case you're wondering how other projects handle automatically bumping up their package versions and publishing npm packages, i can't recommend [changesets](https://github.com/changesets/changesets) enough. it can be configured as a github action with ~5 lines of code. lot's of other well known monorepos use it, e.g svelte, mobx, wagmi, graphiql.
> > the only tradeoff is that when we want our PRs to trigger a version bump and a publish, we need to write a changeset. this changeset file captures what kind of version bump this change needs (i.e major vs minor vs patch), as well as a description which later becomes a part of the changelog. everything else is taken care of, including bumping versions of dependencies of dependencies and so on.
> 
> @mechanical-turk i looked at changesets. The implementation seemed both amateurish and limited, lerna seemed to do exactly what we wanted, but maybe I missed something? Maybe I was looking at the docs to run it myself and all we needed was the actions.
> 
> My concerns
> 
> * Lerna allows you to force publish all non-public packages
> * lerna lets you publish all packages to the same number
> * lerna lets us split the tagging and publishing, so we can run ci between
> * lerna uses got tags so we can easily sync the came commit as the release
> 
> I actually fucked up the first published version using change sets any fixed it with Lerna
> 
> that said, lerna was obviously built to be run by hand not in ci, our hotshot contractor Miguel told me to use changesets, and now your recommendation - I will not be offended to find out it really is better! If we can publish all packages to the same number even if one doesn’t have changes, and if we can easily add new packages, or if there’s a good reason not to do that - then great. Let’s use changesets.

Also - if we take the CI bits out of this script, it gets much simpler.

.github/workflows/Publish_npm_packages.yml
texuf: Also this looks like it’s just cut and pasted from my script. Can we just update that one and call it instead. Yes obviously the action needs to be yaml, the reason we put the logic in a bash script is so that anyone can run it anywhere (which also makes it easy to iterate on and debug.)

jterzis: i dont believe the auth expires as the auth token is read from `.npmrc` file. The token itself does expire in 2050. I was not able to create a token in npm w/o an expiry. 

There are a few modifications actually from your script, which i will integrate back in and then also call from the workflow action.

PR #380, opened 2024-07-11
Tak/hnt 8197 create new media stream per town to store the towns image

tak-hntlabs:
Based on discussion, the create media stream rules are updated to support media stream creation for a space, in addition to a channel.


packages/sdk/src/client.ts
```diff
@@ -728,6 +730,57 @@ export class Client
         return { streamId: streamId, prevMiniblockHash: streamView.prevMiniblockHash }
     }
 
+    async createSpaceMediaStream(
+        spaceId: string | undefined,
+        chunkCount: number,
+        streamSettings?: PlainMessage<StreamSettings>,
+    ): Promise<{ streamId: string; prevMiniblockHash: Uint8Array }> {
+        assert(spaceId !== undefined, 'userStreamId must be set')
```
texuf, line 738: This assert is wrong (cut and paste error looks like)

packages/sdk/src/client.ts
```diff
@@ -728,6 +730,57 @@ export class Client
...
+        const spaceIdBytes = streamIdAsBytes(spaceId)
+
+        this.logCall('createSpaceMediaStream', streamId)
+
+        const inceptionEvent = await makeEvent(
+            this.signerContext,
+            make_MediaPayload_Inception({
+                streamId: streamIdBytes,
+                spaceId: spaceIdBytes,
+                channelId: spaceIdBytes,
```
texuf, line 753: I don’t think we should set the channel id to the space id. This seems weird to me.

packages/sdk/src/client.ts
```diff
@@ -728,6 +730,57 @@ export class Client
...
+            [],
+            unpackedResponse.prevSnapshotMiniblockNum,
+            undefined,
+            [],
+            undefined,
+        )
+
+        check(isDefined(streamView.prevMiniblockHash), 'prevMiniblockHash must be defined')
+
+        return { streamId: streamId, prevMiniblockHash: streamView.prevMiniblockHash }
```
texuf, line 781: Are you sure we need a new function here? Can we just make channel id optional and add the scope param to the existing makeMediaStream function?

protocol/protocol.proto
```diff
@@ -497,6 +497,7 @@ message MediaPayload {
         int32 chunk_count = 3;
         StreamSettings settings = 4;
         optional bytes space_id = 5;
+        optional PublicScope public_scope = 6;
```
texuf, line 500: Can you explain what this scope is for? Currently it serves no purpose.

packages/sdk/src/client.ts
tak-hntlabs: agreed. My first cut was to set channelId to `new UintArray(0)` and `new UIntArray(32)` (tried both). But we have strict rule checks on the river node that rejects attempt to create a media stream with an invalid channelid, not tied to any channelId, and without a membership to the associated `channel` stream. 
Tried to disentangled to strict coupling between media and channel, but when i debug the code, I realize that these are good checks to prevent unintended 'mistakes'. e.g. stream creator needs to be a member of the stream already before a media stream associated with that stream can be created. 

Since `channel` is a concept for storing contents, my rationale for setting the 'channelId' to the spaceId is that all the channel rules and logic should apply. Down the road, this `space channel` can be an  `admin channel` for other contents, like bot communication, and what not.

If not, then I'll need your help to understand what rules I can break / change between the space channel, and a regular channel.

protocol/protocol.proto
tak-hntlabs: yes - scope is used to signal that this is a public media channel for storing content. Anyone can decrypt the content stored in this stream with a key that will be openly available. Like a Town's image. This unblocks scenario like NFT image. 
There will be 2 such publicly scoped media streams. One for Town; the other for User's avatar image.

packages/sdk/src/client.ts
tak-hntlabs: There are at least 3 differences:
1. The assert rules are different between a regular media stream and a public media stream
2. The inception payloads are different
3. A 'public decryption key' will be generated and made available. Doing this in a follow up PR.

I had a first cut with `if..else`, but it looked a bit messy to me. So I separated them. I can re-do it in if..else to see how it reads.

tak-hntlabs: fixed

protocol/protocol.proto
texuf: Signal to what or to whom? The fact that keys are or are not available to publicly download has no bearing on the internal state of this structure. 

Asked in another way - how does code change if you delete this property? (Aside from the checks, but changing that is fine)

tak-hntlabs: PublicScope is part of the protocol because the normal group encryption algo does not apply in a public stream. The public media stream will be using a symmetric key for encryption. Only the space owner is allowed to upload to the stream. Anyone can use the public symmetric key to decrypt content in the stream.

So besides the checks, encryption / decryption message processing is also different at the protocol level. Haven't work out the rest of the details yet.

Here is the next chunk of changes with a test that shows how the stream owner will encrypt and upload to the public stream: https://github.com/river-build/river/pull/395

sergekh2 (changes requested):
Lets meet and hash out details of protocol and encryption changes.

protocol/protocol.proto
```diff
@@ -770,6 +771,12 @@ enum ChannelOp {
     CO_UPDATED = 4;
 }
 
+enum PublicScope {
+    PS_UNSPECIFIED = 0;
+    PS_SPACE = 1;
+    PS_USER = 2;
+}
+
```
texuf, line 779: Public scope is unused now right?

core/node/rules/can_create_stream.go
```diff
@@ -242,6 +242,7 @@ func (ru *csParams) canCreateStream() ruleBuilderCS {
 			).
 			requireMembership(
 				inception.ChannelId,
+				inception.SpaceId,
 			).
 			requireChainAuth(ru.getChainAuthForMediaStream)
```
texuf, line 247: Don’t you need to update this chain auth function?

can you make sure at least one of your tests is running against the entitled backend?

protocol/protocol.proto
tak-hntlabs: removed

core/node/rules/can_create_stream.go
tak-hntlabs: added chainAuth for space

core/node/rules/can_create_stream.go
```diff
@@ -242,6 +242,7 @@ func (ru *csParams) canCreateStream() ruleBuilderCS {
 			).
 			requireMembership(
 				inception.ChannelId,
+				inception.SpaceId,
```
texuf, line 245: if spaceId is nil (which is the case when posting images to dm's) the check for RequiredMemberships is going to fail. Looks like we should add a test for uploading media to a dm.

tak-hntlabs: this is handled in the change in `rule_builder.go` in this PR. Take a look and lmk if the change needs more tweaks.
There are two existing media tests that verify creating the media for dm and gdm with nil spaceId. In media.test.ts:
- userCanUploadMediaToDmIfMember
- userCanUploadMediaToGdmIfMember

texuf: Nice.

texuf (approved):

tak-hntlabs:
> Lets meet and hash out details of protocol and encryption changes.

@sergekh2 i can;t merge at all. Do you have specific comments you'd like me to address?

packages/proto/src/gen/payloads_pb.ts
sergekh2, line 1: this file is generated by build and was committed by mistake before, it is now gitignored, please remove it

packages/proto/src/gen/protocol_pb.ts
sergekh2, line 1: same here

sergekh2 (approved):

packages/proto/src/gen/payloads_pb.ts
tak-hntlabs: deleted

packages/proto/src/gen/protocol_pb.ts
tak-hntlabs: deleted

PR #381, opened 2024-07-11
Contract changes to support ERC1155

clemire:
RuleEntitlement, EntitlementGated, and EntitlementDataQueryable are now deprecated for new spaces in favor of RuleEntitlementV2, EntitlementGatedV2, and EntitlementDataQueryableV2, all of which expose their respective V1 interfaces for legacy interoperability. Old contracts are left in code for reference and for the sake of generating bindings, but should no longer be deployed once these contract changes are deployed.

Since V1 contracts are deprecated, contract tests are updated to run against V2 contract versions.

I added a `RuleDataUtil` library that converts between V2 and V1 rule data so that v2 contracts can return RuleData in v1 format for stream and xchain nodes. Role creation and updating would have required creating creating V2 versions of entitlements manager, IEntitlements, and many other contracts, but this happens exclusively through the client, so we can just upgrade all role creation to use V2 rule data and force client upgrades once these contracts are deployed.

The client is updated entirely to use V2 rule data, but stream nodes and xchain nodes are still using V1 endpoints on the new contracts. There will be a follow-up PR that migrates core services to use V2 endpoints.

Tested locally on harmony with some staged changes ready for a follow-up PR and gating on spaces and channels works fine.


contracts/src/factory/facets/architect/ImplementationStorage.sol
```diff
@@ -20,7 +20,7 @@ library ImplementationStorage {
   struct Layout {
     ISpaceOwner spaceToken;
     IUserEntitlement userEntitlement;
-    IRuleEntitlement ruleEntitlement;
+    IRuleEntitlementV2 ruleEntitlement;
```
giuseppecrj, line 23: Do not update, add to the end of the storage

giuseppecrj: keep the other one

giuseppecrj (changes requested):

contracts/src/spaces/facets/gated/EntitlementGatedBaseV2.sol
```diff
@@ -0,0 +1,180 @@
...
+    if (!transaction.hasBenSet) {
+      transaction.hasBenSet = true;
+      transaction.entitlement = entitlement;
+      transaction.clientAddress = msg.sender;
+    }
+
+    transaction.roleIds.push(roleId);
+
+    for (uint256 i = 0; i < selectedNodes.length; i++) {
+      transaction.nodeVotesArray[roleId].push(
```
giuseppecrj, line 55: use EnumerableSet not arrays

contracts/src/spaces/facets/gated/EntitlementGatedStorageV2.sol
```diff
@@ -0,0 +1,28 @@
...
+// contracts
+
+library EntitlementGatedStorageV2 {
+  // keccak256(abi.encode(uint256(keccak256("facets.entitlement.gated.storage.v2")) - 1)) & ~bytes32(uint256(0xff))
+  bytes32 internal constant STORAGE_SLOT =
+    0xc9f85609eb60ac71988cfb4c314695ea9c394b99f001405aaba56d724f4cf800;
+
+  struct Layout {
+    IEntitlementChecker entitlementChecker;
+    mapping(bytes32 => IEntitlementGatedBaseV2.Transaction) transactions;
```
giuseppecrj, line 19: avoid using interface objects from storage objects, use a new transaction struct strictly written inside the library or split into multiple mappings.

giuseppecrj (changes requested):

contracts/src/spaces/facets/gated/EntitlementGatedBaseV2.sol
clemire: This is a straight copy of previous implementation, do you think we could defer to a follow-up PR?

contracts/src/spaces/facets/gated/EntitlementGatedStorageV2.sol
clemire: same comment here

packages/web3/src/entitlement.ts
```diff
@@ -65,7 +67,8 @@ export type CheckOperation = {
     checkType: CheckOperationType
     chainId: bigint
     contractAddress: Address
-    threshold: bigint
+    threshold?: bigint
```
brianathere, line 70: This should be params bytes, not specific data, so we never have to go through this again for new operation types.

PR #382, opened 2024-07-11
Kerem/metrics discovery without volumes

mechanical-turk:
- This PR removes persistent volume requirements in order to simplify the metrics aggregator deployment. This way we won't need to maintain additional cloud components like EFS or EBS. 
- We're also setting up docker builds in this same PR.


.github/workflows/Metrics_discovery_docker.yml
```diff
@@ -0,0 +1,64 @@
...
+    workflow_dispatch:
+
+env:
+    SLACK_WEBHOOK_URL: ${{ secrets.SLACK_CD_WORKFLOW_WEBHOOK_URL }}
+
+jobs:
+    build:
+        name: Build docker image
+
+        runs-on: ubuntu-latest-8-cores
```
jterzis, line 18: Does this need to run on 8-cores ?

mechanical-turk: good question. it doesn't. using ubuntu-latest instead

jterzis (approved):

PR #408, opened 2024-07-15
node: Make stream sync more resilient

bas-vk:
Currently is one or more streams from a client sync are managed by a remote node that goes down/or is unreachable the entire stream sync subscription is cancelled. This is undesirable.

With this change a new sync op message type `SyncOp_SYNC_DOWN` is introduced. When the node that managed the subscription for the client detects a remote node is down it will send `SyncOp_SYNC_DOWN` message to the client to report that it will not get updates for this stream and drops all streams managed by the remote node from the existing subscription. It is up to the client after that to add these streams again.

### Architecture
Introduces the `rpc/sync` package. This defines a handler interface that the service uses to forward incoming client requests. The handlerImpl implements this interface and keeps a mapping of active `StreamSyncOperation`. It will either create a subscription (SyncStreams) or find and forwards the request to the sync operation. The sync operation manages a set of syncers. A syncer is either local (streamCache) or remote (grpc api). Syncers are defined in the `rpc/sync/client` package. Syncers write messages to an internal channel from which the sync operation sends them to the client.


core/node/rpc/sync/operation.go
```diff
@@ -0,0 +1,269 @@
...
+			return err
+		case <-syncOp.ctx.Done():
+			return RiverError(Err_CANCELED, "sync operation cancelled").Tags("syncId", syncOp.SyncID)
+		}
+	}
+
+	return RiverError(Err_CANCELED, "sync operation cancelled").Tags("syncId", syncOp.SyncID)
+}
+
+func (syncOp *StreamSyncOperation) sendMsg(cmd *subCommand) bool {
```
sergekh2, line 262: while works this way, one possible refactor is changing sendMsg to do both write and reply read (and returning error if context is done) - then there is no need to explicitly do it in multiple functions.

core/node/rpc/sync/client/local.go
```diff
@@ -0,0 +1,152 @@
...
+}
+
+func newLocalSyncer(
+	ctx context.Context,
+	localAddr common.Address,
+	streamCache events.StreamCache,
+	cookies []*SyncCookie,
+	messages chan<- *SyncStreamsResponse,
+) (*localSyncer, error) {
+	syncStreamCtx, syncStreamCancel := context.WithCancel(ctx)
```
sergekh2, line 34: is this cancel ever called?

(parent context gets canceled as well, is there a need for separate cancelable context here?)

core/node/rpc/sync/operation.go
```diff
@@ -0,0 +1,269 @@
...
+		streamCache:     streamCache,
+		nodeRegistry:    nodeRegistry,
+	}, nil
+}
+
+// Run the stream sync until either sub.Cancel is called or until sub.ctx expired
+func (syncOp *StreamSyncOperation) Run(
+	req *connect.Request[SyncStreamsRequest],
+	res *connect.ServerStream[SyncStreamsResponse],
+) error {
```
sergekh2, line 74: would it make sense to add this here?
```suggestion
) error {
defer syncOp.cancel()
```

core/node/rpc/sync/client/syncer_set.go
```diff
@@ -0,0 +1,268 @@
...
+	ss.muSyncers.Lock()
+	ss.syncerTasks.Add(len(ss.syncers))
+	for syncerAddr, syncer := range ss.syncers {
+		go func() {
+			syncer.Run()
+			ss.muSyncers.Lock()
+			delete(ss.syncers, syncerAddr)
+			ss.muSyncers.Unlock()
+			ss.syncerTasks.Done()
+		}()
```
sergekh2, line 133: does this capture work correctly? should `syncerAddr, syncer` be args to the goroutine instead?

core/node/rpc/sync/client/syncer_set.go
```diff
@@ -0,0 +1,268 @@
...
+		if syncer, err = newRemoteSyncer(ss.ctx, ss.syncID, nodeAddress, client, []*SyncCookie{cookie}, ss.messages); err != nil {
+			return err
+		}
+	}
+
+	ss.syncers[nodeAddress] = syncer
+	ss.streamID2Syncer[streamID] = syncer
+
+	ss.syncerTasks.Add(1)
+	go func() {
```
sergekh2, line 182: looks like this can be a functions on ss, then here and above in can be go-called instead of same code.

core/node/rpc/sync/client/remote.go
```diff
@@ -0,0 +1,257 @@
...
+	"github.com/river-build/river/core/node/dlog"
+	. "github.com/river-build/river/core/node/protocol"
+	"github.com/river-build/river/core/node/protocol/protocolconnect"
+	. "github.com/river-build/river/core/node/shared"
+)
+
+type remoteSyncer struct {
+	syncStreamCtx    context.Context
+	syncStreamCancel context.CancelFunc
+	syncID           atomic.Value
```
sergekh2, line 22: it seems it doesn't have to be atomic: it is written once during object creation?

core/node/rpc/sync/client/remote.go
```diff
@@ -0,0 +1,257 @@
...
+						Nonce:  fmt.Sprintf("%d", now.Unix()),
+					})); err != nil {
+						s.syncStreamCancel()
+						return
+					}
+					return
+				}
+			case <-s.syncStreamCtx.Done():
+				return
+			case <-s.syncStreamCtx.Done():
```
sergekh2, line 134: same case twice

core/node/rpc/sync/client/remote.go
```diff
@@ -0,0 +1,257 @@
...
+}
+
+func (s *remoteSyncer) Run() {
+	defer s.responseStream.Close()
+
+	var latestMsgReceived atomic.Value
+
+	latestMsgReceived.Store(time.Now())
+
+	go func() {
```
sergekh2, line 96: please put this func in the method and go-call it here: it's relatively long and makes Run() less readable as such

core/node/rpc/sync/client/syncer_set.go
bas-vk: From go 1.22 this isn't an issue anymore.

core/node/rpc/sync/operation.go
```diff
@@ -0,0 +1,269 @@
...
+			if !ok { // messages is closed in syncers when syncOp.ctx is cancelled
+				_ = res.Send(&SyncStreamsResponse{
+					SyncId: syncOp.SyncID,
+					SyncOp: SyncOp_SYNC_CLOSE,
+				})
+				return nil
+			}
+
+			// use the syncID as used between client and subscription node
+			msg.SyncId = syncOp.SyncID
```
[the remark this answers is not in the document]
bas-vk: It must be set for `SYNC_NEW` and `SYNC_CLOSE`. It is debatable if it needs to be set for other updates. Currently it is set and several unittests test (go + sdk) do check for it.

sergekh2 (approved):

PR #409, opened 2024-07-15
node: add status code support to gRPC metrics

bas-vk:
Add the following metrics:

- `grpc_unary_status_code[proc,status]`, unary calls grouped by procedure and status
- `grpc_server_stream_status_code[proc,status]`, server streams grouped by procedure and status

Client streams are more complicated to capture the status code. Since we don't those client streams metrics are not enhanced with status.

### DIFF
--- a/core/node/rpc/metrics_interceptor.go
+++ b/core/node/rpc/metrics_interceptor.go
@@ -2,9 +2,13 @@ package rpc
 
 import (
 	"context"
+	"errors"
+	"strings"
 
 	"connectrpc.com/connect"
 	"github.com/prometheus/client_golang/prometheus"
+	"github.com/river-build/river/core/node/base"
+	"github.com/river-build/river/core/node/infra"
 	"github.com/river-build/river/core/node/shared"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/trace"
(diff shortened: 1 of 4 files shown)


core/node/rpc/metrics_interceptor.go
```diff
@@ -13,18 +14,22 @@ type streamIdProvider interface {
...
+	serverStreamsStatusCode *prometheus.GaugeVec
 }
 
 func (s *Service) NewMetricsInterceptor() connect.Interceptor {
 	return &metricsInterceptor{
-		rpcDuration:       s.rpcDuration,
-		unaryInflight:     s.metrics.NewGaugeVecEx("grpc_unary_inflight", "gRPC unary calls in flight", "proc"),
-		openClientStreams: s.metrics.NewGaugeVecEx("grpc_open_client_streams", "gRPC open client streams", "proc"),
-		openServerStreams: s.metrics.NewGaugeVecEx("grpc_open_server_streams", "gRPC open server streams", "proc"),
+		rpcDuration:             s.rpcDuration,
```
sergekh2, line 27: i'm not sure why durations is initialized this way, I think it can be removed from Service and be moved here for consistently. (also it's "method" on duration and "proc" here, lets unify)

core/node/rpc/metrics_interceptor.go
```diff
@@ -74,11 +92,37 @@ func (i *metricsInterceptor) WrapStreamingHandler(next connect.StreamingHandlerF
...
+	if err == nil {
+		return "success"
+	}
+
+	var connectErr *connect.Error
+	if ok := errors.As(err, &connectErr); ok {
+		return connectErr.Code().String()
+	}
+
+	var riverErr *base.RiverErrorImpl
```
sergekh2, line 122: river error needs to come first: at this stage usually it's connect error which wraps river error, so first if will always succeed (river error code is superset of grpc code and is more detailed).

sergekh2 (approved):

PR #418, opened 2024-07-16
Protocol changes we talked about in meeting

texuf:


texuf:
@tak-hntlabs this is just a draft for reference. whenever you make changes I'll delete this PR.

PR #419, opened 2024-07-16
Move pins to the membership payload, we can now pin to any stream

texuf:


protocol/protocol.proto
```diff
@@ -190,6 +191,15 @@ message MemberPayload {
         bytes token_id = 3;
     }
 
+    message Pin {
```
clemire, line 194: backwards compatibility: If a new client (post-this-PR) sends this payload to an old stream node on omega, is that going to work?

texuf: The request will fail until the node is upgraded.

protocol/protocol.proto
```diff
@@ -190,6 +191,15 @@ message MemberPayload {
         bytes token_id = 3;
     }
 
+    message Pin { 
+        bytes event_id = 1;
```
ttillberg, line 195: could you please add the userId of the creator of the pin here too?

texuf: the creator address is in every single stream event already `bytes creator_address = 1;` it's verified against the signature of the event. Additionally, if we put it here users would be able to forge pins from other people.

texuf: I see how we lost it, i'll fix the streamStateView_Members, one sec.

PR #422, opened 2024-07-17
Update RuleEntitlement to save data as bytes

giuseppecrj:


contracts/src/spaces/entitlements/rule/RuleEntitlementV2.sol
```diff
@@ -190,6 +191,8 @@ contract RuleEntitlementV2 is
     }
 
     delete entitlementsByRoleId[roleId];
+    delete entitlementsByRoleId[roleId].grantedBy;
```
brianathere, line 194: These need to be before the 

```
delete entitlementsByRoleId[roleId];
```

brianathere (approved):
Comment inline.

Why do we want to change this?

giuseppecrj:
> Comment inline.
> 
> Why do we want to change this?

If we're in accordance into how we save data, I want to move to using a namespaced storage approach as per EIP-7201
https://eips.ethereum.org/EIPS/eip-7201

brianathere:
> > Comment inline.
> > Why do we want to change this?
> 
> If we're in accordance in how we save data, I want to move to using a namespaced storage approach as per EIP-7201 https://eips.ethereum.org/EIPS/eip-7201

Thanks, I'll read. I assume it makes sense :-)

contracts/src/spaces/entitlements/rule/RuleEntitlementV2.sol
```diff
@@ -38,17 +37,20 @@ contract RuleEntitlementV2 is
...
     address grantedBy;
     uint256 grantedTime;
-    RuleData data;
+    bytes data;
   }
 
-  mapping(uint256 => Entitlement) internal entitlementsByRoleId;
-
-  address public SPACE_ADDRESS;
+  struct Layout {
```
bas-vk, line 50: Why not annotate `Layout` with `@custom:storage-location` and bump the solc version to ^0.8.20 as erc7201 describes?

PR #423, opened 2024-07-17
Delete and ignore generated files for ts protos

sergekh2:
(not ignored by mistake during package move)


packages/proto/src/.gitignore
```diff
@@ -0,0 +1 @@
+/gen/
```
texuf, line 1: can you update the root .gitignore? it also ignores these files already 

```
packages/proto/src/gen/*
```

texuf: not sure why we need .gitignores scattered everywhere when we already do repo wide ignores in the root .gitignore

PR #426, opened 2024-07-18
Remove react-reply, add react-only permission to protocol

clemire:
WRITE and REACT now separate - users cannot REACT with WRITE, need REACT for reactions.


texuf (approved):

giuseppecrj:
todo: needs proposal on testnet and mainnet

PR #442, opened 2024-07-22
partner registry on space factory

giuseppecrj:


contracts/src/factory/facets/partner/PartnerRegistryBase.sol
```diff
@@ -0,0 +1,133 @@
...
+
+  bytes32 internal constant CURRENT_VERSION = keccak256("1");
+
+  function __PartnerRegistryBase_init(
+    uint256 registryFee,
+    uint256 maxPartnerFee
+  ) internal {
+    PartnerRegistryStorage.Layout storage ds = PartnerRegistryStorage.layout();
+
+    ds.partnerSettingsByVersion[CURRENT_VERSION] = PartnerRegistryStorage
```
jterzis, line 24: I'm not sure what we're trying to achieve with CURRENT_VERSION. This data storage is mutable though I intuitively think that what is trying to be achieved (correct me) is an immutable data store of versioned partner registry settings. If we update the implementation of PartnerRegistry (facet upgrade) but want to use some prior version, we should be able to pass in that version and check that it exists, and then use it if so right ? 

Also, is the intent to be able to support different versions per partner ? If not, I'm not sure I understand why we need to store multiple versions of settings in this map at all.

giuseppecrj: right, this can be removed

pfives (approved):
!

jterzis (approved):
👍

PR #444, opened 2024-07-22
HNT-9124 Add SYNC_DOWN message to synced streams

texuf:


texuf:
@bas-vk can you give me a test plan for this? not sure how to test it myself.

tak-hntlabs (approved):

packages/sdk/src/syncedStreams.ts
```diff
@@ -589,6 +593,30 @@ export class SyncedStreams {
...
+        if (!stream) {
+            this.log('syncDown: stream not found', streamIdAsString(streamId))
+            return
+        }
+        const cookie = stream.view.syncCookie
+        if (!cookie) {
+            this.logError('syncDown: syncCookie not found', streamIdAsString(streamId))
+            return
+        }
+        this.addStreamToSync(cookie).catch((err) => {
```
bas-vk, line 615: If a stream is down it is likely down for some time. For instance when a node restarts after a release upgrade. Adding it immediately after it's reported down will likely fail. 

Not sure if this retry mechanism is something that the sdk needs to support or if it is something that the app using the sdk must do. I do think that the sdk should have the ability to report that it a stream is down. This allows an app to report to the user that some streams won't get updates.

bas-vk:
> @bas-vk can you give me a test plan for this? not sure how to test it myself.

The info endpoint got an extra sub cmd `drop_stream`. It works in the same way as `make_miniblock` and accepts a `sync id` and `stream id` as arguments to point to the stream within a sync op to simulate a down for. The node will forward the request to the node if it is not managing the stream locally, so you can pass in any stream within the sync op. Interesting test cases are:
- drop a random stream from a sync op and ensure that the sdk received the down message and no further updates, even when events are added to the stream.
- drop a random stream from a sync op, add the stream to the sync op and ensure that the latest sync cookie is received that doesn't contain events (should give the same results as when a stream is added for the first time to a sync op with an up to date sync cookie)
- drop a random stream from a sync op, add some events to the stream and add the stream to the sync op and ensure that the new events are received (should give the same results as when a stream is added for the first time to a sync op with an outdated sync cookie)

The go tests have several additional scenarios such as simulating that all streams on a remote node for the client node are down (this cancels the node2node sync op).

PR #447, opened 2024-07-23
Clean up sdk/package.json by deleting old test labels

sergekh2:
Also delete old skipped test


.github/workflows/ci.yml
```diff
@@ -284,7 +284,9 @@ jobs:
                   echo "SSL_CERT_DIR=$SSL_CERT_DIR"
 
             - name: Run node (without entitlements)
-              run: ./core/node/run_single.sh --de -c -r &
+              run: |
+                  ./core/node/run_single.sh --de -c
```
mechanical-turk, line 288: @sergekh2 is `run_single.sh` not a long-running process when executed with `-c`? If it's not, this line should get the `&` too

sergekh2: -c is "config", -r is "run", it was timing out on the wait below, so I split it in the same was as it's split in run_multi jobs.

PR #450, opened 2024-07-23
Update client and stream server to allow redactions from REACT user.

clemire:
Update the react permission implementation to ensure that REACT users can redact their reactions.

### DIFF
--- a/core/node/rules/can_add_event.go
+++ b/core/node/rules/can_add_event.go
@@ -204,7 +204,7 @@ func (params *aeParams) canAddChannelPayload(payload *StreamEvent_ChannelPayload
 	case *ChannelPayload_Redaction_:
 		return aeBuilder().
 			check(params.creatorIsMember).
-			requireChainAuth(params.channelMessageWriteEntitlements)
+			requireChainAuth(params.redactChannelMessageEntitlements)
 	default:
 		return aeBuilder().
 			fail(unknownContentType(content))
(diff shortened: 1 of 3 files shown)


core/node/rules/can_add_event.go
```diff
@@ -204,7 +204,7 @@ func (params *aeParams) canAddChannelPayload(payload *StreamEvent_ChannelPayload
 	case *ChannelPayload_Redaction_:
 		return aeBuilder().
 			check(params.creatorIsMember).
-			requireChainAuth(params.redactChannelMessageEntitlements)
```
texuf, line 207: a sendChannelMessage_Redaction does not result in a ChannelPayload_Redaction_, it's a ChannelPayload_Message. `redactMessage` results in a ChannelPayload_Redaction_, does this change your assumptions about this PR?

clemire: Ah, I made a mistake here. My original PR sent a redaction and I fixed it to send a message with a redaction payload and just added a bug.

clemire: Thank you for the edifying comment!

clemire: I believe the channel message with redaction payload type is the right one here... We discussed it a while ago - this is the self-redaction option, does not require special permissions. and redactMessage is an admin-only client method.

texuf (approved):

PR #465, opened 2024-07-25
Contract interfaces and implementations for updated architect to use RuleEntitlementV2

clemire:
Building on G's RuleEntitlementV2 changes, this PR adds changes to Architect and creates deployment scripts for V2 spaces with testing. Contract implementation also included.


contracts/src/factory/facets/architect/ArchitectBase.sol
```diff
@@ -89,6 +89,14 @@ abstract contract ArchitectBase is Factory, IArchitectBase {
       _deployEntitlement(ims.ruleEntitlement, spaceAddress)
     );
 
+    // Confirm that deployed RuleEntitlement is expected version.
```
clemire, line 92: Inserted this as a sanity check that we don't accidentally call either endpoint with the wrong contract, since the V1 contract cannot properly decode v2 rule data upon space creation

contracts/src/factory/facets/architect/ArchitectBase.sol
```diff
@@ -338,3 +346,195 @@ abstract contract ArchitectBase is Factory, IArchitectBase {
...
+    IUserEntitlement userEntitlement = IUserEntitlement(
+      _deployEntitlement(ims.userEntitlement, spaceAddress)
+    );
+
+    // deploy token entitlement (Assume the implementation is an IRuleEntitlementV2)
+    IRuleEntitlementV2 ruleEntitlementV2 = IRuleEntitlementV2(
+      _deployEntitlement(ims.ruleEntitlement, spaceAddress)
+    );
+
+    if (
```
clemire, line 389: Straight copy of previous implementation with changed types, except for this extra entitlement version check here

contracts/src/factory/facets/architect/ArchitectBase.sol
```diff
@@ -338,3 +346,195 @@ abstract contract ArchitectBase is Factory, IArchitectBase {
...
+            data: abi.encode(requirements.users)
+          })
+        );
+      }
+
+      IRoles(spaceAddress).addRoleToEntitlement(
+        roleId,
+        IRolesBase.CreateEntitlement({
+          module: ruleEntitlementV2,
+          data: requirements.ruleDataV2
```
clemire, line 482: minor deviation: no decoding here since ruleDataV2 is bytes, as discussed with @giuseppecrj

scripts/gen-river-node-bindings.sh
```diff
@@ -57,17 +78,16 @@ generate_go river river IRiverConfig river_config_v1
...
 generate_go river/deploy deploy MockRiverRegistry mock_river_registry
 
-# The following structs get included twice in the generated code, this utility removes them from a file
-#
-#		"IRuleEntitlementBaseCheckOperation":   true,
-#		"IRuleEntitlementBaseLogicalOperation": true,
-#		"IRuleEntitlementBaseOperation":        true,
-#		"IRuleEntitlementBaseRuleData":         true,
+# Each contract will contain a definition of all types it uses as parameters or return types
+# of methods, even if that struct was defined in another contract. This results in duplicate
```
clemire, line 82:
```suggestion
# of methods, even if that struct was defined in another contract, and this sometimes results in duplicate
```

scripts/gen-river-node-bindings.sh
```diff
@@ -57,17 +78,16 @@ generate_go river river IRiverConfig river_config_v1
...
-# The following structs get included twice in the generated code, this utility removes them from a file
-#
-#		"IRuleEntitlementBaseCheckOperation":   true,
-#		"IRuleEntitlementBaseLogicalOperation": true,
-#		"IRuleEntitlementBaseOperation":        true,
-#		"IRuleEntitlementBaseRuleData":         true,
+# Each contract will contain a definition of all types it uses as parameters or return types
+# of methods, even if that struct was defined in another contract, and this sometimes results in duplicate
+# struct declarations. Here, we remove the duplicate struct declarations from the generated
+# files.
```
clemire, line 84:
```suggestion
# Each contract will contain a definition of all types it uses as parameters or return types
# of methods, even if that struct was defined in another contract, and this sometimes results
# in duplicate struct declarations. Here, we remove the duplicate struct declarations from the
# generated files.
```

clemire:
closes HNT-8539

contracts/src/spaces/entitlements/rule/IRuleEntitlement.sol
```diff
@@ -106,6 +106,25 @@ interface IRuleEntitlementBase {
     CheckOperationV2[] checkOperations;
     LogicalOperation[] logicalOperations;
   }
+
+  // Mock check events sometimes utilize the 'threshold' field
+  // as a way to branch the evaluation result for testing.
+  struct MockParams {
```
giuseppecrj, line 112: Why do we need test structs inside production code?

contracts/src/factory/facets/architect/ArchitectBase.sol
```diff
@@ -338,3 +346,195 @@ abstract contract ArchitectBase is Factory, IArchitectBase {
     );
   }
 }
+
+abstract contract ArchitectBaseV2 is ArchitectBase {
```
giuseppecrj, line 350: move this logic to ArchitectBase

giuseppecrj (changes requested):
logic should just live in Architect, rather than duplicate code. that way we don't have to mess with deploy scripts

contracts/scripts/deployments/DeploySpaceFactory.s.sol
```diff
@@ -235,3 +243,50 @@ contract DeploySpaceFactory is DiamondDeployer {
   //   vm.stopBroadcast();
   // }
 }
+
+contract DeploySpaceFactoryV2 is DeploySpaceFactory {
```
giuseppecrj, line 247: we should not be deploying 2 Diamond for such a small change, this will potentially deviate us from the upgrading pattern. I would propose either a legacy facet or just managing it in the new facet.

PR #466, opened 2024-07-25
smart contract work to override permissions by channel

giuseppecrj:
TESTNET PROPOSAL: https://testnets.llama.xyz/orgs/river/base-sepolia/actions/158

## Summary by CodeRabbit

- **New Features**
	- Introduced channel permission management with new methods for setting, getting, and clearing channel permission overrides.
	- Added events for tracking changes in channel role permissions to enhance auditing capabilities.
	- Enhanced role management functionality by streamlining permission checks and introducing mappings for channel-specific permissions.

- **Bug Fixes**
	- Improved internal checks to prevent operations on non-existent roles, enhancing security and reducing errors.

- **Tests**
	- Added tests for new channel permission features to ensure proper functionality and error handling.


contracts/scripts/deployments/facets/DeployRoles.s.sol
```diff
@@ -22,6 +22,12 @@ contract DeployRoles is FacetHelper, Deployer {
     addSelector(Roles.getPermissionsByRoleId.selector);
     addSelector(Roles.addRoleToEntitlement.selector);
     addSelector(Roles.removeRoleFromEntitlement.selector);
+
+    // channel permission overrides
+    addSelector(Roles.createCustomChannelPermissions.selector);
```
brianathere, line 27: I suggest Override instead of Custom.

createChannelOverridePermissions

contracts/src/spaces/facets/entitlements/extensions/EntitlementDataQueryable.sol
```diff
@@ -67,38 +70,56 @@ contract EntitlementDataQueryable is
...
 
-    // Count the number of roles that have the requested permission and record their ids.
+    // iterate through channel roles and check for the requested permission
     for (uint256 i; i < channelRolesLength; i++) {
       uint256 roleId = channelRoles[i];
 
-      RolesStorage.Role storage role = ds.roleById[roleId];
+      RolesStorage.Role storage role = rs.roleById[channelRoles[i]];
 
+      // skip immutable roles
```
brianathere, line 90: I don't understand why we're skipping these

contracts/src/spaces/facets/entitlements/extensions/EntitlementDataQueryable.sol
```diff
@@ -67,38 +70,56 @@ contract EntitlementDataQueryable is
...
+      bool hasPermission = false;
+
+      // check if role is associated with the channel and has the requested permission
+      if (rs.channelsByRole[roleId].contains(channelId)) {
+        StringSet.Set storage permissions = rs.permissionByChannelIdByRoleId[
+          roleId
+        ][channelId];
+        hasPermission = permissions.contains(permission);
+      }
+      // check global permissions if role is not channel-specific
```
brianathere, line 104: check the default permissions if this role didn't have a channel override.

contracts/src/spaces/facets/roles/Roles.sol
```diff
@@ -90,7 +84,41 @@ contract Roles is IRoles, RolesBase, Entitled {
     CreateEntitlement memory entitlement
   ) external {
     _validatePermission(Permissions.ModifyRoles);
-    _checkRoleExists(roleId);
     _removeRoleFromEntitlement(roleId, entitlement);
   }
+
+  // custom channel permissions
+  function createCustomChannelPermissions(
```
brianathere, line 91: Again, overrides, not custom

brianathere: Also, set, not create

contracts/src/spaces/facets/roles/Roles.sol
```diff
@@ -90,7 +84,41 @@ contract Roles is IRoles, RolesBase, Entitled {
...
+  }
+
+  function getCustomChannelPermissions(
+    uint256 roleId,
+    bytes32 channelId
+  ) external view returns (string[] memory permissions) {
+    return _getCustomChannelPermissions(roleId, channelId);
+  }
+
+  function updateCustomChannelPermissions(
```
brianathere, line 107: if we have set, update should be needed

contracts/src/spaces/facets/roles/Roles.sol
```diff
@@ -90,7 +84,41 @@ contract Roles is IRoles, RolesBase, Entitled {
...
+    uint256 roleId,
+    bytes32 channelId,
+    string[] memory permissions
+  ) external {
+    _validatePermission(Permissions.ModifyRoles);
+    _updateCustomChannelPermissions(roleId, channelId, permissions);
+  }
+
+  function removeCustomChannelPermissions(
```
brianathere, line 116: clear, not remove, and shouldn't need a string[] of permissions, as it will remove the override completely.

contracts/src/spaces/facets/roles/RolesBase.sol
```diff
@@ -318,6 +324,118 @@ abstract contract RolesBase is IRolesBase {
     emit RoleRemoved(msg.sender, roleId);
   }
 
+  // =============================================================
+  // Channel Permissions
+  // =============================================================
+  function _getCustomChannelPermissions(
```
brianathere, line 330: I think these need to be adjusted to be get/set/remove(or clear)

brianathere (approved):
With comments

contracts/src/spaces/facets/roles/RolesBase.sol
```diff
@@ -318,6 +324,118 @@ abstract contract RolesBase is IRolesBase {
...
+
+    RolesStorage.Layout storage rs = RolesStorage.layout();
+
+    rs.channelsByRole[roleId].add(channelId);
+
+    // check permissions is not empty
+    uint256 permissionsLen = permissions.length;
+    for (uint256 i = 0; i < permissionsLen; i++) {
+      _checkEmptyString(permissions[i]);
+      if (rs.roleById[roleId].permissions.contains(permissions[i])) {
```
clemire, line 364: This condition is confusing - sorry if I misunderstood! Are you checking that the default permissions completely don't overlap with override permissions?

contracts/src/spaces/facets/roles/RolesBase.sol
```diff
@@ -318,6 +324,118 @@ abstract contract RolesBase is IRolesBase {
...
+    // get current permissions
+    uint256 permissionsLen = permissions.length;
+
+    for (uint256 i = 0; i < permissionsLen; i++) {
+      _checkEmptyString(permissions[i]);
+      if (!permissionsSet.contains(permissions[i])) {
+        revert Roles__PermissionDoesNotExist();
+      }
+      permissionsSet.remove(permissions[i]);
+    }
```
clemire, line 434: If the create method above modifies roles like so,

```
rs.channelsByRole[roleId].add(channelId);
```

Would we also potentially want to do the remove here?

contracts/test/spaces/roles/Roles.t.sol
```diff
@@ -1016,4 +1029,91 @@ contract RolesTest is BaseSetup, IRolesBase, IEntitlementBase {
...
     roles.removeRoleFromEntitlement(roleId, entitlement);
   }
+
+  // =============================================================
+  // Channel Permissions
+  // =============================================================
+  function test_createCustomChannelPermissions()
+    external
+    givenRoleExists
+    givenRoleIsInChannel
```
clemire, line 1039: Are we always assuming the role is in the channel before anyone calls the createCustomChannelPermissions method?

clemire:
Quick work, nice!

contracts/src/spaces/facets/roles/RolesBase.sol
giuseppecrj: it's a set, not an array.

contracts/src/spaces/facets/roles/Roles.sol
giuseppecrj: so set would have to remove already created permissions and add the new ones? what if you only want to remove a single permission? this would require multiple writes for a single change

contracts/src/spaces/facets/roles/RolesBase.sol
giuseppecrj: thanks for the catch!

contracts/src/spaces/facets/entitlements/extensions/EntitlementDataQueryable.sol
```diff
@@ -67,38 +70,56 @@ contract EntitlementDataQueryable is
...
 
-      // Check if the role has the requested permission.
-      if (role.permissions.contains(permission)) {
-        matchedRoleIds[roleCount] = roleId;
-        roleCount++;
+      bool hasPermission = false;
+
+      // check if role is associated with the channel and has the requested permission
+      if (rs.channelsByRole[roleId].contains(channelId)) {
+        StringSet.Set storage permissions = rs.permissionByChannelIdByRoleId[
```
clemire, line 99: nonblocking comment: If possible I'd love for `permissionByChannelIdByRoleId` to be renamed to something that specifies these are override permissions - with the word `override` in the naming, so the code is a little more self-documenting, and it's more obvious why we would privilege this permission set and return on line 106.

Not sure what to suggest through, I think `overridePermissionsByChannelIdByRoleId` is probably a bit too long.

contracts/src/spaces/facets/roles/RolesBase.sol
```diff
@@ -318,6 +324,91 @@ abstract contract RolesBase is IRolesBase {
...
+    ][channelId];
+
+    // get current permissions
+    string[] memory currentPermissions = permissionsSet.values();
+    uint256 currentPermissionsLen = currentPermissions.length;
+    for (uint256 i = 0; i < currentPermissionsLen; i++) {
+      permissionsSet.remove(currentPermissions[i]);
+    }
+
+    rs.channelsByRole[roleId].remove(channelId);
```
clemire, line 407: So, question: if a role applies to a channel and we remove the override permissions, but the role still applies to the channel with it's default permissions, we would not want to remove it from this map here, right? 

I think I didn't clearly get at this, but I'm wondering if adding or removing channels from this map actually makes sense at all from these entrypoints...

contracts/src/spaces/facets/entitlements/extensions/EntitlementDataQueryable.sol
giuseppecrj: might be legacy code, removing for now.

contracts/src/spaces/facets/roles/RolesBase.sol
giuseppecrj: this is strictly for override so we can check if there exists some settings that were placed there, entitlement data queryable uses this storage check to diff between global and channel level permissions

contracts/src/spaces/facets/entitlements/extensions/EntitlementDataQueryable.sol
giuseppecrj: done.

clemire (approved):

texuf (approved):

PR #468, opened 2024-07-26
protocol and snapshot changes for public chunked media

tak-hntlabs:
Main changes:
- `deriveKeyAndIV` function that will return the same `key` and `iv` for the same `spaceId`.
- space snaphot that will return the latest chunked media with the `DerivedAESGCM` encryption.


protocol/protocol.proto
```diff
@@ -252,6 +258,40 @@ message SpacePayload {
     oneof content {
         Inception inception = 1;
         ChannelUpdate channel = 2;
+        ChunkedMedia space_image = 3;
```
texuf, line 261: I know we spec'd this property out to be unencrypted, i think we need to encrypt it (it should also be encrypted with a derived key). 
I don't want unencrypted thumbnails in the space stream.

texuf: @brianathere or @sergekh2  what do you think?

tak-hntlabs: The protocol.proto has this:

```
oneof encryption {
        AESGCM aesgcm = 101;
        DerivedAESGCM derived = 102;
    }
```
We can apply the encryption algo to the thumbnail as well.
Also, I have a question about the original definition of AESGCM:

```
 message AESGCM {
        bytes iv = 1;
        bytes secretKey = 2;
    }
```
What encrypts this?

brianathere: > @brianathere or @sergekh2 what do you think?

We should encrypt the thumbnail as well. For all of the streams. Yes, it means you need to downlod it to decrypt it before render, but we should not have plaintext in the stream

texuf: Tak we should encrypt the whole payload - it becomes `EncryptedData space_image = 3;`

protocol/protocol.proto
```diff
@@ -227,6 +227,12 @@ message SpacePayload {
         Inception inception = 1;
         // channels: sorted by channel_id
         repeated ChannelMetadata channels = 2;
+        SnappedSpaceMedia space_media = 3;
+    }
+
+    message SnappedSpaceMedia {
+        bytes creator_address = 1;
+        EncryptedData space_image = 2;
```
texuf, line 235: can we make these names line up? I think this matches what we're doing elsewhere:
```
SnappedSpaceImage space_image = 3

message SnappedSpaceImage {
    bytes creator_address = 1;
    EncryptedData data = 2;
}
```

packages/sdk/src/encryptedContentTypes.ts
```diff
@@ -13,6 +13,23 @@ export function isEncryptedContentKind(kind: string): kind is EncryptedContent['
...
+    return (
+        typeof data.ciphertext === 'string' &&
+        typeof data.algorithm === 'string' &&
+        typeof data.senderKey === 'string' &&
+        typeof data.sessionId === 'string' &&
+        (typeof data.checksum === 'string' || data.checksum === undefined) &&
+        (typeof data.refEventId === 'string' || data.refEventId === undefined)
+    )
+}
+
```
texuf, line 32: looks like `isEncryptedData` is only for testing can we move it to a test utils file?

packages/sdk/src/client.ts
```diff
@@ -696,10 +708,7 @@ export class Client
             )
         }
 
-        const streamId =
-            !channelId && spaceId
-                ? makeMediaStreamIdFromSpaceId(spaceId)
-                : makeUniqueMediaStreamId()
```
texuf, line 702: looks like you stopped using makeMediaStreamIdFromSpaceId, can we delete the function in id.ts?

protocol/protocol.proto
tak-hntlabs: looked through protocol.proto. The Snapped* and Wrapped* fields have specific names. Not a generic name like `data`

packages/sdk/src/encryptedContentTypes.ts
tak-hntlabs: moved

packages/sdk/src/client.ts
tak-hntlabs: removed

protocol/protocol.proto
tak-hntlabs: example: 
```
 message SnappedPin {
        bytes creator_address = 1;
        Pin pin = 2;
    }
```

texuf: WrappedEncryptedData has a data field. That is what I want to copy. 
Pins is snapshotted in a list, which has different usage syntax.

tak-hntlabs: renamed

core/node/crypto/encrypt.go
```diff
@@ -0,0 +1,5 @@
+package crypto
+
+const (
+	AES_GCM_DERIVED_ALGORITHM = "r.aes-256-gcm.derived"
+)
```
texuf, line 5: can you delete this file and this constant? the stream nodes don't care about what encryption you're using (yet). It's currently only for testing, please just send the algorithm string from the test

protocol/protocol.proto
```diff
@@ -227,6 +227,12 @@ message SpacePayload {
         Inception inception = 1;
         // channels: sorted by channel_id
         repeated ChannelMetadata channels = 2;
+        SnappedSpaceMedia space_media = 3;
+    }
+
+    message SnappedSpaceMedia {
+        bytes creator_address = 1;
+        EncryptedData data = 2;
```
texuf, line 235: we had a miscommunciation about the rename
the idea is that there is a payload
```
EncryptedData space_image = 3;
```
and then there is a snapshotted version of this payload, where the name matches
`SnappedSpaceImage`


the current pr has a `space_image` getting snapped into a `space_media` object named SpaceMedia, which are not the same



NEW

[12:54](https://hntlabs.slack.com/archives/D03DTUS95UL/p1722455644938149)
the idea is that there is a payload
EncryptedData space_image = 3;
and then there is a snapshotted version of this payload, where the name matches
SnappedSpaceImage
12:54
the current pr has a space_image getting snapped into a space_media object, which are not the same

texuf (approved):

PR #473, opened 2024-07-26
chore(contract): Update README and dependencies

shuhuiluo:
Reorganize dependencies by moving development-related packages to `devDependencies`. Enhance the README for better readability with structured formatting and shell command snippets, and add a `.prettierignore` file to exclude certain directories from Prettier formatting.

Introduced a new `.solhint.json` configuration file and updated key dependencies in `package.json`. Improved code formatting rules in `.prettierrc.js` and unified style for conditional deployer address assignment in deployment scripts. Fixed a linting command issue in `makefile`.

### DIFF
--- a/contracts/.prettierignore
+++ b/contracts/.prettierignore
@@ -1,3 +1,0 @@
-cache/
-deployments/
-out/
--- a/contracts/.prettierignore
+++ b/contracts/.prettierignore
@@ -1,0 +1,3 @@
+cache/
+deployments/
+out/
--- a/contracts/.prettierrc.js
+++ b/contracts/.prettierrc.js
@@ -1,4 +1,5 @@
 module.exports = {
+    plugins: ['prettier-plugin-solidity'],
     arrowParens: 'always',
     endOfLine: 'lf',
     printWidth: 80,
--- a/contracts/.prettierrc.js
+++ b/contracts/.prettierrc.js
@@ -1,5 +1,4 @@
 module.exports = {
-    plugins: ['prettier-plugin-solidity'],
     arrowParens: 'always',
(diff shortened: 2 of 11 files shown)


texuf:
Nice

contracts/.solhint.json
```diff
@@ -0,0 +1,7 @@
+{
+    "extends": "solhint:default",
```
giuseppecrj, line 2: There's already a solhint.json at the root of the project. https://github.com/river-build/river/blob/main/.solhint.json

contracts/.prettierrc.js
```diff
@@ -1,4 +1,5 @@
 module.exports = {
+    plugins: ['prettier-plugin-solidity'],
```
giuseppecrj, line 2: can you extend from our current prettier-config and add this as an extension

contracts/.prettierignore
```diff
@@ -0,0 +1,3 @@
+cache/
```
giuseppecrj, line 1: there's already a prettierignore at the root of the project

giuseppecrj (changes requested):

contracts/.prettierrc.js
shuhuiluo: Done.

contracts/.prettierignore
shuhuiluo: Gotcha.

contracts/.solhint.json
shuhuiluo: Have to make some changes to pass CI.

giuseppecrj (approved):

PR #482, opened 2024-07-27
refactor(sdk): Switch to `BaseProvider` in entitlement.ts

shuhuiluo:
Updated multiple functions to use `ethers.providers.BaseProvider` instead of `StaticJsonRpcProvider` for broader compatibility. This change ensures a more flexible provider usage across different wallet operations.

### DIFF
--- a/packages/web3/src/entitlement.ts
+++ b/packages/web3/src/entitlement.ts
@@ -330,7 +330,7 @@ export function treeToRuleData(root: Operation): IRuleEntitlementBase.RuleDataSt
 async function evaluateAndOperation(
     controller: AbortController,
     linkedWallets: string[],
-    providers: ethers.providers.StaticJsonRpcProvider[],
+    providers: ethers.providers.BaseProvider[],
     operation?: AndOperation,
 ): Promise<EntitledWalletOrZeroAddress> {
     if (!operation?.leftOperation || !operation?.rightOperation) {
@@ -396,7 +396,7 @@ async function evaluateAndOperation(
 async function evaluateOrOperation(
     controller: AbortController,
     linkedWallets: string[],
-    providers: ethers.providers.StaticJsonRpcProvider[],
+    providers: ethers.providers.BaseProvider[],
     operation?: OrOperation,
 ): Promise<EntitledWalletOrZeroAddress> {
     if (!operation?.leftOperation || !operation?.rightOperation) {
@@ -458,7 +458,7 @@ async function evaluateOrOperation(
 async function evaluateCheckOperation(
     controller: AbortController,
     linkedWallets: string[],
-    providers: ethers.providers.StaticJsonRpcProvider[],
+    providers: ethers.providers.BaseProvider[],
     operation?: CheckOperation,
 ): Promise<EntitledWalletOrZeroAddress> {
     if (!operation) {
@@ -565,7 +565,7 @@ async function evaluateCheckOperation(
 export async function evaluateOperationsForEntitledWallet(
     operations: Operation[],
     linkedWallets: string[],
-    providers: ethers.providers.StaticJsonRpcProvider[],
+    providers: ethers.providers.BaseProvider[],
 ) {
     const controller = new AbortController()
     const result = evaluateTree(
@@ -581,7 +581,7 @@ export async function evaluateOperationsForEntitledWallet(
 export async function evaluateTree(
     controller: AbortController,
     linkedWallets: string[],
-    providers: ethers.providers.StaticJsonRpcProvider[],
+    providers: ethers.providers.BaseProvider[],
     entry?: Operation,
 ): Promise<EntitledWalletOrZeroAddress> {
     if (!entry) {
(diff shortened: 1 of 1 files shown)


scripts/deploy-contracts.sh
```diff
@@ -47,6 +47,12 @@ if [ "${BASE_EXECUTION_CLIENT}" != "geth_dev" ]; then
 fi
 cast rpc evm_setAutomine true --rpc-url $RIVER_ANVIL_RPC_URL
 
+# Only anvil supports setCode but this might be a local geth node
+if [ "${BASE_EXECUTION_CLIENT}" != "geth_dev" ]; then
+    cast rpc anvil_setCode $SEQUENCE_MULTICALL_ADDRESS $SEQUENCE_MULTICALL_BYTECODE --rpc-url $BASE_ANVIL_RPC_URL
+fi
+cast rpc anvil_setCode $SEQUENCE_MULTICALL_ADDRESS $SEQUENCE_MULTICALL_BYTECODE --rpc-url $RIVER_ANVIL_RPC_URL
+
```
texuf, line 55: Both of these lines look the same. Should there be an else, or different code for geth?

shuhuiluo: I mimicked line 45-48. As far as I understand, the River chain is always deployed by Anvil. But the Base chain may be on Anvil or a local Geth.

texuf: yes you're right. thank you.

packages/web3/src/entitlement.ts
```diff
@@ -494,7 +495,7 @@ export async function evaluateOperationsForEntitledWallet(
 export async function evaluateTree(
     controller: AbortController,
     linkedWallets: string[],
-    providers: ethers.providers.StaticJsonRpcProvider[],
+    providers: ethers.providers.BaseProvider[],
```
texuf, line 498: I'm 100% for making this change to the BaseProvider so that a user can use whatever provider they want. 

I'm not excited about routing all of our calls through someone elses contract via an address that is updated via NPM

Can we deploy our own multi call contract and use that address with this library?

contracts/.env.localhost
```diff
@@ -21,3 +21,7 @@ RIVERSCAN_URL=
...
...0195508f603f870112610b3a57600094508485fd5b8c86013594508a851115610b5057610b5061093b565b610b608d85601f88011601610993565b93508484528f82868801011115610b775760008081fd5b848287018e86013760009484018d019490945250918201528352509186019190860190610a72565b9998505050505050505050565b604080825283519082018190526000906020906060840190828701845b82811015610be7578151151584529284019290840190600101610bc9565b50505083810382850152845180825282820190600581901b8301840187850160005b83811015610c55577fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffe0868403018552610c438383516108de565b94870194925090860190600101610c09565b50909998505050505050505050565b600060208284031215610c7657600080fd5b5051919050565b7f4e487b7100000000000000000000000000000000000000000000000000000000600052603260045260246000fd5b60008251610cbe8184602087016108ba565b9190910192915050565b828152604060208201526000610ce160408301846108de565b949350505050565b60007fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff8203610d41577f4e487b7100000000000000000000000000000000000000000000000000000000600052601160045260246000fd5b506001019056fea264697066735822122049ef042991a3ec1b25eb7d0bd2d2faaceec6986d39c10052cc3d76bbe0bb4ad664736f6c63430008120033
```
giuseppecrj, line 27: Not sure I'm happy saving this bytecode into environment variables, we have our own MultiCall.sol contract, I would rather use that instead of an external contract for this.

packages/web3/src/entitlement.ts
giuseppecrj: We have our own MultiCall contract for this https://github.com/river-build/river/blob/main/contracts/src/diamond/utils/multicall/Multicall.sol

giuseppecrj (changes requested):

packages/web3/src/entitlement.ts
shuhuiluo: 0xsequence should be an established team. Unfortunately their [`MulticallProvider`](https://www.npmjs.com/package/@0xsequence/multicall) uses their own [`MultiCallUtils`](https://github.com/0xsequence/wallet-contracts/blob/master/contracts/modules/utils/MultiCallUtils.sol) which is deployed and verified [universally](https://blockscan.com/address/0xdbbFa3cB3B087B64F4ef5E3D20Dda2488AA244e6). We can deploy our own instance but it will still have to adhere to their interface.
If using viem multicall, the [`Multicall3`](https://www.multicall3.com/) contract is used.

texuf: @shuhuiluo are you still going to pull the StaticJsonRpcProvider ==> BaseProvider change out of this pr and submit it? I'd love to see that change get in.

shuhuiluo: Sure.

texuf (approved):
approved once fully merged of course.

giuseppecrj (approved):

PR #576, opened 2024-07-30
Rename and deprecate scrollbackToDate

ttillberg:
`scrollbackToDate` is misleading since it actually compares `Date.now() - toDate` under the hood

- rename it to scrollbackByMs
- for  backwards compatibility, keeping `scrollbackToDate` available with a `@deprecated` disclaimer


texuf (approved):
🤦

PR #577, opened 2024-07-30
Add alpha config for cmd line river_node tool

sergekh2:
+add missing addrs for other envs.
For example
```
./core/node/env/alpha/run.sh registry streams --count
```
Prints stream count from alpha registry contract.


mechanical-turk (changes requested):
@sergekh2 reminding you to hold off on merging the alpha addresses for now - since they are about to change

sergekh2:
> @sergekh2 reminding you to hold off on merging the alpha addresses for now - since they are about to change

updated now

PR #587, opened 2024-07-30
Add autojoin and show join leave events

clemire:
space payload snapshot channel metadata includes new `ChannelSettings` message with fields for
- autojoin
- hide user join and leave events

These can be optionally included in the Channel Inception message, if not all fields are false by default.
Two new messages on Space Payload to manually toggle these two settings, as well as client endpoints.

migration sets autojoin as true for default channels (22 trailing 0s), but from this PR on all channels will be not autojoin unless explicitly toggled on. snapshot migrations apply these defaults.

e2e tests validate:
- settings properly set on channel inception
- client events emitted on changes to both settings
- client space streamview state updates appropriately in response to inception and update messages
- only permitted users can update these properties


core/node/rules/can_add_event.go
```diff
@@ -261,6 +271,15 @@ func (params *aeParams) canAddSpacePayload(payload *StreamEvent_SpacePayload) ru
...
 		}
+	case *SpacePayload_UpdateChannelAutojoin_:
+		ru := &aeUpdateAutojoinRules{
+			params:         params,
+			updateAutojoin: content.UpdateChannelAutojoin,
+		}
+		return aeBuilder().
+			check(ru.validUpdateAutojoin).
+			requireChainAuth(params.updateAutojoinMessageChannelModifyRequirements)
+
```
clemire, line 282: TODO: case for show user join/leave with validation
TODO: correct permission?

core/node/events/migrations/snapshot_migration_0002.go
```diff
@@ -0,0 +1,25 @@
...
+	case *Snapshot_SpaceContent:
+		for _, channel := range snapshot.SpaceContent.Channels {
+			channelId, err := shared.StreamIdFromBytes(channel.ChannelId)
+			if err != nil {
+				panic(err)
+			}
+			if shared.IsDefaultChannelId(channelId) {
+				channel.Autojoin = true
+			} else {
+				channel.Autojoin = false
```
clemire, line 19: unneeded setting to false

texuf, line 19: you don't need to set things to false

core/node/events/migrations/snapshot_migration_0002.go
```diff
@@ -0,0 +1,25 @@
...
+			}
+			channel.ShowUserJoinLeaveEvents = true
```
clemire, line 21: Change this to "Hide"

texuf, line 21: i suggest naming this HideuserJoinLeaveEvents so that the default value is what we want by default.

core/node/events/snapshot.go
```diff
@@ -191,14 +191,51 @@ func update_Snapshot_Space(iSnapshot *Snapshot, spacePayload *SpacePayload, even
...
+		if channel.Op == ChannelOp_CO_CREATED {
+			channel.Autojoin = shared.IsDefaultChannelId(channelStreamId)
+			channel.ShowUserJoinLeaveEvents = true
+		}
+
 		snapshot.SpaceContent.Channels = insertChannel(snapshot.SpaceContent.Channels, channel)
 		return nil
+	case *SpacePayload_UpdateChannelAutojoin_:
+		var channel *SpacePayload_ChannelMetadata = nil
+		for _, c := range snapshot.SpaceContent.Channels {
```
clemire, line 215: use `findChannel` helper

texuf, line 215: there's a findChannel function, please use that instead

core/node/events/snapshot.go
```diff
@@ -191,14 +191,51 @@ func update_Snapshot_Space(iSnapshot *Snapshot, spacePayload *SpacePayload, even
...
 			ChannelId:         content.Channel.ChannelId,
 			Op:                content.Channel.Op,
 			OriginEvent:       content.Channel.OriginEvent,
 			UpdatedAtEventNum: eventNum,
 		}
+		// Apply default channel settings on creation
+		if channel.Op == ChannelOp_CO_CREATED {
+			channel.Autojoin = shared.IsDefaultChannelId(channelStreamId)
+			channel.ShowUserJoinLeaveEvents = true
+		}
```
clemire, line 209: retain Autojoin, etc on update

core/node/events/snapshot.go
```diff
@@ -191,14 +191,51 @@ func update_Snapshot_Space(iSnapshot *Snapshot, spacePayload *SpacePayload, even
...
+
```
texuf, line 210: we need to update the existing channel metadata here otherwise we lose the new properties

core/node/events/snapshot_test.go
```diff
@@ -123,7 +123,7 @@ func TestMakeSnapshot(t *testing.T) {
 	wallet, _ := crypto.NewWallet(ctx)
 	streamId := UserStreamIdFromAddr(wallet.Address)
 	inception := make_User_Inception(wallet, streamId, t)
-	snapshot, err := Make_GenisisSnapshot([]*ParsedEvent{inception})
+	snapshot, err := Make_GenesisSnapshot([]*ParsedEvent{inception})
```
clemire, line 126: These can go in separate PRs w/o approval

core/node/rules/can_add_event.go
```diff
@@ -261,6 +271,24 @@ func (params *aeParams) canAddSpacePayload(payload *StreamEvent_SpacePayload) ru
...
+	case *SpacePayload_UpdateChannelShowUserJoinLeaveEvents_:
+		ru := &aeUpdateChannelShowUserJoinLeaveEventsRules{
+			params: params,
+			update: content.UpdateChannelShowUserJoinLeaveEvents,
+		}
+		return aeBuilder().
+			check(ru.validUpdateChannelShowUserJoinLeaveEvents).
+			requireChainAuth(params.updateChannelShowUserJoinLeaveEventsRequirements)
+
```
texuf, line 291: i think we should add in creatorIsMember checks for these here

core/node/rules/can_add_event.go
```diff
@@ -261,6 +271,24 @@ func (params *aeParams) canAddSpacePayload(payload *StreamEvent_SpacePayload) ru
...
+			requireChainAuth(params.updateAutojoinMessageChannelModifyRequirements)
+
+	case *SpacePayload_UpdateChannelShowUserJoinLeaveEvents_:
+		ru := &aeUpdateChannelShowUserJoinLeaveEventsRules{
+			params: params,
+			update: content.UpdateChannelShowUserJoinLeaveEvents,
+		}
+		return aeBuilder().
+			check(ru.validUpdateChannelShowUserJoinLeaveEvents).
+			requireChainAuth(params.updateChannelShowUserJoinLeaveEventsRequirements)
```
clemire, line 290: write common method, channelExists and have both valid methods call it

core/node/rules/can_add_event.go
```diff
@@ -1048,6 +1076,56 @@ func (params *aeParams) channelMessageReactEntitlements() (*auth.ChainAuthArgs,
...
+		return nil, err
+	}
+
+	chainAuthArgs := auth.NewChainAuthArgsForSpace(
+		spaceId,
+		userId,
+		auth.PermissionAddRemoveChannels,
+	)
+
+	return chainAuthArgs, nil
```
texuf, line 1126: can you create a third function called "channelExistsInSpace" and call it from these two new functions?

core/node/rules/can_add_event.go
```diff
@@ -1256,6 +1334,48 @@ func (ru *aeUnpinRules) validUnpin() (bool, error) {
...
+		return false, err
+	}
+
+	// check if the channel exists
+	_, err = view.GetChannelInfo(channelId)
+	if err != nil {
+		return false, err
+	}
+
+	return true, nil
```
texuf, line 1376: can you create a third function called "channelExistsInSpace" and call it from these two new functions?

core/node/rules/can_add_event.go
texuf, line 1126: the existing function that also uses add remove channels is much smaller, can you follow that convention? `getCreateSpaceChainAuth`

core/node/rules/can_add_event.go
clemire: try getCreateSpaceChainAuth

core/node/shared/ids.go
```diff
@@ -43,21 +44,57 @@ func AddressFromSpaceId(spaceId StreamId) (common.Address, error) {
...
+	// leave the rest unwritten, which defaults to zeroes
 	if spaceId.Type() != STREAM_SPACE_BIN {
 		return StreamId{}, RiverError(Err_BAD_STREAM_ID, "invalid stream type for space", "streamId", spaceId)
 	}
 	var b [32]byte
 	b[0] = STREAM_CHANNEL_BIN
 	copy(b[1:], spaceId[1:21])
-	_, err := rand.Read(b[21:])
+	return StreamIdFromBytes(b[:])
+}
```
texuf, line 83: can you rename this function? I thought it was something with a space prefix?
maybe just `makeChannelPrefix`?

core/node/shared/ids.go
```diff
@@ -43,21 +44,57 @@ func AddressFromSpaceId(spaceId StreamId) (common.Address, error) {
...
 
+func MakeSpaceId() (StreamId, error) {
+	var b [32]byte
+	b[0] = STREAM_SPACE_BIN
+	_, err := rand.Read(b[1:21])
+	if err != nil {
+		return StreamId{}, RiverError(Err_INTERNAL, "failed to create random bytes", "err", err)
+	}
+	return StreamIdFromBytes(b[:])
+}
```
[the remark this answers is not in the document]
texuf: nevermind just leave this here

core/node/shared/ids.go
texuf: or `makeChannelIdPrefixBytes`

protocol/protocol.proto
```diff
@@ -240,6 +240,8 @@ message SpacePayload {
         bytes channel_id = 2;
         EventRef origin_event = 3;
         int64 updated_at_event_num = 6;
+        bool autojoin = 7;
+        bool show_user_join_leave_events = 8;
```
texuf, line 244: ChannelProperties channel_properties = 8;

clemire:
channel inception fields to channel metadata
in channel inception event I can add whatever I want... including properties from another payload.
want to enable devs to set all of these properties when they create a channel.
cannot send a bundle of events, have createStream and addEvent.

create a channel requires you are a member of the space, appends an event to the space stream saying, we will create this channel (channel update event), then it creates the channel.
add optional channelProperties to update autojoin, show...

pipe channel properties all the way through to client on channel creation.

protocol/protocol.proto
texuf: Then we can add ChannelProperties to the ChannelUpdate and the ChannelPayload.Inception and set them all on channel creation.

core/node/rules/can_add_event.go
```diff
@@ -261,6 +271,24 @@ func (params *aeParams) canAddSpacePayload(payload *StreamEvent_SpacePayload) ru
 				check(params.creatorIsValidNode).
 				check(ru.validSpaceChannelOp)
 		}
+	case *SpacePayload_UpdateChannelAutojoin_:
+		ru := &aeUpdateAutojoinRules{
+			params:         params,
+			updateAutojoin: content.UpdateChannelAutojoin,
+		}
+		return aeBuilder().
+			check(ru.validUpdateAutojoin).
```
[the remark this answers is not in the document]
clemire: just creatorIsMember for now, node updates come through SpacePayload_Channel

core/node/shared/ids.go
```diff
@@ -43,21 +44,57 @@ func AddressFromSpaceId(spaceId StreamId) (common.Address, error) {
...
+		return StreamId{}, err
+	}
+	_, err = rand.Read(b[21:])
+	if err != nil {
+		return StreamId{}, RiverError(Err_INTERNAL, "failed to create random bytes", "err", err)
+	}
+	return StreamIdFromBytes(b[:])
+}
+
+func makeSpacePrefixChannelBytes(spaceId StreamId) (StreamId, error) {
```
texuf, line 72: this is returning a "channel prefix" but the function name makes me think it's returning a space prefix something, it's very unclear - can you simplify this name some more?

texuf (commented):
It looks great. Can you fix the merge and compile errors?

core/node/shared/ids.go
clemire: Oh my bad, I thought I renamed this. Will do!

core/node/events/migrations/snapshot_migration_0002.go
```diff
@@ -0,0 +1,25 @@
...
+func snapshot_migration_0002(iSnapshot *Snapshot) *Snapshot {
+	switch snapshot := iSnapshot.Content.(type) {
+	case *Snapshot_SpaceContent:
+		for _, channel := range snapshot.SpaceContent.Channels {
+			if channel.Settings == nil {
+				channel.Settings = &SpacePayload_ChannelSettings{}
+			}
+			channelId, err := shared.StreamIdFromBytes(channel.ChannelId)
+			if err != nil {
+				panic(err)
```
texuf, line 17: i'm a little worried about this panic... and i realize we don't have a context to log anything either... can we just return here instead?

core/node/rules/can_add_event.go
```diff
@@ -1280,6 +1307,54 @@ func (ru *aeUnpinRules) validUnpin() (bool, error) {
 	return false, RiverError(Err_INVALID_ARGUMENT, "message is not pinned")
 }
 
+type HasChannelIdBytes interface {
+	ChannelIdBytes() ([]byte, error)
+}
+
+type autojoinWrapper struct {
+	update *SpacePayload_UpdateChannelAutojoin
+}
```
texuf, line 1316: I really like this! can we follow the convention and make a aeSpaceUpdateChannelRules struct at the top of this file?

Can we move the HasChannelIdBytes interface to the top of the file?

Why is HasChannelIdBytes public?

Can we follow convention for hideUserJoinLeaveEventsWrapper as well?

core/node/rules/can_create_stream.go
```diff
@@ -484,13 +484,14 @@ func (ru *csChannelRules) derivedChannelSpaceParentEvent() (*DerivedEvent, error
 	}
 
 	payload := events.Make_SpacePayload_ChannelUpdate(
```
texuf, line 486: it looks like Evan might not get to this work this week, it's not his highest priority, so we need to make sure it's backwards compatable with old clients by checking for a nil ChannelSettings here, and if ChannelSettings is nil, adding an autoJoin=true if the channelId is a default channel id?

clemire: whoa good catch

clemire: Well actually, I think the snapshot will render the channel setting as autojoin=true when the channelupdate is emitted with empty channel settings for the default channel... but if you'd prefer this change, no problem.

texuf: Oh wow - yeah looks like that would happen - can we make this change anyway in case that code changes or we start looking at the snapshot in a different way in the future?

texuf: actually never mind, lets leave this how it is

clemire: Oh I went ahead and made it, but I left the old one there too

core/node/events/migrations/snapshot_migration_0002.go
clemire: I can also just continue and we can try to migrate as many channels as we can

clemire: panic removed

texuf: Yes good call, continue is better

texuf (approved):

PR #590, opened 2024-07-31
dlog: pretty print multiline only on dev mode

### LINKED ISSUE #544: Bots - 4th week of july - Maintence #3
### Description

previous issue https://github.com/river-build/river/issues/543

the logs on datadog are currently split over lots of lines

we need to fix it so that it prints properly https://app.datadoghq.com/logs?query=env%3Agamma%20service%3Astress-test-node%20&agg_m=count&agg_m_source=base&agg_t=count&cols=env%2Cservice%2Ccontainer_index&event=AgAAAZD_wsFEOSHwlAAAAAAAAAAYAAAAAEFaRF93dFNDQUFESDJEbUlxUEtTcXdBQwAAACQAAAAAMDE5MGZmYzMtYzk0Ny00MTEwLTgyYzgtMDY4ZTA3ZWQwMGEz&fromUser=true&messageDisplay=inline&refresh_mode=sliding&saved-view-id=2705025&storage=hot&stream_sort=desc&viz=stream&from_ts=1722029116522&to_ts=1722030016522&live=true

for example, this should just be one log

image

you can run the tests as many times as you like with the dropdown on this task https://github.com/HereNotThere/harmony/actions/workflows/Run_river_stress_tests.yml

miguel-nascimento:
This PR changes the dlog implementation to print multiline only in dev mode. 
Hoping to fix the issue with logging services like Datadog where we had each line in a different log

Closes #544


packages/dlog/src/dlog.ts
```diff
@@ -111,10 +109,10 @@ const _cloneAndFormat = (
                 }
             }
         }
-        return newObj
+        return JSON.stringify(newObj)
     }
 
-    return obj
+    return JSON.stringify(obj)
```
texuf, line 115: i don't think we want to json things here. this will be really slow.
is this the only way?

texuf (approved):

PR #602, opened 2024-08-01
Harden against adding duplicate events to db; do not load stream if dub is in db

sergekh2:


bas-vk (approved):

core/node/events/minipool.go
```diff
@@ -22,13 +22,15 @@ func newMiniPoolInstance(events eventMap, generation int64, eventNumOffset int64
...
-func (m *minipoolInstance) copyAndAddEvent(event *ParsedEvent) *minipoolInstance {
+func (m *minipoolInstance) tryCopyAndAddEvent(event *ParsedEvent) *minipoolInstance {
 	m = &minipoolInstance{
 		events:         m.events.Copy(1),
 		generation:     m.generation,
 		eventNumOffset: m.eventNumOffset,
 	}
-	m.events.Set(event.Hash, event)
+	if !m.events.Set(event.Hash, event) {
+		return nil
```
mechanical-turk, line 32: @sergekh2 should we do an error log here, since this was originally undesired?

sergekh2: it triggers error creation up the stack and then logging.

PR #604, opened 2024-08-01
Set lastAccessedTime when initializing streams

texuf:
Is this necessary? It seems like the cache will get flushed imediately if not, but maybe i missed something.


bas-vk (approved):
I believe you are right and this needs to be added. We should use a builder function to create streamImpl instances instead of these in line constructions.

PR #605, opened 2024-08-01
chore(contract): Remove base64 library and update forge-std dependency

shuhuiluo:
Removed the base64 library from the project as it's no longer needed, replaced all its references with solady's Base64. Additionally, updated the forge-std dependency to version 1.9.1. This simplifies dependencies and ensures consistency.


texuf:
lgtm

jterzis:
I see were swapping out OZ strings library with LibString from solady here as well. Can we replace the legacy strings library from `SpaceOwner.t.sol` https://github.com/river-build/river/blob/main/contracts/test/spaces/owner/SpaceOwner.t.sol#L12 so that were only depending on LibString for string utilities ?

shuhuiluo:
> I see were swapping out OZ strings library with LibString from solady here as well. Can we replace the legacy strings library from `SpaceOwner.t.sol` https://github.com/river-build/river/blob/main/contracts/test/spaces/owner/SpaceOwner.t.sol#L12 so that were only depending on LibString for string utilities ?

I'm glad you asked. In fact, I kept OZ `Strings` library in tests on purpose to cross check with Solady `LibString`, even though Solady libraries are efficient and extensively tested.

jterzis (approved):

PR #607, opened 2024-08-02
space metadata message for { space image, spaceId }

tak-hntlabs:
When fixing up the client to use the space image, I'm finding places in the client that needs more context than just an EncryptedData space_image in the SpacePayload.  

This change is similar to the `ChannelMetadata` in the SpacePayload:

```
message SpaceMetadata {
        bytes space_id = 1;
        EncryptedData space_image = 2;
}
```

This is also a future-looking struct for additional metadata we want to store in the stream.


protocol/protocol.proto
```diff
@@ -255,10 +257,16 @@ message SpacePayload {
         EventRef origin_event = 3;
     }
 
+    message SpaceMetadata {
+        bytes space_id = 1;
+        EncryptedData space_image = 2;
+    }
```
texuf, line 263: I'm very confused - space_id is just the stream id right? Why do you need it here? Maybe we can sync up tomorrow.

tak-hntlabs: - Yes, space_id is the same as stream_id.
- User scenario: Alice updates her town image. Bob receives a SpacePayload event. The payload only has the encrypted data.
- On client, the React components use the ZTEvent to render changes. Seeing the Pin / Unpin / SpaceUsernameEvent / SpaceDisplayNameEvent - all have a userId field that their respective React components can react to.
- SpaceImageEvent does not have a spaceId in the payload to use when processing the incoming SpacePayload metadata. Adding a space_id would give the React components a prop to react to.

Sent you a meeting invite to sync up.

tak-hntlabs: caught up with Austin. The use case on client is only for sample app. Discard this pr.

PR #611, opened 2024-08-02
Add Stream Registry API for fetching the nodes a batch of streams are on to use in the notification service to optimize startup

brianathere:
This will allow the notification service to efficiently fetch the stream nodes it should route syncStreams requests to.


contracts/src/river/registry/facets/stream/StreamRegistry.sol
```diff
@@ -252,6 +252,20 @@ contract StreamRegistry is IStreamRegistry, RegistryModifiers {
...
 
+  function getStreams(
+    bytes32[] calldata streamIds
+  ) external view returns (uint256 foundCount, StreamWithId[] memory) {
+    uint256 streamCount = streamIds.length;
+    StreamWithId[] memory streams = new StreamWithId[](streamCount);
+    for (uint256 i = 0; i < streamCount; ++i) {
+      bytes32 streamId = streamIds[i];
+      Stream storage stream = ds.streamById[streamId];
+      if (stream.nodes.length == 0) continue;
```
texuf, line 263: in solidity what happens if `stream` is null here? will it error or will it still resolve nodes.length? Is this something you worry about in solidity?

brianathere: It will return a zero length arrry. I'll add test coverage

giuseppecrj (approved):

sergekh2 (changes requested):
(removed comment, misunderstood use case for the change)

sergekh2:
Practically it's impossible to get all streams in one call, due to the returned data size limitation.

getPaginatedStreams should be used to get all stream like this:
https://github.com/river-build/river/blob/544c40a73439f13b136cf3f43e61f5d49a79839b/core/node/registries/river_registry_contract.go#L295

brianathere:
> Practically it's impossible to get all streams in one call, due to the returned data size limitation.
> 
> 
> 
> getPaginatedStreams should be used to get all stream on a specific block number like this:
> 
> https://github.com/river-build/river/blob/544c40a73439f13b136cf3f43e61f5d49a79839b/core/node/registries/river_registry_contract.go#L295 

This is a different use case where the notification service needs to lookup a small subset of specific streams.

sergekh2 (approved):

texuf (approved):

PR #620, opened 2024-08-05
Refactor spaceImage getter as getSpaceImage() async

tak-hntlabs:
Should not be doing an async decrypt inside applySnapshot. Timing issue when getting the spaceImage from the streamStateView_Space. `spaceImage` can be undefined while the payload is being decrypted.

Instead of doing the decryption inside `applySnapshot`, store the encrypted data only. The caller uses `await streamStateView_Space.spaceContent.getSpaceImage()` to get the decrypted chunked media.


packages/sdk/src/streamStateView_Space.ts
```diff
@@ -27,21 +27,13 @@ export type ParsedChannelProperties = {
 export class StreamStateView_Space extends StreamStateView_AbstractContent {
     readonly streamId: string
     readonly spaceChannelsMetadata = new Map<string, ParsedChannelProperties>()
-    private _spaceImage: ChunkedMedia | undefined
+    private spaceImage: ChunkedMedia | EncryptedData | undefined
```
texuf, line 30: I really don’t think we need a union here. How about two variables? One for the encrypted payload, one for the decrypted one.

tak-hntlabs: sure

packages/sdk/src/streamStateView_Space.ts
```diff
@@ -131,10 +112,20 @@ export class StreamStateView_Space extends StreamStateView_AbstractContent {
...
+            const spaceAddress = contractAddressFromSpaceId(this.streamId)
+            const context = spaceAddress.toLowerCase()
+            const plaintext = await decryptDerivedAESGCM(context, this.encryptedSpaceImage)
+
+            // Ensure the state has not changed during the async operation
+            if (this.encryptedSpaceImage === initialEncryptedData) {
+                this.spaceImage = ChunkedMedia.fromBinary(plaintext)
+                this.encryptedSpaceImage = undefined
+            }
+        }
```
texuf, line 127: This is almost perfect. However, now we have issues calling this multiple times in a row. Which, as you know, in react - is absolutely something that happens all the time.

There are two issues
1) in the normal case, calling multiple times in a row will kick off multiple calls to decryptDericedAESGM - this isn't terrible, just a big waste of cpu
2) if encryptedSpaceImage changes between calls, the first call will return undefined.

I don't think #2 is going to happen much, but if it does we should return an image instead of undefined (rendering the old image is better than a broken image)

tak-hntlabs: you might have seen the in-between commits. There is a guard `initialEncryptedData` that should prevent the first call from overriding newer calls.

tak-hntlabs: needed help from chatgpt. Changes:
- if multiple calls to `getSpaceImage()` are made while an decryption is running, then return the same promise if the `this.encryptedSpaceImage` is the same.
- if `this.encryptedSpaceImage` is updated while the decryption function is running, then add to the `decryptionQueue` so that the decryption requests are processed in order.

packages/sdk/src/streamStateView_Space.ts
```diff
@@ -150,6 +131,62 @@ export class StreamStateView_Space extends StreamStateView_AbstractContent {
...
+            this.decryptionQueue = this.decryptionQueue.then(async () => {
+                // Assign the decryption process to decryptionInProgress and return void to avoid ESLint warnings
+                this.decryptionInProgress = this.performDecryption(context, currentEncryptedImage)
+                return this.decryptionInProgress.then(() => undefined) // Ensure void return
+            })
+
+            await this.decryptionQueue
+
+            // Always return the last known good state of spaceImage, even if it wasn't updated
+            return this.spaceImage
```
texuf, line 164: you can delete line 163 and 164, you don't need duplicate returns

packages/sdk/src/streamStateView_Space.ts
```diff
@@ -150,6 +131,62 @@ export class StreamStateView_Space extends StreamStateView_AbstractContent {
...
+                this.encryptedSpaceImage = undefined
+            }
+
+            return this.spaceImage
+        } finally {
+            // Clear the in-progress promise once it resolves or rejects
+            this.decryptionInProgress = undefined
+        }
+    }
+
```
texuf, line 189: I'm not sure chat gpt is the best at this. Here what about this? https://github.com/river-build/river/pull/629/files

tak-hntlabs: first caller might get spaceImage undefined if it completed the decryption, but a call to applySnapshot had updated the value of encryptedSpaceImage during its await.

packages/sdk/src/streamStateView_Space.ts
tak-hntlabs: good eye. removed

packages/sdk/src/streamStateView_Space.ts
tak-hntlabs: the decryptionQueue ensures that decryption requests are handled sequentially. It prevents overlapping decryptions. More orderly execution at runtime. The caller should get a predictable result; not undefined.

texuf: > irst caller might get spaceImage undefined if it completed the decryption, but a call to applySnapshot had updated the value of encryptedSpaceImage during its await.
Pretty sure that's not the case. Can you look again.

texuf: > the decryptionQueue ensures that decryption requests are handled sequentially. It prevents overlapping decryptions. More orderly execution at runtime. The caller should get a predictable result; not undefined.

1) I don't think that's a problem
2) i don't think your queue is doing what you think it is. The way I read it, it's also doing overlapping decryptions.

texuf: your code will return an undefined space image for the first caller if the source is updated between calls.

tak-hntlabs:
close by mistake

texuf (approved):

PR #622, opened 2024-08-05
support pins in all related stream types

erikolsson:
if i'm not mistaken, we also need to check for dm+gdm channel payloads here since we now support pins in all 3 channel types — did not yet test this, just noticed it when going through the code :)


texuf (approved):
good catch

PR #623, opened 2024-08-05
core/node: add optel to connectrpc handlers

bas-vk:
Enable https://connectrpc.com/docs/go/observability/#enabling-opentelemetry-for-connect

This gives more open-telemetry details information about requests/streams.

### DIFF
--- a/core/node/rpc/server.go
+++ b/core/node/rpc/server.go
@@ -14,12 +14,9 @@ import (
 	"time"
 
 	"connectrpc.com/connect"
+	"connectrpc.com/otelconnect"
 	"github.com/ethereum/go-ethereum/common"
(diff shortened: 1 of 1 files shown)


core/node/rpc/server.go
```diff
@@ -571,6 +571,9 @@ func (s *Service) initHandlers() {
 	ii := []connect.Interceptor{}
 	if s.otelConnectIterceptor != nil {
 		ii = append(ii, s.otelConnectIterceptor)
+		if i, err := otelconnect.NewInterceptor(); err == nil {
```
sergekh2, line 574: It looks like this is redundant: otel connect interceptor is created and configured in otel.go and appended on the previous line?

PR #630, opened 2024-08-06
core/node/rpc: Use x/http2.Server instead of http.Server

bas-vk:
There seems to be a problem with the http.Server for http2 that could cause the stream to hang.
- https://github.com/golang/go/issues/42534

Proxy projects ([caddy](https://github.com/caddyserver/caddy/blob/master/modules/caddyhttp/app.go#L412)/[traefik](https://github.com/traefik/traefik/blob/bd93e224deb434b4cd8ffe5c85e52c8b0f54f57a/pkg/server/server_entrypoint_tcp.go#L678)) switched over to use `http2.Server` with success.


sergekh2:
Tracking issue #640

PR #633, opened 2024-08-06
rpc/sync: Add timeout when sending stream update to client

bas-vk:
Sending a stream update to client can block in the `stream.Send` method. This method doesn't accept a timeout and can hang indefinitely as we are experiencing on testnet. A proposed work around for a send with timeout is here: https://github.com/grpc/grpc-go/issues/1229#issuecomment-300938770

The drawback of that approach is that it starts a go-routine for each send message, adding overhead. This change is inspired by that but uses an unbuffered channel. The message to send to the client is wrapped in a reply object that also contains an error chan. One go-routine per sync operation is started that reads from this unbuffered channel, calls `stream.Send` and sends the resulting err to the reply.err channel. If the result of `stream.Send` is not available within a 5 seconds timeout the sync operation is cancelled. Returning from the SyncStreams method should (not able to reproduce locally, so not able to verify this) also exit the stuck `stream.Send`.

### DIFF
--- a/core/node/rpc/sync/operation.go
+++ b/core/node/rpc/sync/operation.go
@@ -33,6 +33,9 @@ type (
 		streamCache events.StreamCache
 		// nodeRegistry is used to get the remote remoteNode endpoint from a thisNodeAddress address
 		nodeRegistry nodes.NodeRegistry
+		// SendTimeout is the timeout when sending a stream update to the client. If it takes longer the sync operation
+		// is cancelled.
+		sendTimeout time.Duration
 	}
 
 	// subCommand represents a request to add or remove a stream and ping sync operation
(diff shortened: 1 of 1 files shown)


core/node/rpc/sync/operation.go
```diff
@@ -78,15 +78,66 @@ func NewStreamsSyncOperation(
...
+	select {
+	case sendQueue <- msg:
+		select {
+		case err := <-msg.err:
+			if !sendTimeout.Stop() {
+				<-sendTimeout.C
+			}
+			return err
+		case <-sendTimeout.C:
+			return RiverError(Err_DEADLINE_EXCEEDED, "timeout sending sync stream update to client")
```
clemire, line 125: Is there any information difference between this and the error on line 128 that would cause it to make sense to distinguish tme?

bas-vk: I don't think it's relevant, in both case the send operation is taking too long due to a problem deep in the stack. In the first case the work loop accepted the request (should always be the case because otherwise the previous msg was stuck and should have timed out) and in the second case the send operation took too long.

PR #635, opened 2024-08-06
tools/xchain/xchain-entitlementcheck-tool Initial commit

bas-vk:
Add xchain entitlement checker analysis tool. This tool can be used to retrieve requested entitlement check requests after a block or within a block range. It is added in a new directory called `tools` on the same level as core that enabled it to use a similar directory structure as `core/xchain` and it can have its own go.mod instead of using the `core/go.mod`.


tools/xchain/entitlement-checker-analysis/entitlement/generated/i_entitlement_checker.go
brianathere, line 1: Can this reference core/contracts/base

tools/xchain/entitlement-checker-analysis/entitlement/generated/i_entitlement_gated.go
brianathere, line 1: Can this reference core/contracts/base

brianathere (commented):
Can this reference core/contracts/base

bas-vk:
> Can this reference core/contracts/base

Yes, I have updated the module to use them.

PR #637, opened 2024-08-06
perf(contract): Optimizations and clean ups

shuhuiluo:
Replaced internal `IERC165` imports with those from OpenZeppelin, and renamed a local interface for better clarity. Additionally, refactored `IntrospectionHelper` contract initialization to improve code readability and consistency.

Refactor internal deploy functions to use memory-safe assembly and add explicit reverts on failure. Integrate `LibClone` for address prediction to improve clarity and reliability.

Refactor loops to improve performance by reducing redundant evaluations and increasing readability. Added `unchecked` blocks for vote counting to minimize gas usage. Streamlined the fetching of rule data directly within return statements.

Modified function parameters to use `calldata` over `memory` for better gas efficiency. This change impacts functions in both `Architect.sol` and `ArchitectBase.sol` enhancing performance by reducing the cost of memory allocation.

Changed the method to use `membership.settings` directly, significantly reducing code redundancy. Also ensured `feeRecipient` defaults to `msg.sender` if unset, maintaining intended functionality while improving code readability.

### DIFF
--- a/contracts/src/factory/facets/architect/Architect.sol
+++ b/contracts/src/factory/facets/architect/Architect.sol
@@ -40,7 +40,7 @@ contract Architect is
   //                            Space
   // =============================================================
   function createSpace(
-    SpaceInfo memory spaceInfo
+    SpaceInfo calldata spaceInfo
   ) external nonReentrant whenNotPaused returns (address) {
     return _createSpace(spaceInfo);
   }
--- a/contracts/src/utils/Factory.sol
+++ b/contracts/src/utils/Factory.sol
@@ -1,6 +1,8 @@
 // SPDX-License-Identifier: MIT
 pragma solidity ^0.8.23;
 
+import {LibClone} from "solady/utils/LibClone.sol";
+
 /**
  * @title Factory for arbitrary code deployment using the "CREATE" and "CREATE2" opcodes
  */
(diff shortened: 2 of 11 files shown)


contracts/src/utils/Factory.sol
```diff
@@ -35,34 +39,35 @@ abstract contract Factory {
...
-
-    if (deployment == address(0)) revert Factory__FailedDeployment();
   }
 
   /**
    * @notice calculate the _deployMetamorphicContract deployment address for a given salt
    * @param initCodeHash hash of contract initialization code
    * @param salt input for deterministic address calculation
-   * @return deployment address
+   * @return deployment deployment address
```
brianathere, line 57: double deployment?

brianathere: Is this to name the variable?

shuhuiluo: Yes. This is to suppress compiler warning.

brianathere (approved):
All of the changes look good. Would be ideal to split them into independent PR's in case something needs to be bisected or reverted later.

contracts/src/factory/facets/architect/Architect.sol
```diff
@@ -40,7 +40,7 @@ contract Architect is
   //                            Space
   // =============================================================
   function createSpace(
-    SpaceInfo memory spaceInfo
+    SpaceInfo calldata spaceInfo
```
jterzis, line 43: Nice improvement. It seems spaceInfo is read-only so we can use calldata. Curious why not declare spaceInfo arguments as `calldata` all the way down the stack to _createSpace in [ArchitectBase](https://github.com/river-build/river/blob/main/contracts/src/factory/facets/architect/ArchitectBase.sol#L35) ?

jterzis:
If not referencing a specific issue link, can you please add a PR description with the changes in the PR to help reviewers follow along ?

jterzis (approved):

contracts/src/factory/facets/architect/Architect.sol
shuhuiluo: The arguments to internal functions in the base contract were also changed to calldata.

shuhuiluo:
> If not referencing a specific issue link, can you please add a PR description with the changes in the PR to help reviewers follow along ?

@jterzis Added. To be fair the commit messages were generated with JetBrains AI and are quite accurate.

For future PRs, we may introduce CodeRabbit to auto-gen PR summaries and reviews.

PR #648, opened 2024-08-06
XChain ERC20, Custom IsEntitled

clemire:
E2E tests for ERC20, implementation and E2E tests for custom entitlements. (remaining: eth balance and 1155.)

## Summary by CodeRabbit

- **New Features**
  - Enhanced entitlement management to allow individual updates for multiple user addresses.
  - New utilities for testing ERC20 token interactions and custom entitlement checks.
  - Mock contracts for `MockCustomEntitlement` and `MockERC20` introduced to facilitate development and testing.

- **Bug Fixes**
  - Improved handling of entitlement checks in the application logic, ensuring accurate status retrieval.

- **Documentation**
  - Updated function signatures to clarify that multiple users can be processed, enhancing API usability.

- **Tests**
  - Expanded test suite with new cases for ERC20 token interactions and custom entitlement logic.
  - Refactored and organized existing tests for enhanced readability and maintainability.


packages/web3/src/TestGatingNFT.ts
```diff
@@ -278,3 +158,10 @@ export async function balanceOf(nftName: string, address: Address): Promise<numb
 
     return Number(balanceEncoded)
 }
+
+export const TestNFT = {
```
brianathere, line 162: Should this be TestERC721, to make room for TestERC1155, and MockERC20 be TestERC20?

brianathere (approved):

packages/web3/src/TestGatingNFT.ts
clemire: Seems reasonable

clemire:
@brianathere  can you PTAL? You saw before custom entitlement work was finished

packages/web3/src/entitlement.ts
```diff
@@ -685,8 +693,27 @@ async function evaluateERC20Operation(
     )
 }
 
+async function evaluateCustomEntitledOperation(
+    operation: CheckOperation,
+    controller: AbortController,
+    provider: ethers.providers.StaticJsonRpcProvider,
+    linkedWallets: string[],
+): Promise<EntitledWalletOrZeroAddress> {
```
texuf, line 701: this doesn't return `EntitledWalletOrZeroAddress`... it returns the first wallet in the linked wallets list. Also why did you choose to return the zero address instead of a boolean here?

packages/web3/src/TestGatingUtils.ts
```diff
@@ -0,0 +1,141 @@
...
+export async function deployContract(
+    contractName: string,
+    abi: Abi,
+    bytecode: Address, // bytecode is a hex string
+    constructorArgs: any[] = [],
+): Promise<Address> {
+    let retryCount = 0
+    let lastError: unknown
+    while (retryCount++ < 5) {
+        try {
```
texuf, line 85: there is a ton of copy pasta here. I'm 99.9% sure we fixed the errors by using the `const privateKey = generatePrivateKey()` here (how can nonce be too low if we create a new account for every call?).

Can you remove all the mutexes and retries and try catches and just see if this works without all the extra stuff? It it gets unstable make a bug and assign it to me and I'll put it all back.

texuf: Oh wait! you refactored the other stuff into this one. Sorry i missed that. Lets leave it!

packages/web3/src/TestGatingNFT.ts
```diff
@@ -233,7 +110,7 @@ export async function publicMint(nftName: string, toAddress: Address): Promise<n
     throw Error('No mint event found')
 }
 
-export async function burn(nftName: string, tokenId: number): Promise<void> {
+async function burn(nftName: string, tokenId: number): Promise<void> {
```
texuf, line 113: just a comment, i'm pretty sure all usages of createTestClient should use a throwaway account, but we can do that in a follow up pr.

texuf: * especially any usage of createTestClient that writes to a contract. Now I think we should fix this here.

packages/web3/src/entitlement.ts
clemire: `EntitledWalletOrZeroAddress` is a bit of a strange type - it is intended to be an address that is either valid or the zero address - the zero address is the placeholder that indicates there is no entitled wallet for this permission in the user's set of wallets, i.e. they are not entitled. In the first case I return the first wallet because the entitled call is checking all of the wallets and so we don't know which of the wallets in particular is entitled, although multiple wallets may also be entitled. But I can definitely refactor this so that we test each wallet independently and only return a wallet that passed on the contract.

texuf: Definitely don't think we should call these individually, that seems like a waste. Do we ever use the EntitledWallet return value anywhere?

packages/web3/src/TestGatingNFT.ts
clemire: How come?

packages/web3/src/entitlement.ts
clemire: Yes, I think that Evan uses it in `joinTown` and also `mintMembershipTransaction` on the towns client in harmony, but I'm not 100% sure what it's doing.

packages/web3/src/TestGatingNFT.ts
clemire: (refactored)

contracts/test/mocks/MockCustomEntitlement.sol
```diff
@@ -9,12 +9,19 @@ contract MockCustomEntitlement is ICustomEntitlement {
   constructor() {}
 
   function setEntitled(address[] memory user, bool userIsEntitled) external {
```
texuf, line 11: this was a bug yeah? Can you please change the parameter name to `users`?

clemire: ok fixed!

texuf (approved):

packages/web3/src/TestCustomEntitlement.ts
```diff
@@ -0,0 +1,115 @@
...
+    })
+
+    const receipt = await client.waitForTransactionReceipt({ hash: txnReceipt })
+    expect(receipt.status).toBe('success')
+    logger.log(
+        `Set custom entitlement to ${entitled} for users ${userAddresses.join(
+            ',',
+        )} for contract ${customEntitlementContractName}`,
+    )
+}
```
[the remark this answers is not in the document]
clemire: I think throwaway is our policy

Issue #640, opened 2024-08-06
hang on http2 write


sergekh2:
### Description

Server periodically hangs on http2 write with callstacks like this:

```
net/http.(*http2serverConn).writeDataFromHandler(0xc0008549c0, 0xc001301860, {0xc0015d2ffb, 0x15699, 0x1f005}, 0x0)  
net/http/h2_bundle.go:4917 +0x1ea  
net/http.(*http2responseWriterState).writeChunk(0xc001141c00, {0xc0015d2ffb, 0x15699, 0x1f005})  
net/http/h2_bundle.go:6506 +0x53c  
net/http.http2chunkWriter.Write({0xc000c38140?}, {0xc0015d2ffb?, 0x0?, 0x0?})  
net/http/h2_bundle.go:6379 +0x1d  
bufio.(*Writer).Write(0xc000c38140, {0xc0015d2000?, 0x7f34ff8a45b8?, 0x10?})  
bufio/bufio.go:679 +0xea  
net/http.(*http2responseWriter).write(0x0?, 0xc0018b6560?, {0xc0015d2000?, 0x4148e5?, 0x5?}, {0x0?, 0x877b01?})  
net/http/h2_bundle.go:6795 +0x111  
net/http.(*http2responseWriter).Write(0xc0007f2630?, {0xc0015d2000?, 0x87773f?, 0xc0018b65e8?})  
net/http/h2_bundle.go:6769 +0x2a  
connectrpc.com/connect.(*envelope).WriteTo(0xc000694090, {0x7f34b8831448?, 0xc0019970d0?})  
connectrpc.com/connect@v1.14.0/envelope.go:85 +0x155  
connectrpc.com/connect.writeSender.Send(...)  
connectrpc.com/connect@v1.14.0/duplex_http_call.go:343  
connectrpc.com/connect.(*envelopeWriter).write(0xc001377b30?, 0x6719fa?)  
connectrpc.com/connect@v1.14.0/envelope.go:210 +0x29  
connectrpc.com/connect.(*envelopeWriter).Write(0xc00002c198, 0xc000694078)  
connectrpc.com/connect@v1.14.0/envelope.go:165 +0x252  
connectrpc.com/connect.(*envelopeWriter).marshalAppend(0xc00002c198, {0x16a21e0, 0xc000cd8070}, {0x7f34b8a5a850, 0x26ec4c0})  
connectrpc.com/connect@v1.14.0/envelope.go:193 +0x285  
[...]

PR #661, opened 2024-08-07
Just build go 1.22.6 from our dockerfile (copied from oficial dockerfile)

sergekh2:
Upgrading go/libs to see if it helps with https://github.com/river-build/river/issues/640


brianathere:
👍

PR #668, opened 2024-08-07
stress: send a message with `n_clients/total clients` in channel

### LINKED ISSUE #654: Stress tests: better summary
### Description

I hate counting all the emojis by hand.

The root client sends the first message and sends several replies to it. 

Can the root client send an initial reply that's "0 of 40 clients checking in" and then edit that message throughout the execution with the count of the number of reactions the root message gets?

Reasoning: I don't want to edit the root message because our sdk doesn't automatically collapse edits yet and it would probably break things.

miguel-nascimento:
Closes #654

image

## Summary by CodeRabbit

- **New Features**
	- Enhanced messaging capabilities during chat stress tests, including real-time updates on connected clients.
	- Introduced statistics reporting to monitor client interactions and engagement in the chat.

- **Improvements**
	- Clarified variable naming for event IDs related to chat messages for better readability.
	- Expanded the `ChatConfig` interface, allowing for flexible event handling.

- **Bug Fixes**
	- Managed lifecycle of statistics reporting to ensure accurate performance monitoring during chat sessions.

### DIFF
--- a/packages/stress/src/mode/chat/kickoffChat.ts
+++ b/packages/stress/src/mode/chat/kickoffChat.ts
@@ -35,6 +35,7 @@ export async function kickoffChat(rootClient: StressClient, cfg: ChatConfig) {
         `Clients: 0/${cfg.clientsCount} 🤖`,
     )
 
+    cfg.kickoffMessageEventId = kickoffMessageEventId
     cfg.countClientsMessageEventId = countClientsMessageEventId
 
     const initialStats = {
--- a/packages/stress/src/mode/chat/kickoffChat.ts
+++ b/packages/stress/src/mode/chat/kickoffChat.ts
@@ -26,10 +26,16 @@ export async function kickoffChat(rootClient: StressClient, cfg: ChatConfig) {
     const shareKeysDuration = Date.now() - shareKeysStart
 
     logger.log('send message')
-    const { eventId } = await rootClient.sendMessage(
+    const { eventId: kickoffMessageEventId } = await rootClient.sendMessage(
         announceChannelId,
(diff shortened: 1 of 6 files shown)


packages/stress/src/mode/chat/root_chat.ts
```diff
@@ -127,6 +130,20 @@ export async function startStressChat(opts: {
...
+                chatConfig.kickoffMessageEventId,
+            )
+            await updateCountClients(
+                clients[0],
+                chatConfig.announceChannelId,
+                chatConfig.countClientsMessageEventId,
+                chatConfig.clientsCount,
+                reactionCount,
+            )
+        }
```
texuf, line 146: this only runs on process one after waiting for each span of process 0's clients. Its not going to report any clients that join after process 0 completes.
 
we should kick off an async task before this block that counts up the reactions in a loop with a reasonable sleep timeout and then cancel it when we stop the clients. Make sense? We can call it reportStats and add more stats to it as we get more sophisticated.

texuf: Approved in case you want to check it in and see things working, but we should iterate on it.

texuf (approved):

packages/stress/src/mode/chat/kickoffChat.ts
```diff
@@ -44,7 +51,7 @@ export async function kickoffChat(rootClient: StressClient, cfg: ChatConfig) {
         `System Info: ${makeCodeBlock(getSystemInfo())} Initial Stats: ${makeCodeBlock(
             initialStats,
         )}`,
-        { threadId: eventId },
+        { threadId: kickoffMessageEventId },
```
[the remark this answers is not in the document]
texuf: I'm happy to have the error thrown here, because it gets picked up in the logs

PR #676, opened 2024-08-08
Update CODEOWNERS

erikolsson:
moving myself into only the /src folder. I end up getting too many notifications for releases etc, but would still like to follow what's happening inside the SDK.


texuf (changes requested):
Please put everyone’s name. Each line replaces the previous, so only you will get notifications now.

texuf (approved):

PR #677, opened 2024-08-08
node/rpc/sync: supply a cause when sync streams operation exits

bas-vk:
Use `context.CancelWithCause` to pass a cause when the streams sync operation is cancelled.


core/node/rpc/service_test.go
```diff
@@ -4,7 +4,10 @@ import (
 	"bytes"
 	"context"
 	"crypto/tls"
+	"errors"
 	"fmt"
+	"github.com/river-build/river/core/node/base"
+	"go.uber.org/atomic"
```
sergekh2, line 10: It seems regular `sync\atomic` has `Bool`?
If this package is to be used, then `go mod tidy` should be run.

core/node/rpc/sync/client/local.go
```diff
@@ -100,7 +100,11 @@ func (s *localSyncer) OnUpdate(r *StreamAndCookie) {
 	default:
 		log := dlog.FromCtx(s.syncStreamCtx)
 		log.Error("Cancel client sync operation - client buffer full", "syncId", s.globalSyncOpID)
-		s.cancelGlobalSyncOp()
+
+		err := RiverError(Err_BUFFER_FULL, "Client sync subscription message channel is full").
+			Tag("syncOpId", s.globalSyncOpID).
```
sergekh2, line 105: using same tag name allows to group messages
```suggestion
			Tag("syncId", s.globalSyncOpID).
```

core/node/rpc/sync/client/local.go
```diff
@@ -100,7 +100,11 @@ func (s *localSyncer) OnUpdate(r *StreamAndCookie) {
 	default:
 		log := dlog.FromCtx(s.syncStreamCtx)
 		log.Error("Cancel client sync operation - client buffer full", "syncId", s.globalSyncOpID)
-		s.cancelGlobalSyncOp()
+
+		err := RiverError(Err_BUFFER_FULL, "Client sync subscription message channel is full").
```
sergekh2, line 104: suggestion: not to duplicate log and error with same data, it's possible to call .Log() on the error itself to log it.

core/node/rpc/sync/client/local.go
```diff
@@ -126,7 +130,11 @@ func (s *localSyncer) OnStreamSyncDown(streamID StreamId) {
...
+			Tag("syncOpId", s.globalSyncOpID).
```
sergekh2, line 135: same as above

sergekh2 (approved):

PR #678, opened 2024-08-08
Flip UseHttps to DisableHttps

### LINKED ISSUE #511: Flip UseHttps setting to DisableHttps
### Description

```markdown
Desired behavior should be default if setting is not set.
Also, fail to start if certificates fail to load.
```

jakzale:
Adds the following convenience methods to core.config.Config:
- UrlSchema, to return the protocol for accessing the node

In addition
- configuration files for environments were updated to use the disableHttps key, and
- rpc.serviceTester.getConfig was updated to keep disabling HTTPS by default.

closes #511.

## Summary by CodeRabbit

- **New Features**
	- Introduced `DisableHttps` configuration option, enhancing control over HTTPS settings.
	- Added methods for easier access to HTTPS state and schema.

- **Bug Fixes**
	- Improved clarity in error messages related to HTTPS status.

- **Refactor**
	- Streamlined URL construction for server logging and startup messages, centralizing logic for determining schema.

- **Chores**
	- Updated configuration files across multiple environments to reflect the new `DisableHttps` setting.


core/config/config.go
```diff
@@ -430,6 +430,23 @@ func (c *Config) Init() error {
 	return c.parseChains()
 }
 
+// Return the schema to use for accessing the node.
+func (c *Config) Schema() (s string) {
```
sergekh2, line 434: 1. This name is too generic: when reading code it seems like this is a schema for config itself. Maybe `HttpSchema` (don't have a better name really)

2. this code base generally avoids named return params, which get confusing really quickly. Really no need to use in this case - it's just one if with two returns.

core/config/config.go
```diff
@@ -430,6 +430,23 @@ func (c *Config) Init() error {
...
+	s = "https"
+	if c != nil && c.DisableHttps {
+		s = "http"
+	}
+	return
+}
+
+// Convenience method to check if HTTPS is enabled.
+func (c *Config) IsHttpsEnabled() bool {
```
sergekh2, line 443: lets drop this method - there are only 3 callsites, easy enough to just use the value.

sergekh2 (approved):

PR #679, opened 2024-08-08
xchain eth balance

clemire:
E2E xchain support for eth balance checking. Note the getBalance method on chains returns the balance of the native token so depending on the chain it may be MATIC (e.g. polygon), etc.

testing:
- unit tests in typescript, go
- xchain IT tests
- e2e client-based tests for channels (stream node and client entitlement checks) and spaces (enters through joinSpace and uses client checks and xchain service)

Also added a few new unit tests to round out rule data evaluation according to coderabbit's suggestion.

## Summary by CodeRabbit

- **New Features**
	- Introduced new `ETHBALANCE` and `NATIVE_COIN_BALANCE` operation types for entitlement checks, allowing evaluations based on Ethereum and native coin wallet balances.
	- Enhanced functionality for checking balances across linked wallets for both Ethereum and native coins.
	- Expanded the entitlement management system with new functions and constants for improved balance checks.

- **Bug Fixes**
	- Improved error handling during balance retrieval to prevent nil address flags.

- **Tests**
	- Significantly expanded test coverage for Ethereum and native coin balance checks, incorporating various scenarios and configurations.
	- Added comprehensive tests to validate entitlement checks based on both Ethereum and native coin balances.

- **Documentation**
	- Enhanced clarity and usability of the smart contract's ABI, with more explicit input and output structures.


core/xchain/entitlement/check_operation.go
```diff
@@ -127,6 +129,47 @@ func (e *Evaluator) evaluateIsEntitledOperation(
...
+
+		// Balance is a *big.Int
+		// Iteratively check if the total balance of evaluated wallets is greater than or equal to the
+		// threshold.
+		if op.Threshold.Sign() > 0 && total.Sign() > 0 && total.Cmp(op.Threshold) >= 0 {
+			return true, nil
+		}
+	}
+	return false, nil
+}
```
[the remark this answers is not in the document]
clemire: good point, I'll add those

packages/web3/src/entitlement.ts
```diff
@@ -718,6 +729,44 @@ async function evaluateCustomEntitledOperation(
...
+            } catch (error) {
+                return {
+                    wallet,
+                    balance: ethers.BigNumber.from(0),
+                }
+            }
+        }),
+    )
+
+    const walletsWithAsset = walletBalances.filter((balance) => balance.balance.gt(0))
```
texuf, line 755: nbd but why filter this out here? seems like an unnecessary iteration. Is there a problem calling  acc.add(0)?

texuf: oh i see. you're creating a condition that you depend on later.

texuf (approved):
nice

bas-vk:
If some of the supported chains don't use `ETH` as their native token should we rename `ETHBALANCE` to something like `NATIVE_COIN_BALANCE`?

core/xchain/client_simulator/client_simulator.go
```diff
@@ -315,23 +315,31 @@ func (cs *clientSimulator) executeCheck(ctx context.Context, ruleData *deploy.IR
...
-			)
-			if err != nil {
-				log.Error("Failed to get NewMockEntitlementGated", "err", err)
-				return nil, err
-			}
-			log.Info("NewMockEntitlementGated", "gated", gated.RequestEntitlementCheck, "err", err)
-			tx, err := gated.RequestEntitlementCheck(opts, big.NewInt(0), *ruleData)
-			log.Info("RequestEntitlementCheck called", "tx", tx, "err", err)
-			return tx, err
+			opts.GasLimit = max(gasEstimate, 2_500_000)
```
bas-vk, line 341: It is just the simulator but is this hard coded gas limit also required when requesting a entitlement check? I thought this was only required when posting the entitlement check result to compensate for different code paths in the smart contract when a majority vote is and isn't yet reached?

core/xchain/entitlement/check_operation.go
```diff
@@ -32,13 +32,26 @@ func (e *Evaluator) evaluateCheckOperation(
 	// Sanity checks
 	log := dlog.FromCtx(ctx).With("function", "evaluateCheckOperation")
 	if op.ChainID == nil {
-		log.Info("Chain ID is nil")
-		return false, fmt.Errorf("evaluateCheckOperation: Chain ID is nil")
+		log.Error("Entitlement check: chain ID is nil for operation", "operation", op.CheckType.String())
+		return false, fmt.Errorf("evaluateCheckOperation: Chain ID is nil for operation %v", op.CheckType.String())
```
bas-vk, line 36: You can use `%s` instead of `%v` and drop the call to `String()`.

core/xchain/entitlement/check_operation.go
```diff
@@ -127,6 +143,47 @@ func (e *Evaluator) evaluateIsEntitledOperation(
...
+			"balance", balance.String(),
+			"total", total.String(),
+			"threshold", op.Threshold.String(),
+			"chainID", op.ChainID.String(),
+		)
+
+		// Balance is a *big.Int
+		// Iteratively check if the total balance of evaluated wallets is greater than or equal to the
+		// threshold.
+		if op.Threshold.Sign() > 0 && total.Sign() > 0 && total.Cmp(op.Threshold) >= 0 {
```
bas-vk, line 180: Native coin balances can't be negative. So `op.Threshold` should always be >= 0, otherwise the request makes no sense. That also implies that a single check on `op.Threshold == 0` in the beginning of this function is enough. If it is 0 the result is a pass (if len(wallets) > 0). If it is negative the request is invalid because the coins native balances can't be negative.

core/xchain/entitlement/entitlement.go
```diff
@@ -62,6 +63,8 @@ func (t CheckOperationType) String() string {
 		return "ERC1155"
 	case ISENTITLED:
 		return "ISENTITLED"
+	case ETHBALANCE:
+		return "ETHBALANCE"
```
bas-vk, line 67: If some of the supported chains don't use ETH as their native token should we rename `ETHBALANCE` to something like `NATIVE_COIN_BALANCE`?

core/xchain/client_simulator/client_simulator.go
clemire: I think I can take this out actually, it was showing that the estimate actually reverted and this was due to an out of date deployment of the mock entitlement gated.

core/xchain/entitlement/entitlement.go
clemire: Yes, that's a good thing to highlight about decimals, but we don't actually convert the threshold according to the decimal value in this code, we compare threshold to the returned currency value directly, so I think that's more of a UI detail, to convert ETH to Wei, or whatever native currency we'd like to compare.

clemire: For native token balance I think your comment makes sense, but let me check with Fei about use case as a final gate.

core/xchain/entitlement/check_operation.go
clemire: Actually NM, to this PR I added a threshold check to all checktypes for threshold > 0 and consider threshold == 0 invalid, since that's basically a no-op, along with unit tests. I'll refactor the comparisons too.

Issue #647, opened 2024-08-06
Stress test investigation - Message Too Long


texuf:
### Description

Looks like decryption is failing in some cases... I but I don't understand why we're not seeing this elsewhere or in the app (maybe we are)

Hypothesis - this is a key fulfillment to a user inbox stream that encrypted way too many session keys...

At the very minimum please make this log a little better (include the stream id)

https://app.datadoghq.com/logs?query=env%3Agamma%20service%3Astress-test-node%20&agg_m=count&agg_m_source=base&agg_t=count&cols=env%2Cservice%2Ccontainer_index&event=AgAAAZEpV0ngtPEfQwAAAAAAAAAYAAAAAEFaRXBWMUlvQUFERmdDSHNyTTVkRndBQQAAACQAAAAAMDE5MTI5NWEtMmY0MS00N2U1LTk0ZTQtZDY0YWJkNjZmYWNk&fromUser=true&messageDisplay=inline&refresh_mode=sliding&saved-view-id=2705025&storage=hot&stream_sort=desc&view=spans&viz=stream&from_ts=1722972783766&to_ts=1722976383766&live=true

```
{
	"id": "AgAAAZEpV0ngtPEfQwAAAAAAAAAYAAAAAEFaRXBWMUlvQUFERmdDSHNyTTVkRndBQQAAACQAAAAAMDE5MTI5NWEtMmY0MS00N2U1LTk0ZTQtZDY0YWJkNjZmYWNk",
	"content": {
		"timestamp": "2024-08-06T20:17:52.096Z",
		"tags": [
			"managed_by:terraform",
			"source:cloudwatch",
			"env:gamma",
			"forwardername:datadogintegration-forwarderstack-rufay0-forwarder-hzhenfwrgmr6",
			"forwarder_version:3.71.0",
			"sourcecategory:aws",
			"forwarder_memorysize:1024",
			"container_index:0",
			"datadog.submission_auth:api_key"
		],
		"host": "/ecs/stress-test-node-gamma-0",
		"service": "stress-test-node",
[...]

PR #681, opened 2024-08-08
encryption: log streamId in "message too long" error

miguel-nascimento:
I wasnt able to reproduce #647 anymore, but if we even hit it again at least we will have something in the log to inspect

## Summary by CodeRabbit

- **New Features**
	- Enhanced logging capabilities for improved error reporting during encryption processes.
	- Added contextual information to error messages for payload length checks, improving traceability.
- **Bug Fixes**
	- Improved error handling by catching exceptions in the payload length check and logging detailed error messages.

### DIFF
--- a/packages/encryption/src/encryptionDevice.ts
+++ b/packages/encryption/src/encryptionDevice.ts
@@ -9,10 +9,10 @@ import {
 } from './encryptionTypes'
 import { EncryptionDelegate } from './encryptionDelegate'
 import { GROUP_ENCRYPTION_ALGORITHM, GroupEncryptionSession } from './olmLib'
-import { dlog } from '@river-build/dlog'
+import { dlogger } from '@river-build/dlog'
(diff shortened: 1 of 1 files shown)


packages/encryption/src/encryptionDevice.ts
```diff
@@ -566,8 +566,15 @@ export class EncryptionDevice {
     ): Promise<{ ciphertext: string; sessionId: string }> {
         return await this.cryptoStore.withGroupSessions(async () => {
             log(`encrypting msg with group session for stream id ${streamId}`)
-
-            checkPayloadLength(payloadString)
+            try {
+                checkPayloadLength(payloadString)
```
texuf, line 570: can you add an `opts: { streamId?: string, source: string }` param to checkPayloadLength instead of this try/catch that does a dlogError before throwing so that we get logs from all the different places?

miguel-nascimento: good idea, thank you

texuf (approved):

PR #686, opened 2024-08-09
sdk(sync-agent): user metadata support

miguel-nascimento:
This PR adds stream members and their metadata object: `ensAddress`, `nft`, `displayName` & `username`.

Each stream has a `Members` class, which itself its stored in the DB as a userId list.
This class holds a map of userId -> Member, and some utils like `isUsernameAvaliable`
We can lookup for a specific member by id, getting the Member object, which enable us to get member metadata.
We can lookup for ourselves using `member.myself` (name TBD). This allows us to get and set the metadata


packages/sdk/src/sync-agent/user/models/metadata/displayName.ts
```diff
@@ -0,0 +1,117 @@
...
+const logger = dlogger('csb:userSettings')
+
+type DisplayNameModel = {
+    isEncrypted: boolean
+    displayName: string
+}
+
+export interface UserMetadata_DisplayNameModel extends Identifiable {
+    id: string
+    streamId: string
```
miguel-nascimento, line 21: We need a way to let the client define the rules for the user metadata storage.

Towns for example doesn't care about channel specific display names, they only store display names inside space and dm streams, and not in channel streams. Would be nice to allow those rules.

packages/sdk/src/sync-agent/user/models/metadata/displayName.ts
```diff
@@ -0,0 +1,118 @@
...
+@persistedObservable({ tableName: 'UserMetadata_DisplayName' })
+export class UserMetadata_DisplayName extends PersistedObservable<UserMetadata_DisplayNameModel> {
+    constructor(
+        userId: string,
+        streamId: string,
+        store: Store,
+        private riverConnection: RiverConnection,
+    ) {
+        super(
+            { id: userId, streamId, initialized: false, displayNames: new Map() },
```
texuf, line 35: note - we want to use this in a member object which will exist on every stream, because the same user id can belong to multiple streams, the id should be `${streamId}_${userId}`

packages/sdk/src/sync-agent/user/models/metadata/displayName.ts
miguel-nascimento: later we figure it out

packages/sdk/src/sync-agent/members/members.ts
```diff
@@ -0,0 +1,88 @@
...
+                this.onStreamInitialized(this.data.id)
+            }
+            client.on('streamNewUserJoined', (streamId, userId) =>
+                this.onMemberJoin(streamId, userId),
+            )
+            client.on('streamNewUserInvited', (streamId, userId) =>
+                this.onMemberJoin(streamId, userId),
+            )
+            client.on('streamUserLeft', (streamId, userId) => this.onMemberLeave(streamId, userId))
+
```
texuf, line 38: for the time being we need to listen for streamInitialized event in case it wasn't there yet, or incase it gets re-initialized because sync was out of date. See `onStreamInitialized` space.ts

packages/sdk/src/sync-agent/members/members.ts
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    isUsernameAvailable(username: string): boolean {
+        const streamId = this.data.id
+        const streamView = this.riverConnection.client?.stream(streamId)?.view
+        check(isDefined(streamView), 'stream not found')
+        return streamView.getUserMetadata().usernames.cleartextUsernameAvailable(username)
+    }
+
+    private onStreamInitialized(streamId: string): void {
```
texuf, line 64: if you declare functions as lambdas here `private onStreamInitialized = (streamId) => ...` you don't need to capture them when you add them as event listeners

packages/sdk/src/sync-agent/members/models/metadata/displayName.ts
```diff
@@ -0,0 +1,107 @@
...
+        const oldState = this.data
+        this.setData({ displayName })
+        return this.riverConnection
+            .call(async (client) => {
+                check(isDefined(client.cryptoBackend), 'cryptoBackend is not defined')
+                const encryptedData = await client.cryptoBackend.encryptGroupEvent(
+                    streamId,
+                    displayName,
+                )
+                return client.makeEventAndAddToStream(
```
texuf, line 68: is there a reason you didn't want to use `client.setDisplayName`?

packages/sdk/src/sync-agent/members/models/metadata/displayName.ts
```diff
@@ -0,0 +1,107 @@
...
+            client.on('streamInitialized', this.onStreamInitialized)
+            client.on('streamDisplayNameUpdated', this.onStreamDisplayNameUpdated)
+            return () => {
+                client.off('streamInitialized', this.onStreamInitialized)
+                client.off('streamDisplayNameUpdated', this.onStreamDisplayNameUpdated)
+            }
+        })
+    }
+
+    async setDisplayName(displayName: string) {
```
texuf, line 57: One last question - for all of these, what happens if you call them before `this.data.initialized` is true? We need to handle this case.

miguel-nascimento:
```ts
test('update username', async () => {
    const userIds = space.members.data.userIds
    const metadata = space.members.getMember(bob.userId)
    const promise = metadata?.setUsername('bob123')
    expect(metadata).toBeDefined()

    await bob.start()
    await promise
    expect(userIds).toContain(bob.userId)
    expect(metadata?.username).toBe('bob123')
    await bob.stop()
})
```
This fails: `client.setDisplayName` check if the `client.stream(streamId)` is defined. 
We would need something like `riverConnection.callStream(streamId, () => T)`, which call if stream is defined or enqueue it

packages/sdk/src/sync-agent/members/models/metadata/displayName.ts
```diff
@@ -0,0 +1,95 @@
...
+        const streamId = this.data.streamId
+        const oldState = this.data
+        this.setData({ displayName })
+        return this.riverConnection
+            .call((client) => client.setDisplayName(streamId, displayName))
```
texuf, line 61:
```
.call((client) async => 
await client.waitForStream(streamId)
client.setDisplayName(streamId, displayName))
```

packages/sdk/src/streamStateView_Members_Membership.ts
```diff
@@ -40,6 +40,16 @@ export class StreamStateView_Members_Membership {
         }
     }
 
+    info(userId: string): MembershipOp {
+        const isJoined = this.joinedUsers.has(userId)
+        if (isJoined) return MembershipOp.SO_JOIN
+        const isInvited = this.invitedUsers.has(userId)
+        if (isInvited) return MembershipOp.SO_INVITE
+        const isLeft = !isJoined && !isInvited
+        if (isLeft) return MembershipOp.SO_LEAVE
```
texuf, line 49: can you use `readonly leftUsers = new Set<string>()` to return LEAVE? And handle UNSPECIFIED in the ui as well?

miguel-nascimento: > handle UNSPECIFIED in the ui as well?

What do you mean?

texuf: Sorry, not in this pr, I was just calling out that UNSPECIFIED is a valid state that is different than LEAVE. In our system if you have a LEAVE it means you recently left and we can put up a time and an event hash for that action. UNSPECIFIED means they either never joined or they left before our last snapshot

texuf (approved):

miguel-nascimento:
Lint and build works locally, but CI is refusing to build 🫠

packages/sdk/src/sync-agent/members/members.ts
```diff
@@ -0,0 +1,121 @@
...
+        return this.members[userId]
+    }
+
+    isUsernameAvailable(username: string): boolean {
+        const streamId = this.data.id
+        const streamView = this.riverConnection.client?.stream(streamId)?.view
+        check(isDefined(streamView), 'stream not found')
+        return streamView.getUserMetadata().usernames.cleartextUsernameAvailable(username)
+    }
+
```
texuf, line 79: I renamed getUserMetadata to getMemberMetadata - if you kill all local dev and run ./scripts/yarn-clean.sh you should see the error

texuf (approved):

PR #689, opened 2024-08-09
Set pprof stack labels for sync operations

sergekh2:


sergekh2:
This improves debuggability for #640

PR #704, opened 2024-08-10
Remove derived encryption as an option in the chunked media message

texuf:
We don’t need this value to be derived. The chunked media payload itself is encrypted with a derived key, we don’t need to make any changes around uploading media in our client.
(Event if it is derived, we derive secret and iv so we should just use those values in this payload)

### DIFF
--- a/packages/sdk/src/space.test.ts
+++ b/packages/sdk/src/space.test.ts
@@ -6,12 +6,11 @@ import { isEncryptedData, makeTestClient, makeUniqueSpaceStreamId, waitFor } fro
 import { Client } from './client'
 import { dlog } from '@river-build/dlog'
 import { AES_GCM_DERIVED_ALGORITHM } from '@river-build/encryption'
-import {
-    contractAddressFromSpaceId,
-    makeUniqueChannelStreamId,
-    makeUniqueMediaStreamId,
-} from './id'
-import { MediaInfo, MembershipOp } from '@river-build/proto'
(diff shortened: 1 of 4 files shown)


protocol/payloads.proto
```diff
@@ -178,15 +178,10 @@ message ChunkedMedia {
         bytes secretKey = 2;
     }
 
-    message DerivedAESGCM {
-        string context = 1;
-    }
```
tak-hntlabs, line 183: good insight. Makes sense. Like!

packages/sdk/src/space.test.ts
```diff
@@ -150,8 +148,18 @@ describe('spaceTests', () => {
...
+            info: image,
+            streamId: mediaStreamId,
+            encryption: {
+                case: 'aesgcm',
+                value: { secretKey: key, iv },
+            },
+            thumbnail: undefined,
+        } satisfies PlainMessage<ChunkedMedia>
+
+        await bobsClient.setSpaceImage(spaceId, chunkedMediaInfo)
```
tak-hntlabs, line 162: concern about having the caller derive the key+iv. That such a critical part of the derived-key protocol. If the caller gets it wrong, then a lot of flow breaks.

Simplifying the ChunkedMedia encryption back to only `aesgcm` makes a lot of sense. But the key+iv derivation should be inside `setSpaceImage` to avoid a lot of subsequent decryption issues.

tak-hntlabs (approved):

PR #713, opened 2024-08-11
perf(contract): refactor rule entitlement contracts

shuhuiluo:
Removed redundant struct to streamline the codebase. Simplified conditional checks and optimized loops for consistency. Updated entitlement storage from `memory` to `storage` where appropriate for efficiency.

Switch from `abi.decode` to direct `calldata` decoding using assembly for `RuleData` and `RuleDataV2`. This optimizes structs handling by reading directly from calldata, improving efficiency and reducing gas costs.

### DIFF
--- a/contracts/src/spaces/entitlements/rule/RuleEntitlement.sol
+++ b/contracts/src/spaces/entitlements/rule/RuleEntitlement.sol
@@ -36,12 +36,6 @@ contract RuleEntitlement is
 {
   using EnumerableSet for EnumerableSet.Bytes32Set;
 
-  struct Entitlement {
-    address grantedBy;
-    uint256 grantedTime;
-    RuleData data;
-  }
-
   mapping(uint256 => Entitlement) internal entitlementsByRoleId;
 
   address public SPACE_ADDRESS;
@@ -112,16 +106,12 @@ contract RuleEntitlement is
     uint256 roleId,
     bytes calldata entitlementData
   ) external onlySpace {
+    if (entitlementData.length == 0) return;
+
     // Decode the data
     RuleData memory data = abi.decode(entitlementData, (RuleData));
 
-    if (entitlementData.length == 0 || data.operations.length == 0) {
-      return;
-    }
-
-    // Cache sender and currentTime
-    address sender = _msgSender();
-    uint256 currentTime = block.timestamp;
+    if (data.operations.length == 0) return;
 
     // Cache lengths of operations arrays to reduce state access cost
     uint256 operationsLength = data.operations.length;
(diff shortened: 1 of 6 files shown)


contracts/src/spaces/entitlements/rule/RuleEntitlement.sol
```diff
@@ -112,24 +106,25 @@ contract RuleEntitlement is
...
-      return;
+    if (entitlementData.length == 0) return;
+
+    // equivalent: abi.decode(entitlementData, (RuleData))
+    RuleData calldata data;
+    assembly {
+      // this is a variable length struct, so calldataload(entitlementData.offset) contains the
+      // offset from entitlementData.offset at which the struct begins
+      data := add(entitlementData.offset, calldataload(entitlementData.offset))
     }
```
shuhuiluo, line 117: Ref: https://github.com/Uniswap/universal-router/pull/294

contracts/src/spaces/entitlements/rule/RuleEntitlement.sol
```diff
@@ -166,28 +161,14 @@ contract RuleEntitlement is
...
 
     Entitlement storage entitlement = entitlementsByRoleId[roleId];
```
shuhuiluo, line 163: Do we want to overwrite existing `Entitlement` or revert if the `roleId` has been written?

contracts/src/spaces/entitlements/rule/RuleEntitlement.sol
```diff
@@ -166,28 +161,14 @@ contract RuleEntitlement is
...
-    for (uint256 i = 0; i < logicalOperationsLength; i++) {
-      entitlement.data.logicalOperations.push(data.logicalOperations[i]);
-    }
-
-    for (uint256 i = 0; i < operationsLength; i++) {
-      entitlement.data.operations.push(data.operations[i]);
-    }
+    entitlement.grantedBy = _msgSender();
+    entitlement.grantedTime = block.timestamp;
+    entitlement.data = data;
```
brianathere, line 166: Does this now work because the copy as coming from call data,not memory?

contracts/src/spaces/entitlements/rule/RuleEntitlement.sol
brianathere: Overwrite so edit doesn't require a remove and a set call

brianathere (approved):

contracts/src/spaces/entitlements/rule/RuleEntitlement.sol
shuhuiluo: Yes but only when the dynamic struct array is nested.
https://github.com/ethereum/solidity/issues/14837

brianathere: Very cool, didn't know this

PR #716, opened 2024-08-12
perf(contract): Optimize node retrieval and random selection logic

shuhuiluo:
Refactor `getNodesByOperator` to use memory-safe assembly for updating the array length and improve loop efficiency. Enhance `_getRandomNodes` by using the more efficient unchecked block and streamline random selection and node updating logic.

### DIFF
--- a/contracts/src/base/registry/facets/checker/EntitlementChecker.sol
+++ b/contracts/src/base/registry/facets/checker/EntitlementChecker.sol
@@ -176,7 +176,9 @@ contract EntitlementChecker is IEntitlementChecker, Facet {
     for (uint256 i; i < totalNodeCount; ++i) {
       address node = layout.nodes.at(i);
       if (layout.operatorByNode[node] == operator) {
-        nodes[nodeCount++] = node;
+        unchecked {
+          nodes[nodeCount++] = node;
+        }
       }
     }
     assembly ("memory-safe") {
--- a/contracts/src/base/registry/facets/checker/EntitlementChecker.sol
+++ b/contracts/src/base/registry/facets/checker/EntitlementChecker.sol
@@ -163,32 +163,25 @@ contract EntitlementChecker is IEntitlementChecker, Facet {
   /**
    * @notice Get the nodes registered by an operator
    * @param operator The address of the operator
-   * @return An array of node addresses
+   * @return nodes An array of node addresses
    */
   function getNodesByOperator(
     address operator
-  ) external view returns (address[] memory) {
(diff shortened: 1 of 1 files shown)


contracts/src/base/registry/facets/checker/EntitlementChecker.sol
```diff
@@ -231,7 +228,7 @@ contract EntitlementChecker is IEntitlementChecker, Facet {
     return
       uint256(
         keccak256(
-          abi.encodePacked(block.prevrandao, block.timestamp, seed, msg.sender)
+          abi.encode(block.prevrandao, block.timestamp, seed, msg.sender)
```
giuseppecrj, line 231: why encode? we don't care abount maintaining type information on these values.

contracts/src/base/registry/facets/checker/EntitlementChecker.sol
```diff
@@ -163,32 +163,27 @@ contract EntitlementChecker is IEntitlementChecker, Facet {
...
-  ) external view returns (address[] memory) {
+  ) external view returns (address[] memory nodes) {
     EntitlementCheckerStorage.Layout storage layout = EntitlementCheckerStorage
       .layout();
     uint256 totalNodeCount = layout.nodes.length();
-    uint256 nodeCount = 0;
-    for (uint256 i = 0; i < totalNodeCount; i++) {
+    nodes = new address[](totalNodeCount);
+    uint256 nodeCount;
+    for (uint256 i; i < totalNodeCount; ++i) {
```
giuseppecrj, line 176: no need to use ++i vs i++ since newer compiler optimizes for this, can leave but don't want to go down the rabbit hole of over optimizing small things.

giuseppecrj (approved):

contracts/src/base/registry/facets/checker/EntitlementChecker.sol
shuhuiluo: Because `abi.encodePacked` involves bit shifting and is more expensive.

contracts/src/base/registry/facets/checker/EntitlementChecker.sol
shuhuiluo: Agree.

giuseppecrj:
need to create government proposal before merging @shuhuiluo

shuhuiluo:
> need to create government proposal before merging @shuhuiluo

Postponing tiny optimizations to next material upgrade.

PR #719, opened 2024-08-12
core: add otel support to tx pool

bas-vk:
This change adds 2 trace spans for transaction processing.
- `txpool_submit`, time it takes to create the transaction and submit it to the chain rpc node
- `pending_tx_wait`, describes how long the process was waiting for the transaction receipt

Example:
Screenshot 2024-08-12 at 14 20 27


core/node/crypto/chain_monitor_test.go
```diff
@@ -668,17 +673,22 @@ func TestEventsOrder(t *testing.T) {
...
 
 	// generate blocks until last tx is processed
-	var receipt *types.Receipt
-	for receipt == nil {
-		tc.Commit(ctx)
-		select {
-		case r := <-pendingTx.Wait():
-			receipt = r
-		default:
+	done := make(chan struct{})
```
clemire, line 676: Where is `done` passed so that we could expect it to be closed? I don't see it

core/node/crypto/chain_txpool.go
```diff
@@ -246,8 +253,25 @@ func NewTransactionPoolWithPolicies(
 	return txPool, nil
 }
 
-func (tx *txPoolPendingTransaction) Wait() <-chan *types.Receipt {
-	return tx.listener
+func (tx *txPoolPendingTransaction) Wait(ctx context.Context) (*types.Receipt, error) {
```
sergekh2, line 256: Would it also be useful to set error, if any, on the span?

sergekh2 (approved):

core/node/crypto/chain_txpool.go
bas-vk: In this case the only error that can occur is the result of the ctx being cancelled/expiring. The reason this function returns an error is to force callers to deal with this situation and prevent nil panics when the receipt is nil if such a thing happens. Sop it won't be very interesting to set the error here.

PR #724, opened 2024-08-12
dlog: support SINGLE_LINE_LOGS env

### LINKED ISSUE #722: DLog - After disabling mutliline we now log way too much data
### Description

Logs for this ticket are ~22MB instead of 100's of Kb https://linear.app/hnt-labs/issue/TOWNS-11109/%5Bomega%5D%5Buser-feedback%5D-everytime-i-open-towns-it-takes-me-to-same-town

I think after this change: https://github.com/river-build/river/pull/590 we no longer respect the "depth" property that was preventing us from logging entire objects.

I can think of a few quick fixes
- instead of looking at NODE_ENV we can make a new env var that we only set during stress tests 
- fix the code so that we still recurse the objects as before but we concatenate them at a later point in the logger
- re-write the logger with a better logging library (debug seems very limited)

Would be interested in hearing more. I think the highest priority is to get things working like they were before.

miguel-nascimento:
Change of behavior here. Previously, we would use multiline logs only in dev mode. Now, multiline logs are the default, and if you want single line ones, set using SINGLE_LINE_LOGS.

Currently, we want singe line logs on the stress test, that’s why I set on the Dockerfile.

This should close #722

### DIFF
--- a/packages/stress-testing/Dockerfile
+++ b/packages/stress-testing/Dockerfile
@@ -41,5 +41,6 @@ RUN yarn
 RUN yarn csb:build
 
 ENV MONOREPO_ROOT=/monorepo
+ENV SINGLE_LINE_LOGS="true"
 
 ENTRYPOINT [ "./packages/stress-testing/scripts/start-node.sh" ]
--- a/packages/dlog/src/dlog.ts
+++ b/packages/dlog/src/dlog.ts
@@ -140,7 +140,9 @@ export interface DLogOpts {
 
 const allDlogs: Map<string, DLogger> = new Map()
 
-const isDev = typeof process !== 'undefined' && process.env.NODE_ENV === 'development'
+// github#722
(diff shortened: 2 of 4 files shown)


packages/stress-testing/Dockerfile
```diff
@@ -41,5 +41,6 @@ RUN yarn
 RUN yarn csb:build
 
 ENV MONOREPO_ROOT=/monorepo
+ENV SINGLE_LINE_LOGS="true"
```
texuf, line 44: can you set this in `packages/stress/scripts/start.sh` next to `DEBUG_DEPTH`?

miguel-nascimento: sure

miguel-nascimento: do you want me to remove from the dockerfile?

texuf (approved):

PR #725, opened 2024-08-12
Revert "perf(contract): Optimize node retrieval and random selection …

giuseppecrj:
This reverts commit 33f59b8f2d351afbdd76e30af9cf584997db1845.


jterzis (approved):

shuhuiluo:
Do we need to revert even though we don't upgrade immediately given there's no business logic change?

PR #733, opened 2024-08-12
stress clients: Remove localhost hack for redis storage

texuf:
The param is REDIS_HOST because aws only provides a host, there is no protocol or port. Mimic the same thing for localhost and do the right thing.


miguel-nascimento (approved):
tricky!

PR #734, opened 2024-08-12
dlog: use winston on node

miguel-nascimento:


packages/dlog/src/dlog.ts
```diff
@@ -140,11 +142,57 @@ export interface DLogOpts {
...
+                      level: 'debug',
+                      filename: 'debug.log',
+                      format: format.combine(format.splat(), format.json()),
+                  }),
+                  new transports.File({
+                      filename: 'combined.log',
+                      format: format.combine(format.splat(), format.json()),
+                  }),
+              ]),
+    ],
```
miguel-nascimento, line 192: Should I add `transport.Console` here too? This way we can slowly migrate to file, but we can assure that it works with console first.

miguel-nascimento: ill do it

packages/dlog/src/tests/dlog.test.ts
```diff
@@ -59,7 +59,7 @@ describe('dlogTest', () => {
     })
 
     test('extend', () => {
-        const base_log = dlog('test:dlog')
+        const base_log = dlog('test:dlog', { useWinston: false })
```
texuf, line 62: do these tests pass with winston? if so you can do a matrix test
```
test.each([true, false]), (useWinston: boolean) => { ... })
```

miguel-nascimento: well, ...kinda?! 
we're using our dlog formatting function under the hood with winston, so it handles all the stuff: map, set, bignumbers, etc

but I wasn't able to get the output of winston as a string. And wasn't able to capture the stdout and run tests against it too. Got skill issue'd 😅

mechanical-turk:
@miguel-nascimento thanks a lot for taking a stab at this. i see a few things we need to address here

a) file transports will write to those files in production. i don't see a reason to do this on prod, especially given how we don't employ any patterns to access live file systems. in fact, this pattern has caused us to have runtime failures because we ran out of ephemeral storage. for these reasons, i think we should remove file logging.

b) our production o11y systems work better when logs are formatted in json. colorizing and splatting are meant for development environments that run inside a terminal. these formats confuse log aggregators since they don't run inside a terminal and have no concept of colorizing. for these reasons, we should either only do json logging, or selectively enable colorized logging based on the environment. take a look at the notification service to see how this can be done.

c) the main idea in dlog is namespacing log outputs at runtime. the second we use winston, we're basically defeating the purpose of dlog in the first place. using winston here implies that we're comfortable with not having to filter logs by namespaces for this service. since that's the case, we should drop dlog on this service altogether, and directly use the custom logger.

d) i have a separate PR with tak that attemps to establish pino as a canonical logger implementation that i intend to adopt in the rest of the codebase. it's totally fine to use winston in the time being but i'm just giving you a heads up.

miguel-nascimento:
@mechanical-turk thanks for the comment! 

a&b) Sure, let's drop file logging and only output to the console. I think that we still need splat since some usages include object logging, formatting, but for prod env we can output the log to console without colors and in a JSON format.

c) 100% agree. We need a way to run out of debug log. I added Winston inside debug log, so we can keep our existing code and still have stderr logs. I think this works for now, but it isn't the best solution.

d) @texuf would appreciate our thoughts about this one. Having better error logs would help in the stress investigations, but if that's not the priority: I'm down for closing this PR on behalf of the pino implementation.

texuf:
@miguel-nascimento @mechanical-turk sorry I didn't realize that Kerem already had a change in the works! Yeah lets put this on ice and see how Pino works out.

PR #740, opened 2024-08-13
Use deployment config and RiverRegistry shim from web3 package

tak-hntlabs:
- use what is already available in web3 to get the river config
- use the web3 shims instead of viem package to make contract calls
- remove code that will need fixing when eslint is enabled (next PR)


packages/stream-metadata/src/evmRpcClient.ts
```diff
@@ -1,44 +1,51 @@
...
+	if (!riverRegistry) {
+		throw new Error('Failed to create river registry')
 	}
+
+	return riverRegistry
 }
 
-export function getPublicClient() {
-	const chainId = config.chainId
+export function getRiverRegistry(chainId: number) {
```
texuf, line 43: this is weird - we don't have 1 river registry per chain, we have one river config per environment, which includes a chain id and contract addresses. You should just look at `config.river.chainId`

texuf (approved):
I don't think we need to pass the config around as parameters to these functions when it's just a global. If there's not a good reason I would recommend simplifying that too, but if you have a good reason it's fine.

tak-hntlabs:
> I don't think we need to pass the config around as parameters to these functions when it's just a global. If there's not a good reason I would recommend simplifying that too, but if you have a good reason it's fine.

I'd like to be able to write unit tests with different configs without having to set and reset a global config that might affect entire test suite.

PR #747, opened 2024-08-13
Log ping sync errors as errors

texuf:


brianathere (approved):

tak-hntlabs (approved):

texuf:
I think this should have gotten logged in the catch for this loop in `this.logError('syncLoop error', err)`, but honestly hard to tell.

PR #757, opened 2024-08-14
Reindex miniblock_candidates table with correct index

sergekh2:


sergekh2:
#751 already does this

PR #758, opened 2024-08-14
add health check endpoint to the stream-metadata service

tak-hntlabs:


texuf (approved):

texuf (approved):

texuf:
where are these static check errors coming from all of a sudden? Is it from the go upgrade last week?

packages/stream-metadata/src/handleHealthCheckRequest.ts
```diff
@@ -0,0 +1,32 @@
...
+		riverRegistry = getRiverRegistry(config)
+		if (riverRegistry) {
+			await riverRegistry.getAllNodes()
+			// healthy
+			return reply.code(200).send({ status: 'ok' })
+		}
+
+		// unhealthy
+		logger.error('Failed to get river registry')
+		return reply.code(500).send({ status: 'error' })
```
mechanical-turk, line 26: @tak-hntlabs the `@river-build/web3` library's `getWeb3Deployment` function takes care of getting the river registry, amongst many other things. i think having multiple paths that instantiate a river registry is harmless at best, and confusing at worst. it's an abstraction that we should probably remove. if you think you need to maintain the ability to create different river registries for whatever reason - which i personally think we don't but feel free to correct me - you can still use the `getWeb3Deployment` function.

i also have a nit here: riverRegistry can be initialized as a `const` within the try block

tak-hntlabs: Let's do a screenshare code walkthrough.  `getWeb3Deployment` just returns config info like the riverRegistry address. For health check, I'm making a contract call to the river registry using the @river/web3 RiverRegistry shim. The original code was using the `viem` package function `getContract`. Removed and replaced with `RiverRegistry` shim.

tak-hntlabs:
> where are these static check errors coming from all of a sudden? Is it from the go upgrade last week?

Probably. CI task for `Linting River Node` flagged a bunch of `go` errors even though this PR does not change any of the river node code. Going to create a separate PR to fix them.

texuf:
> > where are these static check errors coming from all of a sudden? Is it from the go upgrade last week?
> 
> Probably. CI task for `Linting River Node` flagged a bunch of `go` errors even though this PR does not change any of the river node code. Going to create a separate PR to fix them.

i hit the same thing https://github.com/river-build/river/pull/764

mechanical-turk (approved):

PR #762, opened 2024-08-14
Sync streams refactor - extract syncedStreamsLoop

texuf:
So that i can re-use it with different data structures.

Right now if the rpc client craps out, we have no way to shut down the syncedStreams loop and restart it without booting up a whole new client. This opens up that possibility


brianathere (approved):

mechanical-turk (approved):
@texuf this is cool. seems like a major stepping stone towards being able to run the sync loop in other services.

`userId` in syncloop appears to be only used for log namespacing, is that right? i suppose we could inject the namespace as a constructor dependency, and completely rid the syncloop of references to userId.

on a similar note, we could have SyncedStream handle events like `isMobileSafariBackgrounded` etc, and hook them into a more generalized interface on the syncloop that makes no assumptions about the runtime (e.g being in a browser, having access to the `document` global). that would be a pretty nice step towards making the sync loop reusable in other services like the notification service

texuf:
> @texuf this is cool. seems like a major stepping stone towards being able to run the sync loop in other services.
> 
> `userId` in syncloop appears to be only used for log namespacing, is that right? i suppose we could inject the namespace as a constructor dependency, and completely rid the syncloop of references to userId.
> 
> on a similar note, we could have SyncedStream handle events like `isMobileSafariBackgrounded` etc, and hook them into a more generalized interface on the syncloop that makes no assumptions about the runtime (e.g being in a browser, having access to the `document` global). that would be a pretty nice step towards making the sync loop reusable in other services like the notification service

Excellent feedback, I agree with both. I'll make the userId change, but i'll leave the safari mobile change for a follow up. It will require structural changes and i already touched a lot here.

PR #770, opened 2024-08-14
Make OR default when creating rule datas with multiple tokens or nfts

clemire:
along with options to specify logical op for creating ands if desired.


packages/web3/src/entitlement.ts
```diff
@@ -677,7 +680,7 @@ export function createOperationsTree(
             if (i + 1 < operations.length) {
                 newOperations.push({
                     opType: OperationType.LOGICAL,
-                    logicalType: LogicalOperationType.AND,
+                    logicalType: logicalOp as LogicalOperationType.AND | LogicalOperationType.OR,
```
texuf, line 683: is this right? I was expecting just `logicalType: logicalOp` here...

clemire: linter was complaining that you cannot assign type `LogicalOperation` to `LogicalOperation.AND | LogicalOperation.OR` - I think this is because `Operation` is an OR of `AndOperation and OrOperation` (and some other types), and each of them have an explicit value for their logical operation type?

clemire: This is the built-in linter with vs code, I can try with cli tool to confirm

clemire: `yarn build` fails locally if I remove this.

salzbrenner (approved):

PR #771, opened 2024-08-14
refactor stream-metadata server startup to improve startup orchestration, and fail fast if misconfigured

tak-hntlabs:


mechanical-turk (approved):
@tak-hntlabs thanks for doing this. 

by the way, i'm still not feeling great about the `void` workaround for the promise. how do you feel about moving the try catch logic out and into `main().then().catch()`?

Something along the lines of:

```
async function main() {
  await registerPlugins()
  setupRoutes()
  await startServer(config.port);
}

main().then(() => {
  logger.info(`Server listening at port: ${config.port}`);
}).catch(() => {
  logger.error('Error starting server', err)
  process.exit();
})
```

Also does `startServer`'s await actually resolve before `logger.info('Server started')` executes? i.e do you see this `Server started` log output when the server starts?

PR #775, opened 2024-08-15
react-sdk: member metadata

miguel-nascimento:
image


packages/react-sdk/src/useDisplayName.ts
```diff
@@ -0,0 +1,27 @@
...
+export const useDisplayName = (spaceId: string, userId: string) => {
+    const sync = useSyncAgent()
+    const member = useMemo(
+        () => sync.spaces.getSpace(spaceId).members.getMember(userId),
+        [sync, spaceId, userId],
+    )
+    const { data, ...rest } = useObservable(member?.observables.displayName)
+    return {
+        ...data,
+        ...rest,
```
miguel-nascimento, line 15: Here I'm spreading the observable return so we can get a good API like 
```ts
const { username } = useUsername(spaceId, userId)
```

If I don't spread, I would need to do 
```ts
const { data: { username } } = useUsername(spaceId, userId)
```
which doesn’t look good and formatters can format this into two lines, making it even worse

packages/react-sdk/src/useDisplayName.ts
```diff
@@ -0,0 +1,27 @@
+import { useMemo } from 'react'
+import { useSyncAgent } from './useSyncAgent'
+import { useObservable } from './useObservable'
+import { useAction } from './internals/useAction'
+
+export const useDisplayName = (spaceId: string, userId: string) => {
```
miguel-nascimento, line 6: I'm not handling per channel metadata, since we need a way to know if the dev wants the metadata from the channel or the metadata from the space

I'm thinking about the following:
```ts
useDisplayName(userId, spaceId, channelId, { priority: 'channel' | 'space' })
```

packages/react-sdk/src/useEnsAddress.ts
```diff
@@ -0,0 +1,27 @@
...
+}
+
+export const useSetEnsAddress = (spaceId: string) => {
+    const sync = useSyncAgent()
+    const member = useMemo(
+        () => sync.spaces.getSpace(spaceId).members.getMember(sync.userId),
+        [sync, spaceId],
+    )
+    const { action: setEnsAddress, ...rest } = useAction(member, 'setEnsAddress')
+    return { setEnsAddress, ...rest }
```
texuf, line 26: 1) I think we can probably figure out way to have useAction return an object with a key named how we want so we don't have to deconstruct and reconstruct this object, but otherwise i think we need useMemo again

packages/react-sdk/src/useObservable.ts
```diff
@@ -57,11 +57,17 @@ export function useObservable<T>(
...
+                : () => undefined,
+        () => observable?.value,
     )
 
     useEffect(() => {
+        if (!value) {
+            return
+        }
         if (isPersisted(value)) {
             if (value.status === 'loaded') {
```
texuf, line 72: i was surprised you didn't have to add a !value check below in the use memo here:
```
const data = useMemo(() => {
        if (isPersisted(value)) {
            const { data, status } = value
```

texuf (commented):
loving it!

packages/react-sdk/src/useObservable.ts
miguel-nascimento: Hm. Good catch, the `as ObservableValue<T>` type assert in the return hides the possible error here. 

I'll revisit the type implementation later, we need a better way to distinguish between `PersistedObservable` <> `Observable` in the type level

miguel-nascimento: Tried using function overloading. https://github.com/river-build/river/commit/51ba8f3c7dc84fd49ca80016cd9d1bd44ea99709
but it's not worth the implementation cost: makes hard to reason about, error messages gets 10x worse...

miguel-nascimento: https://github.com/river-build/river/pull/775/commits/ee99997338f90a3fd25e683d1d300d0272d78355 should fix it.

texuf: how did we get into a situation where the observable was undefined?

packages/react-sdk/src/useDisplayName.ts
texuf: can it be `useDisplayName(member: Member)` so we don't need to have optional observables? 
i realize like eventually it will probably be necessary, but we can kick it down the road.

miguel-nascimento: `useDisplayName(member: Member)` will raise the undefined check to the caller

```ts
const Component = () => {
   const bob = space.members.getMember(bobId)
   const { username } = useUsername(bob) ❌ 
   // bob can be undefined
}
```
```ts
const Component = () => {
   const bob = space.members.getMember(bobId)
   // ...
   return <>{bob && <MemberData member={bob} />}</>
}

const MemberData({ member } : { member: Member }) {
   const { username } = useUsername(member) ✅
   const { displayName } = useDisplayName(member) ✅
  // ...
}
```

I think this works for now, and I like the idea of passing a Member object, but it can cause some mistakes like calling `Member.displayName` directly, but I think it's fine for a v1

packages/react-sdk/src/useObservable.ts
miguel-nascimento:
```ts
Members.getMember(userId: string): Member | undefined
```
We're returning undefined because the `userId` may not be in the members Map
- the userId can be a malformed id
- the userId can represent a user that isn’t in the map

texuf (approved):

packages/react-sdk/src/useObservable.ts
texuf: Members.getMember should always return a Member
the member should have a `membership` param which should be 
image

If you are still loading the space the `isInitalized` should flag be false
If you request a member thats not in the space `membership` should be `SO_UNSPECIFIED`
If you are observing a member while they leave the value should flip to `SO_LEAVE`

miguel-nascimento: https://github.com/river-build/river/pull/686/commits/4a2a7b82fe203ae17b8eb3fe9e9d4c9c5d2826b5 solved this (in the `sync-user-metadata` branch)

texuf (approved):

PR #777, opened 2024-08-15
Fixup typescript in entitlement.ts

texuf:
1) logically group the type definitions
2) use typescript to infer the supported logical operations type so when we implement another we only need to add it in one place
3) fix up the other cast to be the of the supported type

### DIFF
--- a/packages/web3/src/entitlement.ts
+++ b/packages/web3/src/entitlement.ts
@@ -259,10 +259,10 @@ export function ruleDataToOperations(data: IRuleEntitlementBase.RuleDataStruct[]
             const logicalOperation = firstData.logicalOperations[operation.index]
             decodedOperations.push({
                 opType: OperationType.LOGICAL,
-                logicalType: logicalOperation.logOpType as SupportedLogicalOperationType,
+                logicalType: logicalOperation.logOpType,
                 leftOperation: decodedOperations[logicalOperation.leftOperationIndex],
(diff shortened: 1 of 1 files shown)


packages/web3/src/entitlement.ts
```diff
@@ -123,6 +122,8 @@ export const NoopRuleData = {
 type EntitledWalletOrZeroAddress = string
 
 export type LogicalOperation = OrOperation | AndOperation
+export type SupportedLogicalOperationType = LogicalOperation['logicalType']
+
 export type Operation = CheckOperation | OrOperation | AndOperation | NoOperation
```
clemire, line 127: Does it make sense to change this line to 

`export type Operation = CheckOperation | LogicalOperation | NoOperation`
so that we don't have to repeat `AndOperation | OrOperation` anywhere?

texuf: good suggestion, i'll try that and see how it looks

texuf: I looked into it and I'm going to call this a bad suggestion because it logically breaks the `isLogicalOperation` type guard which converts objects to LogicalOperation, since NoOperation is not a logical operation (it doesn't have a left and right op, and doesn't work in the tree)

The existing code doesn't repeat `AndOperation | OrOperation` anywhere so I don't think this is an issue.

clemire (approved):

PR #779, opened 2024-08-15
rename userDeviceKeys stream to userMetadata

texuf:
The purpose of this stream was a lighweight stream that other clients could sync to learn about you.
I think i was too restrictive naming it after device keys.
I think it makes sense to put the profile image pointer here (a chunked data info that contains a media stream id)

Renaming this should be backwards compatible since we’re using protobuf binary serialization.

wdyt?


core/node/shared/ids.go
```diff
@@ -123,7 +123,7 @@ func UserSettingStreamIdFromAddr(addr common.Address) StreamId {
 	return b
 }
 
-func UserDeviceKeyStreamIdFromAddress(addr common.Address) StreamId {
+func UserPublicDataStreamIdFromAddress(addr common.Address) StreamId {
 	var b StreamId
 	b[0] = STREAM_USER_DEVICE_KEY_BIN
```
bas-vk, line 128: Should we not also rename `STREAM_USER_DEVICE_KEY_BIN`?

texuf: yes good catch

texuf:
> not saying that it's better, but UserMetadata would be another alternative

I like that name too. I've only partially convinced myself that this is a good idea. We'll see what serge says tomorrow.

texuf:
> while we're at it... the `User` stream could be more specific. `UserMembership` for example!

I’m trying to make things less lexicographically restrictive! We need to rename the member payload to common or something generic.

sergekh2 (approved):

sergekh2 (approved):

miguel-nascimento (approved):

PR #782, opened 2024-08-15
Kerem/river packagejson cleanup

mechanical-turk:
This PR removes some old / unused root level package.json scripts.

 I'm removing `stream-metadata` from the list of things that `csb:build` builds. Plus, the`csb:build` is coded in a way that requires us to mentally keep track of the dependency graph between packages. We can lean on turbo to take care of that for us. Since `react-sdk`'s dependencies eventually resolve to all the packages that were previously listed in the script, we can just call `turbo build --filter @river-build/react-sdk`. And now that we see what this script actually does (i.e build the react sdk and all its dependencies), maybe we can take the opportunity to remove the `casablanca` reference here and refer to this script as something along the lines of`build:sdks`.

I'm also removing `build:metrics-discovery`. It was just an alias that was polluting the root package.json, and i think it's a bit cleaner this way. 

image


package.json
```diff
@@ -6,10 +6,8 @@
     "scripts": {
         "anvil": "yarn workspace @river-build/contracts exec anvil",
         "build": "turbo build",
-        "build:react": "turbo build --filter @river-build/react-sdk",
-        "build:metrics-discovery": "turbo build --filter @river-build/metrics-discovery",
         "cast": "yarn workspace @river-build/contracts exec cast",
-        "csb:build": "yarn workspace @river-build/proto run build && yarn workspace @river-build/dlog run build && yarn workspace @river-build/web3 run build && yarn workspace @river-build/encryption run build && yarn workspace @river-build/sdk run build && yarn workspace @river-build/react-sdk run build && yarn workspace stream-metadata run build && echo BUILD DONE || (echo BUILD ERROR; exit 1)",
+        "csb:build": "turbo build --filter @river-build/react-sdk",
```
texuf, line 10: i think you can make this just `@river-build/sdk` 
some history - the "Casablanca" server used to be in typescript, so this command would build the server, client sdk, and shared packages. We no longer have that dependency, so pretty sure all this needs to do is build protobufs and make sure there are no errors on the client side. 

the ci still builds everything right?

PR #675, opened 2024-08-08
refactor(sdk): refactor entitlement data ABI encoding and decoding

shuhuiluo:
As the title suggests.

PR #783, opened 2024-08-15
Cleanup unused entitlements functions, add comments

clemire:


packages/web3/src/ConvertersEntitlements.ts
```diff
@@ -1,25 +1,11 @@
 import { ethers } from 'ethers'
 
 import { Address, EntitlementStruct } from './ContractTypes'
-import { Hex, decodeAbiParameters, parseAbiParameters } from 'viem'
 import { encodeEntitlementData } from './entitlement'
 import { IRuleEntitlementBase } from './v3'
 
 const UserAddressesEncoding = 'address[]'
 
-export function decodeRuleData(encodedData: string): string[] {
```
clemire, line 10: Not used and also incorrect - uses UserEntitlement data type below.

packages/web3/src/entitlement.ts
```diff
@@ -18,6 +10,12 @@ import { MOCK_ADDRESS } from './Utils'
 
 const zeroAddress = ethers.constants.AddressZero
 
+// This public client is used purely to infer the typescript type of RuleData.
```
clemire, line 13: @texuf if you know a better way here? I took a quick look at ethers docs but nothing jumped out.

texuf: Can you try the ExtractAbi function that @shuhuiluo used here: https://github.com/river-build/river/pull/675/files

clemire: We actually don't need the function, just the type here. I found another approach, PTAL?

packages/web3/src/entitlement.ts
```diff
@@ -213,22 +179,6 @@ export function encodeEntitlementData(ruleData: IRuleEntitlementBase.RuleDataStr
...
-        })
-
-    if (!getRuleDataAbi) {
-        throw new Error('getRuleData ABI not found')
-    }
-    return decodeAbiParameters(
-        getRuleDataAbi.outputs,
-        entitlementData,
-    ) as unknown as IRuleEntitlementBase.RuleDataStruct[]
-}
```
texuf, line 230: looks like this one is used in tests ^

clemire: added back in

texuf (approved):

PR #784, opened 2024-08-15
Kerem/stream metadata dockerfile

mechanical-turk:
This PR introduces a Dockerfile for the stream-metadata service, as well as a github action that builds and publishes the image. The total image size comes down to `153 mb`, about 20mb above the base nodejs alpine image.

image

### DIFF
--- a/packages/stream-metadata/.env.local.sample
+++ b/packages/stream-metadata/.env.local.sample
@@ -1,9 +1,5 @@
 # Required.
-RIVER_ENV=alpha
-RIVER_CHAIN_RPC_URL=https://mainnet.rpc.river.build/http
+RIVER_ENV=local_multi
+RIVER_CHAIN_RPC_URL=http://localhost:8546
 # port that the server listens at. e.g. http://localhost:3002
 PORT=3002
-
-# IMPORTANT: disable tls cert validation in localhost env ONLY
-# in all other environments, comment out this line.
-NODE_TLS_REJECT_UNAUTHORIZED=0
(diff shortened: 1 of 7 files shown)


packages/stream-metadata/Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+RUN yarn run turbo build --filter @river-build/stream-metadata
+
+# create runner image with only the necessary files
+FROM node:lts-alpine3.20 as runner
+COPY --from=builder /river/packages/stream-metadata/dist /river/packages/stream-metadata/dist
+
+# run the service
+WORKDIR /river/packages/stream-metadata
+ENV NODE_ENV=production 
+CMD ["node", "./dist/node_esbuild.cjs"]
```
brianathere, line 47: Need to pass --enable-source-map to get it to actually pickup the sourcemaps. Not sure why, as the docs say it should be enabled by default, but my experience is the flag is required

texuf (approved):

packages/stream-metadata/.env.local.sample
```diff
@@ -2,8 +2,4 @@
 RIVER_ENV=local_multi
 RIVER_CHAIN_RPC_URL=http://localhost:8546
 # port that the server listens at. e.g. http://localhost:3002
-PORT=3002
-
-# IMPORTANT: disable tls cert validation in localhost env ONLY
-# in all other environments, comment out this line.
-NODE_TLS_REJECT_UNAUTHORIZED=0
```
tak-hntlabs, line 9: don't delete this for .env.local-sample. Otherwise, for local development, making a contract call to the local river chain will fail with 404 due to untrusted cert validation

tak-hntlabs (approved):

PR #785, opened 2024-08-15
Stream node, xchain changes for 1155 Contract Upgrades

clemire:
Edits to go code to bring xchain and stream node servers up to date with erc1155-related contract upgrades.

- Refactors to move shared code for marshalling entitlements into a common location for xchain, node.
- extended marshalling code to unmarshall RuleEntitlementV2s
- convert entitlement evaluation library to take a V2 rule data an added code to convert V1 to V2 rule datas
- update MockEntitlementGate to have the IEntitlementDataQueryable method for retrieving xchain entitlements and updated xchain to use that interface instead of `getRuleData`
- Go code for encoding / decoding v2 rule data params

stream node changes are not tested yet, but the stream node uses all of the same shared logic for evaluating entitlements as xchain so when I add V2 tests, I expect them to "just work".


core/contracts/types/entitlement_data.go
```diff
@@ -0,0 +1,254 @@
...
+	"strings"
+
+	"github.com/ethereum/go-ethereum/accounts/abi"
+	"github.com/ethereum/go-ethereum/common"
+
+	"github.com/river-build/river/core/contracts/base"
+	"github.com/river-build/river/core/node/dlog"
+)
+
+type Entitlement struct {
```
clemire, line 17: migrated from node/space_contract_v3.go b/c shared with xchain now

brianathere (approved):

core/contracts/types/entitlement_types.go
```diff
@@ -0,0 +1,230 @@
...
+			}
+		} else {
+			return nil, errors.New("unknown logical operation type")
+		}
+		log.Debug("Decoded operation", "operation", operation, "decodedOperations", decodedOperations)
+	}
+
+	var stack []Operation
+
+	for _, op := range decodedOperations {
```
texuf, line 201: in typescript we have a while here right? https://github.com/river-build/river/blob/928170dd252f4d3a876bb2fab285b856ec3aa06a/packages/web3/src/entitlement.ts#L650

then we constantly overwrite the stack variable with the updated tree until it only has one root.

tbh I haven't been able to run that code in my head yet - I don't really get how the serialized data fits together with the operations tree, but I'm pretty sure this code is different and wanted to call that out. Good test cases should be able to uncover discrepancies.

PR #787, opened 2024-08-15
fix pino logger printing

tak-hntlabs:
my logger statements only print the 1st argument strings, and ate up the subsequent arguments. Turns out that pino logger expects the arguments in the reverse order. First argument has to be the object; the 2nd argument is the message string.

**does not work (2nd arg eaten):**
```typescript
const obj = { fruit: 'apple' }
logger.info('print this', obj) // output: "print this"
logger.info(spaceContractAddress, 'the space contract address') // output: 0x12345
```

**this works:**
```typescript
const obj = { fruit: 'apple' }
logger.info(obj, 'print this')
// output: "print this"
//                    fruit: "apple"

logger.info({ spaceContractAddress }, 'the space contract address')
// output: "the space contract address"
                        spaceContractAddress: 0x12345
```

### DIFF
--- a/packages/stream-metadata/src/node.ts
+++ b/packages/stream-metadata/src/node.ts
@@ -16,13 +16,16 @@ process.title = 'stream-metadata'
 
 const logger = getLogger('server')
 
-logger.info({
-	riverEnv: config.riverEnv,
-	chainId: config.web3Config.river.chainId,
-	port: config.port,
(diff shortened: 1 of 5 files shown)


mechanical-turk (approved):

texuf (approved):
That’s weird. Kinda makes you want to wrap the pino logger in a class with the expected api.

PR #793, opened 2024-08-16
Remove extra verification target for base verificaiton

jterzis:
Only way I was able to get contract verification working was to remove what looked like an extra verification target in the `verify-base-*` target that due to the fact it didn't inject an api key, failed to verify on basescan. Succeeded verifying PrepayFacet after this pr:

```
➜  contracts git:(pat/prepay) ✗ make verify-base-sepolia contract=PrepayFacet address=0x90a6492d6ebFCFAEAb3884b6669686bD9e98c2Ab
Verifying PrepayFacet on a remote network, with verifier https://api-sepolia.basescan.org/api, api IGIKEVA7U3XB6YGFE9Q4UGB9SFGPJ2BE3Y
Start verifying contract `0x90a6492d6ebFCFAEAb3884b6669686bD9e98c2Ab` deployed on mainnet

Contract [contracts/src/spaces/facets/prepay/PrepayFacet.sol:PrepayFacet] "0x90a6492d6ebFCFAEAb3884b6669686bD9e98c2Ab" is already verified. Skipping verification.
```

### DIFF
--- a/contracts/makefile
+++ b/contracts/makefile
@@ -78,7 +78,7 @@ fork-any :;
 
 verify-any :;
 	@echo "Verifying $(contract) on a remote network"
-	@forge verify-contract ${address} ${contract}	--rpc-url ${rpc}
+	@forge verify-contract ${address} ${contract} --rpc-url ${rpc}
 
 explicit-verify-any :;
 	@echo "Verifying $(contract) on a remote network"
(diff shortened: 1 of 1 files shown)


giuseppecrj:
we need to verify on both basescan and blockscout

giuseppecrj (changes requested):
do not remove extra verification. we need to verify on both basescan and blockscout

giuseppecrj:
basescan seems to be down rn

jterzis:
> do not remove extra verification. we need to verify on both basescan and blockscout

This is currently broken for verification. If we need to verify both basescan and blockscout, then let's use 2 explicit-verify-any calls passing in the appropriate verification url and api key as the default verification, which `verify-any` invoked, uses etherscan and fails since there's no api key provided to the call. 

Thoughts @giuseppecrj ? 

```
➜  contracts git:(jt/fix-verify-base-targets) forge verify-contract --help
...
Verifier options:
      --verifier <VERIFIER>
          The contract verification provider to use
          
          [default: etherscan]
          [possible values: etherscan, sourcify, blockscout, oklink]

      --verifier-url <VERIFIER_URL>
          The verifier URL, if using a custom provider
          
          [env: VERIFIER_URL=]
```

PR #796, opened 2024-08-16
add interact alpha script

giuseppecrj:


brianathere (approved):

contracts/makefile
```diff
@@ -22,7 +22,7 @@ lint :; yarn solhint "{scripts,src,test}/**/*.sol"
 
 anvil :; anvil -m 'test test test test test test test test test test test junk'
 
-clear-anvil-deployments :; rm -rf deployments/alpha/base_anvil deployments/base_anvil deployments/river_anvil
+clear-anvil-deployments :; rm -rf deployments/local_multi
```
texuf, line 25: there are two local deployments, `local_multi` and `local_multi_de` (for disable entitlements). 
the plan is to re-write the sdk tests to run on entitled backends, but that's a chore and will happen once the sdk is more mature.

texuf: Tests cover both paths, so if tests are working it must be doing the setup right, i'm just weary of the hard coded env here.

.gitignore
```diff
@@ -47,6 +47,7 @@ tsconfig.tsbuildinfo
 /contracts/deployments/river_anvil
 /contracts/deployments/alpha/base_anvil
 /contracts/deployments/alpha/river_anvil
+/contracts/deployments/local_multi
```
texuf, line 50: there are other places too, but you'll definitely want it in the git ignore. the two scripts that get called locally are 

```
./scripts/configure-nodes.sh --multi
```
 and
```
./scripts/configure-nodes.sh --multi_ne
```

which means its `local_multi_ne` not de! sorry!

giuseppecrj: fixed in latest push, will take a context and remove that folder

PR #799, opened 2024-08-16
chore(syncAgent) refactor the connection loop and qol improvements

texuf:
This was pulled from the shared lib as is to get it to work, but it was confusing as all get out. Flatten out a few of the conditions and remove some variables. I’m sure we can simplify this further


miguel-nascimento (approved):

packages/sdk/src/sync-agent/syncAgent.ts
```diff
@@ -88,11 +105,18 @@ export class SyncAgent {
     }
 
     async start() {
+        if (this.stopped) {
+            throw new Error('SyncAgent is stopped, please instantiate a new sync agent')
+        }
```
miguel-nascimento, line 110: thank you!

PR #803, opened 2024-08-16
Upgrade client to support V1/V2 space creation

clemire:
- typescript entitlements evaluated against rule data v2s. rule data v1s are converted to v2 before being sent to the evaluation library.
- duplicate spaceDapp endpoints for creating spaces, roles, and for updating roles.
- encoded rule entitlement data now wrapped in a `VersionedRuleData` type that will be exposed in harmony

Testing:
- USE_LEGACY_SPACES to run entitlement tests with legacy space creation endpoints using mock architect
  - New CI job runs these tests
- some small new tests here and there


.github/workflows/ci.yml
```diff
@@ -400,6 +400,9 @@ jobs:
             - name: Build & Test (with entitlements)
               run: yarn csb:turbo-no-sdk
 
+            - name: Run River Tests Against Legacy Spaces (with entitlements)
+              run: yarn workspace @river-build/sdk run test:ci:multi:ent
```
mechanical-turk, line 404: @clemire did you mean to use `test:ci:multi:ent:legacy` here?

clemire: I did - this is a new suite that uses the legacy space creation endpoint on the space diamond and runs all of our entitlement tests. The existing line now runs against the new space creation endpoint. This test suite validates backwards compatibility with V1 spaces.

clemire: I think this could significantly increase how long it take the multinode_ent to run though so I'm willing / happy to split this out into a different job for engineer workflow reasons if you're concerned

mechanical-turk: @clemire yes but you're not using it :) you're using `test:ci:multi:ent:legacy`

clemire: ahhhhh ok thanks for noticing!

clemire: ...I put it in a separate job @mechanical-turk because it was taking > 30min to finish

brianathere (approved):

core/contracts/types/entitlement_data.go
```diff
@@ -155,9 +143,29 @@ func MarshalEntitlement(
 			EntitlementType:   rawEntitlement.EntitlementType,
 			RuleEntitlementV2: &ruleData,
 		}, nil
-	} else {
-		return Entitlement{}, fmt.Errorf("invalid entitlement type '%s'", rawEntitlement.EntitlementType)
 	}
+
+	if rawEntitlement.EntitlementType == ModuleTypeUserEntitlement {
```
texuf, line 148: was there a reason to kill the else/if continuity here? I realize it's technically the same code, but it's harder to maintain and harder to reason about flow if there are discrete blocks of code as apposed to one control block. 

If you don't have a good reason i suggest adding it back.

core/node/auth/auth_impl.go
```diff
@@ -561,7 +560,6 @@ func (ca *chainAuth) evaluateWithEntitlements(
 
 	// 1. Check if the user is the space owner
 	// Space owner has su over all space operations.
-	log.Info("evaluateWithEntitlements", "args", args, "owner", owner.Hex(), "wallets", args.linkedWallets)
```
texuf, line 564: thanks for cleaning up these noisy logs

texuf (approved):
i left some comments, but overall looks great. thank you.

clemire:
> i left some comments, but overall looks great. thank you.

YW!

core/contracts/types/entitlement_data.go
clemire: I think it just helped me fix the indentation from all of my copy pasta and I left it that way thinking it was good either way... happy to convert it back!

core/contracts/types/entitlement_data.go
```diff
@@ -155,9 +141,27 @@ func MarshalEntitlement(
 			EntitlementType:   rawEntitlement.EntitlementType,
 			RuleEntitlementV2: &ruleData,
 		}, nil
-	} else {
-		return Entitlement{}, fmt.Errorf("invalid entitlement type '%s'", rawEntitlement.EntitlementType)
+	} else if rawEntitlement.EntitlementType == ModuleTypeUserEntitlement {
+		abiDef := `[{"name":"getAddresses","outputs":[{"type":"address[]","name":"out"}],"constant":true,"payable":false,"type":"function"}]`
```
tak-hntlabs, line 145: is`ModuleTypeUserEntitlement` "legacy"? There is an order of precedence here. If EntitlementV2 is absent, then check and evaluate UserEntitlement. Would be good to add a comment so that someone less familiar with the code won't accidentally move UserEntitlement before EntitlementV2.

packages/sdk/src/util.test.ts
```diff
@@ -439,25 +454,123 @@ export async function createSpaceAndDefaultChannel(
...
+            wallet,
+        )
+    } else {
+        if (isLegacyMembershipType(membership)) {
+            // Convert legacy space params to current space params
+            membership = {
+                settings: membership.settings,
+                permissions: membership.permissions,
+                requirements: {
+                    everyone: true,
```
tak-hntlabs, line 481: should this be `everyone: await membership.requirements.everyone` ?

packages/sdk/src/util.test.ts
```diff
@@ -439,25 +454,123 @@ export async function createSpaceAndDefaultChannel(
...
+            // Convert legacy space params to current space params
+            createSpaceParams = {
+                spaceName: createSpaceParams.spaceName,
+                uri: createSpaceParams.uri,
+                channelName: createSpaceParams.channelName,
+                membership: {
+                    settings: createSpaceParams.membership.settings,
+                    permissions: createSpaceParams.membership.permissions,
+                    requirements: {
+                        everyone: true,
```
tak-hntlabs, line 526: is `everyone` always true in V2? Or should this be the value of the `createSpaceParams.membership.requirements.everyone` ?

tak-hntlabs (approved):

core/contracts/types/entitlement_data.go
clemire: There's actually no order of precedence here - this could be handled just as well with a switch statement. I can make that change if you'd like.

packages/sdk/src/util.test.ts
clemire: no `await` - but yes, you're right.

packages/sdk/src/util.test.ts
clemire: it should!

PR #804, opened 2024-08-16
update deployment paths for contracts

giuseppecrj:


contracts/scripts/interactions/InteractAlpha.s.sol
```diff
@@ -28,41 +28,42 @@ contract InteractAlpha is Interaction, DiamondHelper, IDiamondLoupeBase {
   DeploySpaceOwner deploySpaceOwner = new DeploySpaceOwner();
 
   function __interact(address deployer) internal override {
+    vm.setEnv("OVERRIDE_DEPLOYMENTS", "1");
     address space = getDeployment("space");
-    address spaceOwner = getDeployment("spaceOwner");
-    address spaceFactory = getDeployment("spaceFactory");
-    address baseRegistry = getDeployment("baseRegistry");
+    // address spaceOwner = getDeployment("spaceOwner");
```
jterzis, line 33: It would help reviewing this PR if this change were in a separate PR. Also, can we add a comment explaining why we're omitting the 4 facets from the upgrade for now ?

jterzis (approved):

contracts/scripts/interactions/InteractAlpha.s.sol
giuseppecrj: CI is slow, wanted to avoid multiple PRs taking twice the time to get verified. Can update or add comment in a separate PR

PR #805, opened 2024-08-16
removed parameter in setupRoutes

mechanical-turk:
line 77 had a reference to `server` directly whereas the other routes used the `srv` param. this pr attempts to simplify and standardize

### DIFF
--- a/packages/stream-metadata/src/node.ts
+++ b/packages/stream-metadata/src/node.ts
@@ -49,16 +49,16 @@ async function registerPlugins() {
 	logger.info('CORS registered successfully')
 }
 
-export function setupRoutes(srv: Server) {
+function setupRoutes() {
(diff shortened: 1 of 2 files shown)


packages/stream-metadata/src/node.ts
```diff
@@ -49,24 +49,24 @@ async function registerPlugins() {
 	logger.info('CORS registered successfully')
 }
 
-export function setupRoutes(srv: Server) {
+function setupRoutes() {
 	/*
 	 * Routes
 	 */
-	srv.get('/health', async (request, reply) => {
+	server.get('/health', async (request, reply) => {
```
tak-hntlabs, line 56: This is intentional. I want to pass in the server instance for testing.

PR #806, opened 2024-08-16
Scope deployments deletion to current environment for deploy script when running local dev

clemire:


scripts/deploy-contracts.sh
```diff
@@ -21,7 +21,7 @@ if [ "${SKIP_CHAIN_WAIT}" != "true" ]; then
     ./scripts/wait-for-riverchain.sh
 fi
 
-rm -rf contracts/deployments/*
```
clemire, line 24: only delete contracts/deployments/local_multi, etc

.gitignore
```diff
@@ -47,7 +47,7 @@ tsconfig.tsbuildinfo
 /contracts/deployments/river_anvil
 /contracts/deployments/alpha/base_anvil
 /contracts/deployments/alpha/river_anvil
-/contracts/deployments/local_multi
+/contracts/deployments/local*
```
clemire, line 50: For convenience, in case we add more

giuseppecrj (approved):

PR #826, opened 2024-08-20
Deploy space owner facet to omega

giuseppecrj:


contracts/makefile
```diff
@@ -136,7 +136,7 @@ deploy-base-anvil :;
 
 # THIS IS FOR TESTING ONLY
 deploy-base-anvil-ledger :;
-	@$(MAKE) deploy-any-local-ledger context=local_multi rpc=base_anvil hd_path="m/44\'/60\'/1\'/0/0" sender=${PRODUCTION_ADDRESS}
+	@$(MAKE) deploy-any-local-ledger context=local_multi rpc=base_anvil hd_path="m/44\'/60\'/1\'/0/0" sender=${SENDER_ADDRESS}
```
jterzis, line 139: This hd_path historically has not worked for me. My wallet uses a different path than the above. I've had to remove the path and supply `--ledger` argument, which prompts for a sig on my hardware device.

giuseppecrj: right, I'm siding towards just allowing the path to be defined through an env variable

PR #833, opened 2024-08-20
update jq path

giuseppecrj:


texuf (commented):
There's also one in `scripts/set-riverchain-config.sh`

texuf:
have we deleted the generated one yet? if so can you update 
CODEOWNERS
.prettierIgnore
the DOCKERFILE in metrics-discovery
the comment in riverConfig.ts
the DOCKERFILE and Readme in stream-metadata
the comments in start.sh
`scripts/node-registration-lib.sh`
`scripts/set-riverchain-config.sh`

giuseppecrj:
@texuf I updated the PR based on your comments, can you take another look when you have a chance.

package.json
```diff
@@ -46,7 +46,8 @@
         "test": "turbo test",
         "test:unit": "turbo test:unit -- --silent",
         "test:react": "vitest --project @river-build/react-sdk",
-        "test:build": "turbo test:build"
+        "test:build": "turbo test:build",
+        "make-config": "node ./scripts/make-config.js"
```
texuf, line 50: this is in the build because we include the file in the web3 package. it's not the best, the localhost addresses _should_ get pushed in via .env files, but that's not how it works now.

PR #846, opened 2024-08-22
user profile image

miguel-nascimento:
This PR adds support for user profile images
Notable changes:
- accept the event in the node
- allow the creation of a media stream only with `userId` (without `channelId`/`spaceId`)
- create event from the sdk, encrypting the media
- decrypt it in the sdk stream view

Should close:
- TOWNS-11297
- TOWNS-11298
- TOWNS-11299


packages/sdk/src/client.ts
```diff
@@ -688,32 +690,44 @@ export class Client
...
-
+        const userStreamId = userId ? makeUserStreamId(userId) : undefined
+        this.logCall('createMedia', channelId ?? spaceId ?? userStreamId, streamId)
         const inceptionEvent = await makeEvent(
             this.signerContext,
             make_MediaPayload_Inception({
                 streamId: streamIdAsBytes(streamId),
                 channelId: channelId ? streamIdAsBytes(channelId) : undefined,
                 spaceId: spaceId ? streamIdAsBytes(spaceId) : undefined,
+                userId: userStreamId ? streamIdAsBytes(userStreamId) : undefined,
```
miguel-nascimento, line 730: When doing some local tests, I found that if I pass `userId` in bytes here, that would be used to create the stream and, of course, would fail. 

Assuming that all the node work is done, and the implementation is correct, the node is expecting a stream_id from this field, so this should be called `userStreamId` instead?

I added a few TODO's in this PR related to this, I'll clean it later.

texuf: We want to pass userId. I messed up in our demo, we want to use the `requireUserAddr` check instead

core/node/rules/can_create_stream.go
```diff
@@ -233,6 +233,7 @@ func (ru *csParams) canCreateStream() ruleBuilderCS {
 				ru.checkMediaInceptionPayload,
 			).
 			requireMembership(
+				inception.UserId,
 				inception.ChannelId,
 				inception.SpaceId,
 			).
```
texuf, line 239: you're right, UserId is not a stream, we want to use `requireUserAddr(inception.UserId)` instead, though `requireUserAddr` does not have a check for nil, you'll need to add that if you don't want to break the other stream types.

texuf:
I think we might have missed this in the space pr too, we need to update snapshot.go to add these to the snapshot

packages/sdk/src/streamStateView_UserMetadata.ts
```diff
@@ -60,6 +68,9 @@ export class StreamStateView_UserMetadata extends StreamStateView_AbstractConten
             case 'encryptionDevice':
                 this.addUserDeviceKey(payload.content.value, encryptionEmitter, stateEmitter)
                 break
```
texuf, line 70: we also need to update `applySnapshot` here

miguel-nascimento: good catch!

We also need to publish the new profile image event right? If so
Q: do I need to publish the event in the encryption emitter too? Or only in the state emitter?

texuf: state emitter - they are actually the same emitter, just two logical groupings of events.

packages/sdk/src/streamStateView_UserMetadata.ts
```diff
@@ -5,17 +5,25 @@ import {
...
+import { decryptDerivedAESGCM } from './crypto_utils'
 
 export class StreamStateView_UserMetadata extends StreamStateView_AbstractContent {
     readonly streamId: string
     readonly streamCreatorId: string
+    private profileImage: ChunkedMedia | undefined
+    private encryptedProfileImage: EncryptedData | undefined
+    private decryptionInProgress:
+        | { encryptedData: EncryptedData; promise: Promise<ChunkedMedia | undefined> }
+        | undefined
```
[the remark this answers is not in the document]
texuf: That client space image decryption wasn’t used. We deleted it. If we do this a third time we should D.R.Y. it

packages/sdk/src/streamStateView_UserMetadata.ts
```diff
@@ -5,17 +5,25 @@ import {
...
 import { logNever } from './check'
 import { UserDevice } from '@river-build/encryption'
 import { StreamEncryptionEvents, StreamStateEvents } from './streamEvents'
 import { getUserIdFromStreamId } from './id'
+import { decryptDerivedAESGCM } from './crypto_utils'
 
 export class StreamStateView_UserMetadata extends StreamStateView_AbstractContent {
     readonly streamId: string
     readonly streamCreatorId: string
+    private profileImage: ChunkedMedia | undefined
```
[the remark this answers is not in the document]
texuf: yes, we only want to decrypt it once. this is the pattern we settled on for space image as well.

texuf (approved):

PR #850, opened 2024-08-22
fix start-stream-metadata-service script and yarn package.json script to support tests

tak-hntlabs:
My test PR depends on starting the stream-metadata service in the `local_mulit_ne` env. Pulling this change out as a separate PR.
Once merged, will pull the scripts over to harmony, and fix up the dev scripts in harmony.


scripts/start-stream-metadata-service.sh
```diff
@@ -1,10 +1,30 @@
 #!/bin/bash
 
+# Initialize variables
+RIVER_ENV=""
+RIVER_ROOT=""
```
mechanical-turk, line 5: @tak-hntlabs there was already a lot going on in this file, and it's getting more crowded. it's also polluting the root scripts directory. i suggest making the following changes: 

1. move all of this logic under `packages/stream-metadata/dev.sh`
2. set the "dev" script of stream-metadata/package.json to run `./scripts/dev.sh`
3. remove all references to RIVER_ROOT. whoever used to run `./river/scripts/start-stream-metadata-service.sh` should now run `yarn workspace @river/stream-metadata dev`. this will make it straightforward to run this script regardless of where you're running it from.
4. instead of having `RIVER_ENV` set by a command line option, have it set through an environment variable. Do not initialize it to an empty string at line 4. Just let the caller pass it as an environment variable. That should allow you to remove another 10-15 lines from this.

packages/stream-metadata/package.json
```diff
@@ -9,7 +9,10 @@
 		"build": "node ./esbuild.config.mjs",
 		"lint": "eslint . --ext .ts",
 		"lint:fix": "eslint . --ext .ts --fix",
-		"dev": "yarn build && node ./dist/node_esbuild.cjs",
+		"dev": "yarn dev:local_multi",
+		"dev:local_multi": "yarn run-build",
+		"dev:local_multi_ne": "RIVER_ENV=local_multi_ne PORT=3003 yarn run-build",
+		"run-build": "dotenv -e .env.local yarn build && node ./dist/node_esbuild.cjs",
```
mechanical-turk, line 15: @tak-hntlabs dotenv should have no impact here. you can remove everything before `yarn build`. we should have no env dependencies at compile time. if you want your runtime to depend on .env files, you should handle that in the source code.

i'm saying this while assuming that you actually don't have any explicit dependencies to the environment at build time. if you do, we should remove them. this isn't frontend code like vite, so we can build once and run in multiple environments without having to rebuild

tak-hntlabs: tested and removed dotenv-cli

scripts/start-stream-metadata-service.sh
tak-hntlabs: great discussions, @mechanical-turk - simplified.

mechanical-turk (approved):

PR #854, opened 2024-08-22
Stress - prevent multiple instanciations of members in the sync agent

texuf:


miguel-nascimento (approved):
thanks for the catch 🙏

PR #857, opened 2024-08-23
bio in stream user metadata

miguel-nascimento:
Should close:
- TOWNS-11406
- TOWNS-11407
- TOWNS-11409


packages/sdk/src/client.ts
```diff
@@ -940,6 +941,33 @@ export class Client
         return this.stream(streamId)?.view.userMetadataContent.getProfileImage()
     }
 
+    async setUserBio(bio: string) {
```
texuf, line 944: Since we're changing this, let's make a new "UserBio" message in payloads.proto and accept that as a parameter. I think we'll probably be moving default display names and default usernames here as well and it will suck to migrate from just a string later.

Sorry for not speccing this out the first time.

PR #858, opened 2024-08-23
use cross-chain entitlements for reactions

erikolsson:
this PR makes sure that we also check across chains for rule entitlements involving reactions, not just read/write.

see related code in `SpaceDapp.ts`:
https://github.com/river-build/river/blob/eae805df5e45d9a8cac21e5a1475ab9ea306aaca/packages/web3/src/v3/SpaceDapp.ts#L732-L736


clemire (approved):

clemire:
One thing to note: I'll be making a similar change soon to add PIN to the list of permissions evaluated this way.

PR #868, opened 2024-08-25
core: bump go dependency version

bas-vk:
Upgrade dependencies for sync issues and in anticipating for the upgrade to go 1.23 that requires an indirect dependency to be upgraded due to restrictions introduced on `go:linkname` in 1.23.


brianathere (approved):
I'm 100% for keeping dependencies up to date. It's hard to ensure nothing will regress with so many at once. Can we break this into a few steps so we can bisect if anything starts failing?

PR #875, opened 2024-08-26
remove child logger instantiation from lower-level functions

tak-hntlabs:
I didn't find these child logger instantiations useful in the DataDog logs. Removed.


tak-hntlabs:
> @tak-hntlabs these are necessary in the event of a critical failure. they will help a lot when we need to quickly see which module is logging what. this is especially useful when two modules can log the same message. this may not be happening right now, but it might happen in the future. they don't overcrowd logs, since the only thing they do is adding an extra metadata field to the log and leaving the message untouched. in the absence of these labels, all logs will appear to be coming from the base fastify logger - which we sometimes want to exclude from all logs while searching for something specific.
> 
> the tl;dr is that i strongly advise against removing them

i don't believe so because the top-level request handlers like fetchSpaceImage, fetchMetadata are still creating the childLogger that identifies the entry points. And that flows down to all the lower-level functions. There is enough context.
The issue is that all these lower-level function child loggers create un-necessary code maintenance. There was a PR that renamed one of the lower-level functions. But did not update the childLogger string. So the childLogger string still use the stale function name.

texuf (approved):

mechanical-turk (approved):

PR #877, opened 2024-08-26
sdk(sync-agent): flatten member onStreamInitialized 

miguel-nascimento:
Previously, all member metadata models were listening to the stream initialized event, causing unneeded subscriptions.

`Membership` model was listening to `streamNewUserJoined`, which is an unneeded listening, since this model lives in the `Member` model, which lives in the `Members` model, which gets the `streamNewUserJoined` too to update the member state.

### DIFF
--- a/packages/sdk/src/sync-agent/members/models/metadata/username.ts
+++ b/packages/sdk/src/sync-agent/members/models/metadata/username.ts
@@ -16,6 +16,8 @@ export interface MemberUsernameModel extends Identifiable {
     isUsernameEncrypted: boolean
 }
 
+// This model doenst listen to events here.
+// They are listened in the members model, which propagates the updates to this model.
 @persistedObservable({ tableName: 'member_username' })
 export class MemberUsername extends PersistedObservable<MemberUsernameModel> {
     constructor(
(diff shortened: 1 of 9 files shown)


packages/sdk/src/sync-agent/members/models/metadata/username.ts
```diff
@@ -40,24 +40,16 @@ export class MemberUsername extends PersistedObservable<MemberUsernameModel> {
...
-                this.onStreamInitialized(this.data.id)
-            }
-            client.on('streamInitialized', this.onStreamInitialized)
             client.on('streamUsernameUpdated', this.onStreamUsernameUpdated)
             client.on('streamPendingUsernameUpdated', this.onStreamUsernameUpdated)
             return () => {
-                client.off('streamInitialized', this.onStreamInitialized)
                 client.off('streamUsernameUpdated', this.onStreamUsernameUpdated)
                 client.off('streamPendingUsernameUpdated', this.onStreamUsernameUpdated)
             }
```
texuf, line 48: I think we need to push all of these from the members class

texuf (commented):
can you also change client.setMaxListeners(5000) back to 100?

texuf (approved):

PR #880, opened 2024-08-26
Tak/towns 10954 stream metadata handle metadata json response

tak-hntlabs:
- implements the /space/:spaceAddress endpoint
- reads the spaceInfo
- returns the json metadata (name, description, image (if set).


packages/stream-metadata/src/contract-utils.ts
```diff
@@ -0,0 +1,14 @@
...
+
+let spaceDapp: SpaceDapp | undefined
+
+export function getSpaceDapp(): SpaceDapp {
+	if (!spaceDapp) {
+		const provider = new ethers.providers.StaticJsonRpcProvider(config.baseChainRpcUrl)
+		spaceDapp = new SpaceDapp(config.web3Config.base, provider)
+	}
+	return spaceDapp
+}
```
mechanical-turk, line 14: @tak-hntlabs what's the reason for the indirection? why not

```
export const spaceDapp = new SpaceDapp(config.web3Config.base, provider);
```

packages/stream-metadata/src/routes/spaceMetadata.ts
```diff
@@ -1,36 +1,138 @@
...
 	request: FastifyRequest,
 	reply: FastifyReply,
 	serverUrl: string,
 ) {
 	const logger = request.log.child({ name: fetchSpaceMetadata.name })
 	const { spaceAddress } = request.params as { spaceAddress?: string }
-	logger.info({ spaceAddress }, 'GET /space/../metadata')
 
 	if (!spaceAddress) {
+		logger.error('spaceAddress parameter is required')
```
mechanical-turk, line 27: @tak-hntlabs this is perfectly fine and expected from the service's perspective, so it should be `.warn` or `.info` even. it would be fine to log an error on the clients that interface with this endpoint.

packages/stream-metadata/src/routes/spaceMetadata.ts
```diff
@@ -1,36 +1,138 @@
...
+async function getImageUrl(logger: FastifyBaseLogger, contractUri: string, spaceAddress: string) {
+	const hasSpaceImageExist = await hasSpaceImage(logger, spaceAddress)
+	if (!hasSpaceImageExist) {
+		return undefined
+	}
+
+	const isDefaultPort =
+		config.riverStreamMetadataBaseUrl.port === '' ||
+		config.riverStreamMetadataBaseUrl.port === '80' ||
+		config.riverStreamMetadataBaseUrl.port === '443'
```
texuf, line 87: Why are you checking ports here? Why are building an image url here at all?

texuf: Is this implementing an opensea spec? if so can you link it in the pr description?

packages/stream-metadata/src/contract-utils.ts
tak-hntlabs: i was thinking of letting this indirection handle baseChainRpc connection failure due to temporary network issue. Subsequent calls can get a new spaceDapp instance once the network recovers. But I don't have enough code here to handle the network issues properly. I'll simplify it for now.

packages/stream-metadata/src/routes/spaceMetadata.ts
tak-hntlabs: no this is for localhost dev only. I'll simplify it by checking config.riverEnv instead.

packages/stream-metadata/src/routes/spaceMetadata.ts
```diff
@@ -1,36 +1,138 @@
...
+
+	// Check if contractUri is empty or starts with the config.riverStreamMetadataHostUrl
```
mechanical-turk, line 89: @tak-hntlabs this entire logic should be contained within `config.baseUrl`. if the app wants their urls to be displayed with a port, the env var `RIVER_BASE_URL` should be set to something like `http://localhost:1234`. no further internal processing should be necessary

packages/stream-metadata/src/routes/spaceMetadata.ts
tak-hntlabs: fixed

packages/stream-metadata/src/routes/spaceMetadata.ts
tak-hntlabs: fixed

packages/stream-metadata/src/environment.ts
```diff
@@ -27,10 +27,15 @@ function makeConfig() {
...
+	const baseUrl = new URL(env.RIVER_STREAM_METADATA_BASE_URL)
 	if (env.RIVER_ENV.startsWith('local')) {
-		riverStreamMetadataBaseUrl.port = env.PORT.toString()
+		baseUrl.port = env.PORT.toString()
 	}
+	// Ensure no trailing slash when converting to string
+	const riverStreamMetadataBaseUrl = `${baseUrl.origin}${baseUrl.pathname.replace(
+		/\/$/,
+		'',
+	)}`.toLowerCase()
```
mechanical-turk, line 38: @tak-hntlabs `baseUrl.origin` by itself should do the trick.

also ideally the source code doesn't deal with any of this, and it just takes the base url as an environment variable - localhost or not

tak-hntlabs: no it doesn't. The test caught the bug. `.origin` adds a trailing / - which is why I cut it out.

mechanical-turk: @tak-hntlabs the issue you're seeing is not coming from `origin`, it's coming from `pathname`. `origin` does not even have a concept of a path to begin with, let alone a trailing slash. you can replace lines 35-38 with the following:

`const riverStreamMetadataBaseUrl` = baseUrl.origin;

For reference:
image

That will get the job done, but i also think the job you're trying to accomplish itself does not belong here. The environment sets the BASE_URL. if the BASE_URL should not have a trailing slash, we can just make it so by setting the correct environment variable. I surmise you're trying to cover development edge cases here but it's polluting the code we're shipping to prod. you can just set the env var in your package.json scripts and call it

packages/stream-metadata/src/environment.ts
```diff
@@ -26,12 +27,17 @@ function makeConfig() {
...
+	if (env.RIVER_ENV.startsWith('local')) {
+		baseUrl.port = env.PORT.toString()
+	}
 
 	return {
 		web3Config,
 		riverEnv: env.RIVER_ENV,
 		baseChainRpcUrl: env.BASE_CHAIN_RPC_URL,
 		riverChainRpcUrl: env.RIVER_CHAIN_RPC_URL,
+		riverSpaceStreamBaseUrl: `${baseUrl.origin}/space`.toLowerCase(),
```
mechanical-turk, line 40: same comments apply here. dev environments should set the proper BASE_URL and call it

packages/stream-metadata/src/environment.ts
tak-hntlabs: are you looking at the latest?

packages/stream-metadata/src/environment.ts
tak-hntlabs: talked offline

packages/stream-metadata/src/routes/spaceMetadata.ts
```diff
@@ -1,36 +1,128 @@
...
+	}
+
+	return longDescription || ''
+}
+
+async function getImageUrl(logger: FastifyBaseLogger, contractUri: string, spaceAddress: string) {
+	const hasSpaceImageExist = await hasSpaceImage(logger, spaceAddress)
+	if (!hasSpaceImageExist) {
+		return undefined
+	}
```
texuf, line 84: we're allowing people to overwrite the contract uri right? so we can't return undefined here. I'm not sure why we need to check if it's set at all, seems like we should handle not found when you call the /image url. 
what exactly are you trying to achieve?

tak-hntlabs: good point. Will remove

mechanical-turk (approved):

packages/stream-metadata/src/routes/spaceMetadata.ts
```diff
@@ -1,36 +1,89 @@
...
+
+	// Not using the default space image service
+	// redirect to the space contract's uri
+	return reply.redirect(spaceInfo.uri)
+}
+
+function getSpaceDecription({ shortDescription, longDescription }: SpaceInfo): string {
+	if (shortDescription && longDescription) {
+		return `${shortDescription}<br><br>${longDescription}`
```
miguel-nascimento, line 81: Why we need HTML tag in this string? Is it a hack to render in opensea?

packages/stream-metadata/src/routes/spaceMetadata.ts
```diff
@@ -1,36 +1,89 @@
...
-		image: `${serverUrl}/space/${spaceAddress}/image`,
+	let spaceInfo: SpaceInfo | undefined
+	try {
+		spaceInfo = await spaceDapp.getSpaceInfo(spaceAddress)
+	} catch (error) {
+		logger.error({ spaceAddress, error }, 'Failed to fetch space contract info')
+		return reply
+			.code(404)
+			.send({ error: 'Not Found', message: 'Failed to fetch space contract info' })
+	}
```
texuf, line 50: i don't like that this will eat any error from infura and return a 404 instead of 500 if infura is down

texuf (approved):

PR #885, opened 2024-08-27
enable caching on the stream-metadata service

tak-hntlabs:
Branch off from [PR 860](https://github.com/river-build/river/pull/860).

- enable image caching using the streamId+secret+iv as the cache key
- add test to verify that the updated space image is returned correctly with caching enabled
- add an optional env config to enable / disable cache. Enabled by default.

### DIFF
--- a/packages/stream-metadata/src/riverStreamRpcClient.ts
+++ b/packages/stream-metadata/src/riverStreamRpcClient.ts
@@ -207,11 +207,11 @@ export async function getMediaStreamContent(
 
 	const secretHex = toHexString(secret)
 	const ivHex = toHexString(iv)
+	const cacheKey = `${fullStreamId}${secretHex}${ivHex}`
 
 	if (config.enableCache) {
-		const concatenatedString = `${fullStreamId}${secretHex}${ivHex}`
-	if (contentCache[concatenatedString]) {
-		return contentCache[concatenatedString]
+		if (contentCache[cacheKey]) {
+		return contentCache[cacheKey]
 		}
 	}
 
@@ -226,8 +226,7 @@ export async function getMediaStreamContent(
 	const result = await mediaContentFromStreamView(logger, sv, secret, iv)
 
 	// Cache the result
-	const concatenatedString = `${fullStreamId}${secretHex}${ivHex}`
-	contentCache[concatenatedString] = result
+	contentCache[cacheKey] = result
 
 	return result
 }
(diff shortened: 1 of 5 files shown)


packages/stream-metadata/src/riverStreamRpcClient.ts
```diff
@@ -206,12 +207,13 @@ export async function getMediaStreamContent(
 
 	const secretHex = toHexString(secret)
 	const ivHex = toHexString(iv)
+	const cacheKey = `${fullStreamId}${secretHex}${ivHex}`
```
brianathere, line 210: I'd suggest making this a hash of these, so you can also have a shorter value to use as a key to store this in redis in the future by the same key.

brianathere: And move it into the if block, so if caching isn't enabled, it isn't computed

packages/stream-metadata/tests/testUtils.ts
```diff
@@ -102,13 +104,21 @@ export async function makeUserContext(wallet: ethers.Wallet): Promise<SignerCont
 	return makeSignerContext(userPrimaryWallet, delegateWallet, { days: 1 })
 }
 
-export function makeJpegBlob(fillSize: number): {
+export function makeImageBlob(mimeType: ImageMimeType, fillSize: number): {
```
brianathere, line 107: I don't see why you needed to make the mimeType settable in this way to test caching. You could have just sent a different jpeg. Right?

brianathere (approved):

texuf (changes requested):
can you hold off on this until we chat tomorrow? Kerem has a task to set up cloudfront for this service. If that is working I'm wondering what this is going to give us aside from out of memory errors.

packages/stream-metadata/src/riverStreamRpcClient.ts
```diff
@@ -218,8 +220,7 @@ export async function getMediaStreamContent(
 	const result = await mediaContentFromStreamView(logger, sv, secret, iv)
 
 	// Cache the result
-	const concatenatedString = `${fullStreamId}${secretHex}${ivHex}`
-	contentCache[concatenatedString] = result
+	contentCache[cacheKey] = result
```
mechanical-turk, line 223: @tak-hntlabs is this temporary until we introduce redis? we'll get out of memory errors - unless you have some custom ttl or cache invalidation logic running somewhere else. if we want to use local memory based caching (regardless of short term vs long term) we should look into [lru-cache](https://www.npmjs.com/package/lru-cache)

tak-hntlabs:
decided not to cache. Leverage cloudfront

PR #889, opened 2024-08-27
perf(contract): Replace `Multicall` with Solady's `Multicallable`

shuhuiluo:
Removed internal `Multicall` implementation and interface. Updated references to use `Multicallable` from the Solady library for batch function calls. This optimizes and simplifies the contract dependencies.


giuseppecrj:
Should we consider https://github.com/Vectorized/multicaller/blob/main/src/Multicaller.sol since this implementation is not payable aware

shuhuiluo:
> Should we consider https://github.com/Vectorized/multicaller/blob/main/src/Multicaller.sol since this implementation is not payable aware

@giuseppecrj This one is a more general multicall for calling external contracts and not a drop-in replacement.

shuhuiluo:
> should also add tests for how multicaller would be used in space proxy context.

There's no dedicated test contract for `SpaceProxy` but I added a fuzz test for multicall.

giuseppecrj:
> > should also add tests for how multicaller would be used in space proxy context.
> 
> There's no dedicated test contract for `SpaceProxy` but I added a fuzz test for multicall.

the space proxy is an instance of a space, you can create test for it by inheriting BaseSetup and using either the `space` or `everyoneSpace` address since they're proxies

shuhuiluo:
> > > should also add tests for how multicaller would be used in space proxy context.
> > 
> > 
> > There's no dedicated test contract for `SpaceProxy` but I added a fuzz test for multicall.
> 
> the space proxy is an instance of a space, you can create test for it by inheriting BaseSetup and using either the `space` or `everyoneSpace` address since they're proxies

@giuseppecrj Added.

giuseppecrj (approved):

PR #895, opened 2024-08-27
Fix txid logging format for xchain when fetching rule data

clemire:


texuf (approved):

core/xchain/server/server.go
```diff
@@ -529,7 +529,7 @@ func (x *xchain) getRuleData(
 	contractAddress common.Address,
 	client crypto.BlockchainClient,
 ) (*base.IRuleEntitlementBaseRuleDataV2, error) {
-	log := x.Log(ctx).With("function", "getRuleData", "req.txid", transactionId)
+	log := x.Log(ctx).With("function", "getRuleData", "req.txid", hex.EncodeToString(transactionId))
```
texuf, line 532: it's weird i thought the logger was converting these for us

clemire: No, if you look in datadog you'll see this particular log prints as a byte array and it's the only place in the code where the TransactionId is not formatted with `hex.EncodeToString`. In all other datadog logs, the transactionId renders in a human-readable way.

PR #897, opened 2024-08-27
Short-circuit ruleData evaluation when first good result is found

clemire:
This underlying call now uses a promise.any and throws an error for a non-entitled result to ensure the promise will terminate as soon as an entitled wallet is found.

### DIFF
--- a/packages/web3/src/v3/SpaceDapp.ts
+++ b/packages/web3/src/v3/SpaceDapp.ts
@@ -692,25 +692,24 @@ export class SpaceDapp implements ISpaceDapp {
                 .map((x) => x.ruleEntitlement!.rules as IRuleEntitlementV2Base.RuleDataV2Struct),
         )
 
-        const entitledWalletsForAllRuleEntitlements = await Promise.all(
+        return await Promise.any(
             ruleEntitlements.map(async (ruleData) => {
                 if (!ruleData) {
                     throw new Error('Rule data not found')
                 }
                 const operations = ruleDataToOperations(ruleData)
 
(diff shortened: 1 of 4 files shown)


packages/web3/src/v3/SpaceDapp.ts
```diff
@@ -692,25 +692,24 @@ export class SpaceDapp implements ISpaceDapp {
...
-
-        const validWallets = entitledWalletsForAllRuleEntitlements.filter(
-            (w) => w !== ethers.constants.AddressZero,
-        )
-        // if any ruleData check passes with an entitled wallet, return the first wallet that passed
-        if (validWallets.length > 0) {
-            return validWallets[0]
-        }
-        return
+        ).catch(() => undefined)
```
texuf, line 712: doing it this way conflates ledgit errors with unentitled, and we'd rather have an error bubble up to the ui than kick someone from a space because they went through a tunnel

packages/web3/tsconfig.json
```diff
@@ -4,5 +4,5 @@
         "baseUrl": "./src" /* Specify the root folder within your source files. */,
         "outDir": "./dist" /* Specify an output folder for all emitted files. */
     },
-    "include": ["src/**/*"]
+    "include": ["src/**/*", "tests/**/*"]
```
clemire, line 7: Can move this into a separate PR if preferred. Otherwise it may make sense to keep this here because the context is, Promise.any and AggregateError were undefined and that's why we made this change.

clemire:
For test flake: https://github.com/river-build/river/pull/933

texuf (approved):
neat!

PR #905, opened 2024-08-27
Custom entitlement factory to enable POAP gating

brianathere:
Simple ICustomEntitlement deployer that enables creation of a Custom Entitlement per eventId that can be used to gate by POAP Event ID on Gnosis. 

@giuseppecrj and I have discussed making it a BeaconProxy so we can change the implementation of isEntitled. I can see both sides of that given the simplicity of this contract and the use case.

Would be nice after using this in anger to improve the ICustomEntitlement interface to also return more information to make the UI in Towns better when people select this contract. @giuseppecrj has another PR to at least add support for detecting contracts claim to implement ICustomEntitlement instead of just evm_call isEntitled to see if the selector is handled. We should consider also a call to return some metadata so the UI can show what the custom entitlement claims, in this case it could have the POAP Event Name.

Still, this is a functional path forward to gating by POAP on Gnosis (and other chains)


contracts/src/spaces/entitlements/poap/PoapEntitlementFactory.sol
```diff
@@ -0,0 +1,48 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.23;
+
+import {Factory} from "contracts/src/utils/Factory.sol";
+import {PoapEntitlement} from "./PoapEntitlement.sol";
+
+contract PoapEntitlementFactory is Factory {
```
giuseppecrj, line 7: provide an interface for PoapEntitlementFactory in separate file

contracts/src/spaces/entitlements/poap/PoapEntitlementFactory.sol
```diff
@@ -0,0 +1,48 @@
...
+    }
+  }
+
+  function getEntitlementAddress(
+    uint256 eventId
+  ) public view returns (address) {
+    return predictEntitlementAddress(eventId);
+  }
+
+  function getSalt(uint256 eventId) public pure returns (bytes32) {
```
giuseppecrj, line 36: make internal

PR #911, opened 2024-08-28
use zod to do schema validation on incoming requests /user/:userId/image and /user/:userId/bio

tak-hntlabs:


miguel-nascimento (approved):
thank you!

PR #915, opened 2024-08-28
Fix post order traversal bug for rule entitlements with multiple checkOps

clemire:
The post order traversal method is used to convert rule datas fetched from contracts into a tree data structure that can be recursively used by the entitlements library to evaluate an entitlement. This bug was manifesting on alpha / gamma for rules with > 2 entitlements.

I added new unit tests that would have caught the bug, and two new e2e tests for channels and spaces that both fail without the bug fix.


packages/sdk/src/channelsWithEntitlements.test.ts
```diff
@@ -675,6 +676,77 @@ describe('channelsWithEntitlements', () => {
         await alice.stopSync()
     })
 
+    test('user with only one entitlement from 3-nested NFT rule data can join channel', async () => {
```
clemire, line 679: The e2e test cases reflect the issue we saw in gamma, with 3 OR'ed rule entitlements.

packages/web3/src/entitlement.ts
```diff
@@ -367,8 +367,8 @@ export function postOrderTraversal(operation: Operation, data: DeepWriteable<Rul
     } else if (isLogicalOperation(operation)) {
         data.logicalOperations.push({
             logOpType: operation.logicalType,
-            leftOperationIndex: data.logicalOperations.length, // Index of left child
-            rightOperationIndex: data.logicalOperations.length + 1, // Index of right child
+            leftOperationIndex: data.operations.length - 2, // Index of left child
```
clemire, line 370: Here is the bug. This change more closely reflects what the go code is doing.

packages/sdk/src/spaceWithEntitlements.test.ts
```diff
@@ -783,6 +784,99 @@ describe('spaceWithEntitlements', () => {
...
+
+        // kill the clients
+        const doneStart = Date.now()
+        await bob.stopSync()
+        await alice.stopSync()
+        await carol.stopSync()
+        await dave.stopSync()
+        await emily.stopSync()
+        log('Done', Date.now() - doneStart)
+    })
```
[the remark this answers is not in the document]
clemire: that goes against convention of how the file is written, but looks like a nice refactor for some other time

brianathere (approved):

texuf (approved):

PR #920, opened 2024-08-29
feat(contracts): move entitlement gated to separate facet in space

giuseppecrj:


contracts/src/spaces/facets/roles/RolesBase.sol
```diff
@@ -96,42 +96,37 @@ abstract contract RolesBase is IRolesBase {
...
-      rolesWithPermission[i] = Role({
-        id: roleId,
-        name: name,
-        disabled: isImmutable,
-        permissions: permissions,
-        entitlements: entitlements
-      });
+    // Resize the array to remove unused slots
+    assembly {
+      mstore(rolesWithPermission, count)
```
brianathere, line 129: This is cool!

foundry.toml
```diff
@@ -21,6 +21,7 @@ fs_permissions = [
 ] # Configures permissions for cheatcodes that touch the filesystem
 extra_output = ["metadata", "abi", "bin"]
 extra_output_files = ["metadata", "abi", "bin"]
+ignored_error_codes = [5574]
```
brianathere, line 24: Why? Comment

brianathere (approved):
Should this be two PR?

foundry.toml
giuseppecrj: removed in latest push, was trying to convert to 0.8.25 but got a bunch of errors

texuf (approved):

PR #940, opened 2024-08-30
fix(contracts): fixes price slippage

giuseppecrj:
PROPOSAL ON TESTNET: https://testnets.llama.xyz/orgs/river/base-sepolia/actions/170
PROPOSAL ON MAINNET: https://app.llama.xyz/orgs/river/base/actions/101

### DIFF
--- a/contracts/deployments/gamma/base/addresses/membershipFacet.json
+++ b/contracts/deployments/gamma/base/addresses/membershipFacet.json
@@ -1,3 +1,3 @@
 {
-  "address": "0x25d87F8d26da7c3cC25C42a2d2E42C264a432B5d"
+  "address": "0x6D0dcdC1DB42FdF90C299CCB89bC5be139aB777e"
 }
--- a/contracts/deployments/gamma/base/addresses/membershipFacet.json
+++ b/contracts/deployments/gamma/base/addresses/membershipFacet.json
@@ -1,0 +1,3 @@
+{
+  "address": "0x25d87F8d26da7c3cC25C42a2d2E42C264a432B5d"
+}
--- a/contracts/src/spaces/facets/membership/MembershipBase.sol
+++ b/contracts/src/spaces/facets/membership/MembershipBase.sol
@@ -30,6 +30,7 @@ abstract contract MembershipBase is IMembershipBase {
 
     if (info.freeAllocation > 0) {
       _verifyFreeAllocation(info.freeAllocation);
+      ds.freeAllocationEnabled = true;
     }
 
     if (info.price > 0) {
--- a/contracts/src/spaces/facets/membership/MembershipBase.sol
+++ b/contracts/src/spaces/facets/membership/MembershipBase.sol
@@ -5,7 +5,6 @@ pragma solidity ^0.8.23;
 import {IMembershipBase} from "./IMembership.sol";
 import {IPlatformRequirements} from "contracts/src/factory/facets/platform/requirements/IPlatformRequirements.sol";
 import {IMembershipPricing} from "./pricing/IMembershipPricing.sol";
-import {IPrepay} from "contracts/src/spaces/facets/prepay/IPrepay.sol";
 import {IPricingModules} from "contracts/src/factory/facets/architect/pricing/IPricingModules.sol";
 // libraries
 import {MembershipStorage} from "./MembershipStorage.sol";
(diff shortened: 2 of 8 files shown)


contracts/test/spaces/membership/unit/MembershipJoinSpace.t.sol
```diff
@@ -518,4 +518,40 @@ contract MembershipJoinSpace is
...
+    for (uint256 i = 0; i < 3; i++) {
+      vm.prank(selectedNodes[i]);
+      IEntitlementGated(contractAddress).postEntitlementCheckResult(
+        transactionId,
+        roleId,
+        IEntitlementGatedBase.NodeVoteStatus.PASSED
+      );
+    }
+
+    assertEq(membership.balanceOf(bob), 1);
```
clemire, line 554: I'm not sure if the above changes + tests fully address a comment Brian made about executing the example transaction in tenderly, where he said "in this specific case, the user had enough value and we should have proceeded and then refunded the balance." Do we have any testing that validates that if a user sends >> value than membership price they get a mint + a refund?

giuseppecrj: don't think we should do refund and mint, you should be sending the exact amount

brianathere: I don't believe we can send exact amount, I believe there's always going to be a slippage window. Happy to talk about it.

giuseppecrj: in Brian's use case the person didn't send more, they sent what was priced but the price dropped

giuseppecrj: can you explain how this would happen with this fix implemented?

brianathere (approved):
Approve as a temporary solution, I would like us to prioritize a more robust fix

contracts/src/spaces/facets/membership/MembershipFacet.sol
```diff
@@ -142,38 +143,43 @@ contract MembershipFacet is
...
+      shouldCharge = false;
       _reducePrepay(1);
       _refundBalance(transactionId, sender);
-    } else {
-      membershipPrice = _getMembershipPrice(totalSupply);
+    }
+
+    if (shouldCharge) {
+      membershipPrice = _getCapturedValue(transactionId);
+      if (membershipPrice == 0) revert Membership__InsufficientPayment();
```
shuhuiluo, line 176:
```suggestion
      paidPrice = _getCapturedValue(transactionId);
      if (paidPrice == 0) revert Membership__InsufficientPayment();
```
nit

contracts/src/spaces/facets/membership/MembershipBase.sol
```diff
@@ -30,6 +29,7 @@ abstract contract MembershipBase is IMembershipBase {
 
     if (info.freeAllocation > 0) {
       _verifyFreeAllocation(info.freeAllocation);
+      ds.freeAllocationEnabled = true;
```
shuhuiluo, line 32: `freeAllocationEnabled` seems redundant. Where do we need it instead of `freeAllocation > 0`?

contracts/test/spaces/membership/unit/MembershipJoinSpace.t.sol
```diff
@@ -518,4 +520,66 @@ contract MembershipJoinSpace is
...
+      freeAllocationInfo
+    );
+
+    MembershipFacet freeAllocationMembership = MembershipFacet(
+      freeAllocationSpace
+    );
+
+    vm.prank(bob);
+    vm.expectRevert(Membership__InsufficientPayment.selector);
+    freeAllocationMembership.joinSpace(bob);
```
shuhuiluo, line 583: Does it revert because the free allocation is taken up by the founder?

contracts/src/spaces/facets/membership/MembershipBase.sol
giuseppecrj: in _getMembershipFreeAllocation, it's a requirement for default memberhsip limit via platform requirements to work and for prepay to work when charging

contracts/test/spaces/membership/unit/MembershipJoinSpace.t.sol
giuseppecrj: correct

shuhuiluo: A comment is helpful.

shuhuiluo (approved):

contracts/deployments/gamma/base/addresses/membershipFacet.json
```diff
@@ -0,0 +1,3 @@
+{
+  "address": "0x6D0dcdC1DB42FdF90C299CCB89bC5be139aB777e"
```
clemire, line 2: I don't know a lot about how we deploy to alpha - do we need one of these files generated for alpha as well?

PR #952, opened 2024-09-03
core/node: reconcile streams with peers on node start

### LINKED ISSUE #494: Reconcile local storage to state specified in contract on start
### Description

_No response_

bas-vk:
With stream replications it is possible that a local stream advances when the node is down. 

This change adds support to the node that at boots it will compare streams from its local database with the streams registry. When the database is out of sync the node requests missing blocks from peers and write them to the database bringing the database and contract in sync.

closes #494

### DIFF
--- a/core/node/events/stream.go
+++ b/core/node/events/stream.go
@@ -141,14 +141,14 @@ func (s *streamImpl) importMiniblocks(
 		return nil
 	}
 
-	blocksToWriteToStorage := make([]*storage.LatestMiniBlock, len(miniblocks))
+	blocksToWriteToStorage := make([]*storage.MiniblockData, len(miniblocks))
 	for i, miniblock := range miniblocks {
 		bytes, err := miniblock.ToBytes()
 		if err != nil {
 			return err
 		}
 
-		blocksToWriteToStorage[i] = &storage.LatestMiniBlock{
+		blocksToWriteToStorage[i] = &storage.MiniblockData{
 			StreamID:      s.streamId,
 			Number:        miniblock.Num,
 			MiniBlockInfo: bytes,
@@ -176,32 +176,30 @@ func (s *streamImpl) importMiniblocks(
 	// apply mini-blocks one by one on view, backup existing view in case
 	// applying/writing miniblocks fails rollback view.
 	var (
-		err              error
-		viewBeforeImport = s.view
+		err  error
+		view = s.view
 	)
+
 	for _, miniblock := range miniblocks {
-		if miniblock.Num <= s.view.LastBlock().Num {
+		if miniblock.Num <= view.LastBlock().Num {
 			blocksToWriteToStorage = blocksToWriteToStorage[1:]
 			continue
 		}
 
-		view, err := s.view.copyAndApplyBlock(miniblock, s.params.ChainConfig, true)
+		view, err = view.copyAndApplyBlock(miniblock, s.params.ChainConfig, true)
 		if err != nil {
-			s.view = viewBeforeImport
 			return err
 		}
-
-		s.view = view
 	}
 
-	err = s.params.Storage.ImportMiniBlocks(ctx, blocksToWriteToStorage)
+	err = s.params.Storage.ImportMiniblocks(ctx, blocksToWriteToStorage)
 	if err != nil {
-		s.view = viewBeforeImport
 		return err
 	}
 
-	// TODO: inform subscribes with rate throttling?
+	s.view = view
 
+	// TODO: inform subscribes with rate throttling?
 	return nil
 }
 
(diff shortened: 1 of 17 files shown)


core/node/events/stream.go
```diff
@@ -134,6 +132,79 @@ func (s *streamImpl) ApplyMiniblock(ctx context.Context, miniblock *MiniblockInf
...
+		if err := s.loadInternal(ctx); err != nil {
+			return err
+		}
+	}
+
+	// apply mini-blocks one by one on view, backup existing view in case
+	// applying/writing miniblocks fails rollback view.
+	var (
+		err              error
+		viewBeforeImport = s.view
```
sergekh2, line 180: it's less error-prone to construct new view and then assign it to the s.view at the end: if there are more than one error returns, it's easy to forget to flip it back (in other words, it's better to have `if err != nil { return err }` without extra cleanup steps)

core/node/storage/pg_storage.go
```diff
@@ -1754,3 +1753,169 @@ func (s *PostgresEventStore) debugReadStreamData(
...
+	streamID StreamId,
+) (*LatestMiniBlock, error) {
+	var ret *LatestMiniBlock
+	err := s.txRunner(
+		ctx,
+		"StreamLastMiniBlocks",
+		pgx.ReadOnly,
+		func(ctx context.Context, tx pgx.Tx) error {
+			var err error
+			ret, err = s.lastMiniBlockForStream(ctx, tx, streamID)
```
sergekh2, line 1768: by convention in this file inner tx functions are called by the name of the calling function with Tx, i.e. this one should be `streamLastMiniBlockTx`

core/node/storage/pg_storage.go
```diff
@@ -1754,3 +1753,169 @@ func (s *PostgresEventStore) debugReadStreamData(
...
+			}
+
+			return s.importMiniBlocks(ctx, tx, miniBlocks)
+		},
+		nil,
+		"streamId", miniBlocks[0].StreamID,
+	)
+}
+
+func (s *PostgresEventStore) importMiniBlocks(ctx context.Context, tx pgx.Tx, miniBlocks []*LatestMiniBlock) error {
```
sergekh2, line 1846:
```suggestion
func (s *PostgresEventStore) importMiniblocksTx(ctx context.Context, tx pgx.Tx, miniBlocks []*LatestMiniBlock) error {
```

core/node/storage/pg_storage.go
```diff
@@ -1754,3 +1753,169 @@ func (s *PostgresEventStore) debugReadStreamData(
...
+	ctx context.Context,
+	tx pgx.Tx,
+	streamID StreamId,
+) (*LatestMiniBlock, error) {
+	miniblockNumsRow, err := tx.Query(
+		ctx,
+		`select o.stream_id, o.seq_num, o.blockdata from miniblocks o inner join (
+			select stream_id, max(seq_num) as max_seq_num from miniblocks
+				WHERE stream_id = $1 group by stream_id
+			) i on o.stream_id = i.stream_id and o.seq_num = i.max_seq_num`,
```
sergekh2, line 1789: Simpler query should be something like this:
```suggestion
"SELECT seq_num, blockdata
FROM miniblocks
WHERE stream_id = $1
ORDER BY seq_num DESC
LIMIT 1;",
```

core/node/storage/pg_storage.go
```diff
@@ -1754,3 +1753,169 @@ func (s *PostgresEventStore) debugReadStreamData(
...
+	defer miniblockNumsRow.Close()
+
+	if miniblockNumsRow.Next() {
+		var (
+			streamId  StreamId
+			maxSeqNum int64
+			blockData []byte
+		)
+
+		if err = miniblockNumsRow.Scan(&streamId, &maxSeqNum, &blockData); err != nil {
```
sergekh2, line 1806: To match suggestion above:
```suggestion
		if err = miniblockNumsRow.Scan(&maxSeqNum, &blockData); err != nil {
```

core/node/storage/pg_storage.go
```diff
@@ -1754,3 +1753,169 @@ func (s *PostgresEventStore) debugReadStreamData(
...
+			MiniBlockInfo: blockData,
+		}, nil
+	}
+
+	return nil, RiverError(Err_NOT_FOUND, "latest miniblock in DB not found for stream").
+		Tags("stream", streamID).
+		Func("lastMiniBlockForStream")
+}
+
+func (s *PostgresEventStore) ImportMiniBlocks(ctx context.Context, miniBlocks []*LatestMiniBlock) error {
```
sergekh2, line 1822: Naming is somewhat inconsistent in many locations: i.e. Miniblock (small b) is used in all other places in code base.

Also: it's not latest miniblock here (i.e. type should be renamed to something more generic)

core/node/storage/pg_storage.go
```diff
@@ -1754,3 +1753,169 @@ func (s *PostgresEventStore) debugReadStreamData(
...
+		return err
+	}
+
+	if seqNum == nil {
+		seqNum = new(int64)
+		*seqNum = -1
+	}
+
+	// clean up minipool
+	_, err = tx.Exec(ctx, "DELETE FROM minipools WHERE slot_num > -1 AND stream_id = $1", streamID)
```
sergekh2, line 1870: lets merge like this, but this can drop events (and make in-memory view inconsistent with db)

sergekh2 (approved):

PR #956, opened 2024-09-03
sdk: allow empty bio in UserBio by adding updatedAt field

miguel-nascimento:
Without the `updated_at` field, protobuf with empty bio field was failing to encrypt.

### DIFF
--- a/protocol/payloads.proto
+++ b/protocol/payloads.proto
@@ -188,7 +188,7 @@ message ChunkedMedia {
 
 message UserBio {
     string bio = 1;
-    optional uint64 updated_at = 2;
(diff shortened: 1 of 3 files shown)


texuf (approved):

protocol/payloads.proto
```diff
@@ -188,6 +188,7 @@ message ChunkedMedia {
 
 message UserBio {
     string bio = 1;
+    optional uint64 updated_at = 2;
```
texuf, line 191: before checking in can you call this updated_at_epoch_ms to match what we have in protocol.proto?. With dates it's good to be explicit.

PR #958, opened 2024-09-03
Implementation of ERC1155 checks in go, including xchain IT tests.

clemire:


core/contracts/base/channels.go
clemire, line 1: This file got updated by the gen-river-node-bindings.sh script, not specifically relevant to this PR but keeping it to keep code up-to-date

texuf (approved):

bas-vk (approved):

PR #961, opened 2024-09-04
Fix broken unittest positive condition for typescript entitlement checks

clemire:
...and repair exposed broken test cases.

positive results were expecting a "truthy" response, which could be any nonempty string, including the zero address. This PR fixes tests to expect a valid non-zero address instead, wherever we expect an entitled wallet to be computed.

### DIFF
--- a/packages/web3/src/entitlement.test.ts
+++ b/packages/web3/src/entitlement.test.ts
@@ -239,7 +239,7 @@ const nftCases = [
         desc: 'eth sepolia (insufficient balance)',
         check: nftMultiCheckEthereumSepolia,
         wallets: [SepoliaTestNftWallet_1Token],
-        provider: baseSepoliaProvider,
+        provider: ethSepoliaProvider,
         expectedResult: false,
     },
     {
@@ -250,7 +250,7 @@ const nftCases = [
             SepoliaTestNftWallet_2Tokens,
             SepoliaTestNftWallet_3Tokens,
         ],
-        provider: baseSepoliaProvider,
+        provider: ethSepoliaProvider,
         expectedResult: true,
     },
     {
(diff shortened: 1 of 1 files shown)


packages/web3/src/entitlement.test.ts
```diff
@@ -285,7 +286,7 @@ const ethSepoliaChainLinkContract: Address = '0x779877A7B0D9E8603169DdbD7836e478
 // Some wallet addresses are duplicated for the sake of self-documenting variable names.
 const sepoliaChainLinkWallet_50Link: Address = '0x4BCfC6962Ab0297aF801da21216014F53B46E991'
 const sepoliaChainLinkWallet_25Link: Address = '0xa4D440AeA5F555feEB5AEa0ddcED6e1B9FaD6A9C'
-const baseSepoliaChainLinkWallet_50Link: Address = '0x4BCfC6962Ab0297aF801da21216014F53B46E991'
+const baseSepoliaChainLinkWallet_25Link2: Address = '0x4BCfC6962Ab0297aF801da21216014F53B46E991'
```
clemire, line 289: This wallet actually had 25 link, as @erikolsson noticed

packages/web3/src/entitlement.test.ts
```diff
@@ -271,7 +271,8 @@ test.each(nftCases)('erc721Check - $desc', async (props) => {
     const controller = new AbortController()
     const result = await evaluateTree(controller, wallets, [provider], check)
     if (expectedResult) {
-        expect(result).toBeTruthy()
+        expect(result as Address).toBeTruthy()
```
clemire, line 274: Change all test cases that expect an entitled wallet to check that the string result is non-zero address

texuf (approved):

PR #967, opened 2024-09-04
Change max number of pins allowed per stream

texuf:
To 100, want something large enough that we don’t have to worry about the limit for a while


ttillberg (approved):
💯

PR #973, opened 2024-09-05
Change can_add_event rules to prevent pinning if you don’t have pinning permissions

texuf:


texuf:
This worked locally, maybe I need to sleep to wait for the cache to bust

core/node/auth/auth_impl.go
```diff
@@ -414,7 +414,7 @@ func (ca *chainAuth) isEntitledToChannelUncached(
 
 	// For read, write and react permissions, fetch the entitlements and evaluate them locally.
 	if (args.permission == PermissionRead) || (args.permission == PermissionWrite) ||
-		(args.permission == PermissionReact) {
```
clemire, line 417: I see you removed these here... https://github.com/river-build/river/pull/974/files

clemire (approved):
This PR looks good, the test case fix is unrelated to this change. Once tests are green this one should be good to submit.

PR #974, opened 2024-09-05
Expand xchan channel permissions to all permission types

texuf:
Isn’t this what we want? Why would we have anything otherwise?

### DIFF
--- a/core/node/auth/auth_impl.go
+++ b/core/node/auth/auth_impl.go
@@ -412,59 +412,44 @@ func (ca *chainAuth) isEntitledToChannelUncached(
 	log := dlog.FromCtx(ctx)
 	log.Debug("isEntitledToChannelUncached", "args", args)
 
-	// For read, write and react permissions, fetch the entitlements and evaluate them locally.
-	if (args.permission == PermissionRead) || (args.permission == PermissionWrite) ||
-		(args.permission == PermissionReact) {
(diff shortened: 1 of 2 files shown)


core/node/auth/auth_impl.go
```diff
@@ -412,59 +412,44 @@ func (ca *chainAuth) isEntitledToChannelUncached(
 	log := dlog.FromCtx(ctx)
 	log.Debug("isEntitledToChannelUncached", "args", args)
 
-	// For read, write and react permissions, fetch the entitlements and evaluate them locally.
-	if (args.permission == PermissionRead) || (args.permission == PermissionWrite) ||
-		(args.permission == PermissionReact) {
```
texuf, line 417: I just removed this if statement, the rest is formatting

packages/web3/src/v3/SpaceDapp.ts
```diff
@@ -877,43 +877,34 @@ export class SpaceDapp implements ISpaceDapp {
 
         const channelId = ensureHexPrefix(channelNetworkId)
 
-        if (
-            permission === Permission.Read ||
-            permission === Permission.Write ||
-            permission === Permission.React ||
-            permission === Permission.PinMessage
-        ) {
```
texuf, line 885: I just removed this if statement, the rest is formatting

clemire (approved):

PR #975, opened 2024-09-05
Add pin and unpin to the sync agent

texuf:
Update return values for pin and unpin to return event id


packages/sdk/src/sync-agent/syncAgents.test.ts
```diff
@@ -75,4 +75,38 @@ describe('syncAgents.test.ts', () => {
...
+            expect(
+                bob.riverConnection.client?.streams.get(channelId)?.view.membershipContent.pins
+                    .length,
+            ).toBe(1),
+        )
+        // bob can unpin
+        const result2 = await channel.unpin(event!.eventId)
+        expect(result2).toBeDefined()
+        expect(result2.error).toBeUndefined()
+
```
clemire, line 104: wdyt about adding a wait for membership content pins length to go back to 0? (is that too pedantic?)

PR #982, opened 2024-09-05
Jt/net 61 ci differ

jterzis:
Initial ci job for running diff tool and upgrading facets. Will implement each constituent job after this PR, however, need this PR merged to be able to test action in a separate branch.


.github/workflows/Network_diff_upgrade.yml
```diff
@@ -0,0 +1,21 @@
...
+        uses: ./.github/workflows/Bytecode_diff_report.yml
+        secrets: inherit
+        with:
+            origin_environment: alpha
+            target_environment: gamma
+
+    upgrade_gamma_facets:
+        name: Upgrade Gamma Facets
+        needs: diff_report
+        uses: ./.github/workflows/Upgrade_network_facets.yml
```
mechanical-turk, line 18: @jterzis the 2 workflows that this workflow refers to on lines 9 and 18 are showing as empty files. 

the other thing i'm noticing is that the upgrade_gamma_facets job has no way of reacting to the outcome of the first job, since these are not 2 steps on the same workflow but rather 2 separate workflows that get triggered in parallel. maybe that's intentional but calling it out just in case

jterzis: Where are you seeing that they are empty ? The two files are included actually in this pr so maybe that is why ? 

As for the second issue, these should be linked steps and I will make the change. I just need to merge the initial workflow file to be able to iterate on it in a subsequent branch. Otherwise I can't test the workflow in github.

mechanical-turk: this is what im seeing, maybe the gui is tripping:

image

mechanical-turk: and about testing flow, sure sounds good to me. i'm approving 👍

mechanical-turk (approved):

PR #984, opened 2024-09-06
sdk: improve type infer on withStream fn

miguel-nascimento:


packages/sdk/src/sync-agent/spaces/models/channel.ts
```diff
@@ -77,21 +77,19 @@ export class Channel extends PersistedObservable<ChannelModel> {
...
+                replyPreview: options?.replyId ? '🙈' : undefined,
+                content: {
+                    body: message,
+                    mentions: options?.mentions ?? [],
+                    attachments: options?.attachments ?? [],
+                },
             })
+        })
         return result
     }
```
texuf, line 95: realizing the pin and unpin should use withStream too, i'll make a note to fix

texuf (approved):
nice!!!!

PR #986, opened 2024-09-06
Update ETH balance go implementation to consider all ether-based chains

clemire:
This PR also includes configuration for stream nodes that defaults to sane values if not explicitly set. It uses the same design as existing Xchain configuration and should be similarly forward-compatible with the move to on-chain configuration.

This PR leaves the repo in an inconsistent state, as the client is still evaluating eth checks on a single chain - however, the eth check is already broken because it does not meet the spec, so I figured this was fine.


core/config/blockchain_info.go
```diff
@@ -8,6 +8,24 @@ type BlockchainInfo struct {
...
+
+		// Testnets
+		11155111, // Ethereum Sepolia
+		84532,    // Base Sepolia
+
+		// Local
+		31337, // Anvil Base
+		31338, // Anvil River
+	}
+}
```
texuf, line 27: did you consider adding this to the BlockChainInfo struct and returning it below?

clemire: Good idea!

core/config/config.go
```diff
@@ -146,6 +146,12 @@ type Config struct {
 	// If value is not set, as a fallback ChainConfigs keys are used.
 	XChainBlockchains []uint64
 
+	// EtherBasedXChainBlockchains is a list of chain IDs that are allowed to be used in xChain checks.
+	// If value is not set, the chains in GetDefaultBlockchainInfo() that use ether as a native fee currency
+	// will be used. This list will always be a subset of XChainBlockchains - exteraneous chain IDs will be discarded.
+	// TODO: this value is going to be moved on-chain so same setting is shared between all nodes and clients.
+	EtherBasedXChainBlockchains []uint64
```
sergekh2, line 153: This PR looks good, however it is not clear why this config param is required: would it make sense just to filter `XChainBlockchains` by `IsEtherBased` and get this list this way? I.e. do we have a scenario when chain is Ehter-based, but we don't want it in the balance check?

clemire: If GetDefaultBlockchainInfo captures information we plan to move on-chain, then I think that's fine, I can refactor.

sergekh2 (approved):

PR #987, opened 2024-09-06
add create space with prepay

giuseppecrj:
TESTNET PROPOSALS:
https://testnets.llama.xyz/orgs/river/base-sepolia/actions/186
https://testnets.llama.xyz/orgs/river/base-sepolia/actions/187


contracts/src/factory/facets/architect/Architect.sol
```diff
@@ -43,11 +43,17 @@ contract Architect is
   //                            Space
   // =============================================================
   function createSpace(
-    SpaceInfo calldata spaceInfo
+    SpaceInfo memory spaceInfo
   ) external nonReentrant whenNotPaused returns (address) {
     return _createSpace(spaceInfo);
   }
 
+  function createSpaceWithPrepay(
```
salzbrenner, line 51: I ran build-contract-types and no TS interface was updated to contain createSpace.
Does this need to be added to IArchitecht.sol?

Looking at ISpaceArchitectShim, it uses IArchitect for typings, but then uses Architect.abi.json, which is a different pattern than the other shims, where it uses typings/Xxx and Xxx.abi.json

I don't know the history of the shim - should it be converted to not use IArchitect but instead Architect?

giuseppecrj: yep, my bad. I'll add it in

contracts/src/factory/SpaceFactory.sol
shuhuiluo, line 1: To quickly trace which facet belongs to which diamond, can we leave this as a placeholder inheriting all the facets it's supposed to use even though it doesn't get deployed like that?

contracts/src/factory/facets/architect/ArchitectBase.sol
```diff
@@ -53,8 +54,29 @@ abstract contract ArchitectBase is Factory, IArchitectBase, PricingModulesBase {
     return ArchitectStorage.layout().spaceByTokenId[tokenId];
   }
 
+  function _createSpaceWithPrepay(
+    CreateSpace memory createSpace
```
shuhuiluo, line 58:
```suggestion
    SpaceInfo calldata spaceInfo,
    Prepay calldata prepay
```
I'd say we change `SpaceInfo` to
```solidity
  struct SpaceInfo {
    Metadata metadata;
    Membership membership;
    ChannelInfo channel;
  }
```
for consistency and pass `Prepay` separately so as to keep everything in calldata.

shuhuiluo (changes requested):

contracts/src/factory/facets/architect/IArchitect.sol
```diff
@@ -42,6 +42,24 @@ interface IArchitectBase {
...
+  struct Prepay {
+    uint256 supply;
+  }
+
+  struct CreateSpace {
+    Metadata metadata;
+    Membership membership;
+    ChannelInfo channel;
+    Prepay prepay;
+  }
```
shuhuiluo, line 61:
```suggestion
  struct CreateSpace {
    SpaceInfo spaceInfo;
    Prepay prepay;
  }
```
Or consider this to use calldata without reconstructing `SpaceInfo` in memory.

shuhuiluo (changes requested):

contracts/src/factory/facets/architect/ArchitectBase.sol
giuseppecrj: we're moving from using SpaceInfo to CreateSpace, while keeping backwards compatibility

giuseppecrj: client side already uses these types for old create space so we can't update the current struct without breaking

contracts/src/factory/facets/architect/IArchitect.sol
shuhuiluo: It doesn't change `SpaceInfo`. By keeping it nested in `CreateSpace` we avoid reconstructing `SpaceInfo` in memory.

giuseppecrj: you have a nested spaceInfo key when creating the client object:

```
{
  "spaceInfo": {},
  "prepay": {}
}
```

instead of

```
{
  metadata: "",
  membership: "",
  channel: "",
  prepay: ""
}
```

PR #991, opened 2024-09-06
Fix bug in threshold validation for typescript entitlements check

clemire:
Also, fix / build out error tests.


packages/web3/src/entitlement.ts
```diff
@@ -555,28 +555,35 @@ async function evaluateCheckOperation(
...
     ) {
-        throw new Error('Invalid contract address for check operation NATIVE_COIN_BALANCE')
+        throw new Error(
+            `Invalid contract address for check operation ${checkOpString(operation.checkType)}`,
+        )
     }
 
     if (
-        operation.checkType in
         [
```
clemire, line 564: `in` keyword does not check for array contains in typescript, switched to `array.includes`

texuf (approved):

PR #992, opened 2024-09-06
stream-metadata: follow up for space refresh

miguel-nascimento:
We're having a hard time debugging opensea refresh, so I'm those logs here.

Also fixed the wrong `chainId` comparison 🤦‍♂️

### DIFF
--- a/packages/stream-metadata/src/routes/spaceRefresh.ts
+++ b/packages/stream-metadata/src/routes/spaceRefresh.ts
@@ -1,12 +1,9 @@
-import { FastifyReply, FastifyRequest, type FastifyBaseLogger } from 'fastify'
+import { FastifyReply, FastifyRequest } from 'fastify'
 import { z } from 'zod'
-import { BigNumber } from 'ethers'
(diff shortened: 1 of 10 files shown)


mechanical-turk (approved):

texuf (approved):

packages/stream-metadata/src/routes/spaceRefresh.ts
```diff
@@ -32,52 +30,21 @@ export async function spaceRefresh(request: FastifyRequest, reply: FastifyReply)
...
 	try {
 		const path = `/space/${spaceAddress}/image`
-		await createCloudfrontInvalidation({ path, logger })
 
-		await refreshOpenSea(spaceAddress)
-		logger.info({ path }, 'OpenSea cache invalidated')
+		await Promise.all([
+			createCloudfrontInvalidation({ path, logger }),
+			refreshOpenSea({ spaceAddress, logger }),
+		])
```
texuf, line 37: maybe this should be all settled? but either way both requests go out so i guess it doesn't really matter

PR #1000, opened 2024-09-06
playground: remove json hover

miguel-nascimento:
super annoying and not that useful


miguel-nascimento:
yay PR 1000!!

texuf (approved):
thank you

PR #1002, opened 2024-09-06
sdk(sync-agent): remove unneeded operations on Member

miguel-nascimento:
Why we need to update a member that aren't being used?
- we can know that the member is "being used" if it is in the members record, since when you get it, we create a record for it.


miguel-nascimento:
actually, let's keep it for now, its way easy to create a mental model and find some bugs

PR #1007, opened 2024-09-07
SyncAgent: Update load priority of spaces and channels

texuf:
They load faster this way


miguel-nascimento (approved):
Nice, thanks for finding this, but I think we need to add a few comments about LoadPriority, when it should be high / low

I think that currently we’re using high almost everywhere

PR #1014, opened 2024-09-09
stream-metadata: fix opensea invalidation url

miguel-nascimento:
I was using the networkId by mistake, the chainId prop from the `web3Config.base` is the right one.

### DIFF
--- a/packages/stream-metadata/src/opensea.ts
+++ b/packages/stream-metadata/src/opensea.ts
@@ -5,16 +5,20 @@ import { FastifyBaseLogger } from 'fastify'
 import { config } from './environment'
 import { spaceDapp } from './contract-utils'
 
-const getOpenSeaAPIUrl = (space: SpaceInfo) => {
+const getOpenSeaAPIUrl = (logger: FastifyBaseLogger, space: SpaceInfo) => {
 	const spaceOwnerAddress = config.web3Config.base.addresses.spaceOwner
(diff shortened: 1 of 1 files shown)


packages/stream-metadata/src/opensea.ts
```diff
@@ -5,16 +5,20 @@ import { FastifyBaseLogger } from 'fastify'
...
 	const tokenId = BigNumber.from(space.tokenId).toString()
 
-	if (space.networkId === String(BASE_MAINNET)) {
+	if (chainId === BASE_MAINNET) {
 		return `https://api.opensea.io/api/v2/chain/base/contract/${spaceOwnerAddress}/nfts/${tokenId}/refresh`
-	} else if (space.networkId === String(BASE_SEPOLIA)) {
+	} else if (chainId === BASE_SEPOLIA) {
+		return `https://testnets-api.opensea.io/api/v2/chain/base_sepolia/contract/${spaceOwnerAddress}/nfts/${tokenId}/refresh`
+	} else if (chainId === 31337) {
 		return `https://testnets-api.opensea.io/api/v2/chain/base_sepolia/contract/${spaceOwnerAddress}/nfts/${tokenId}/refresh`
```
texuf, line 18: should localhost just return example.com or something simple?

miguel-nascimento: This endpoint is a weird spot for local development, I would skip for now

PR #1015, opened 2024-09-09
stream-metadata: fix base url in space metadata endpoint

miguel-nascimento:
Previously:
```json
{
  "name": "My fancy space",
  "description": "idk ;-;",
  "image": "https://gamma.river.deliveryspace/0x9e5318618c141f3f9949250bf73fb98a9b22b03e/image"
}
```

This PR fix the url


texuf (approved):

mechanical-turk (approved):

mechanical-turk:
@miguel-nascimento side note: we should move to using a proper url join instead of doing it manually. This may break local dev setups for example. We shouldn't need to worry about whether the root url is provided with a trailing slash and so on

PR #1025, opened 2024-09-09
feat(contract) initial POC of poap entitlement custom

giuseppecrj:


contracts/src/spaces/entitlements/poap/PoapEntitlement.sol
```diff
@@ -0,0 +1,62 @@
...
+    for (uint256 i = 0; i < users.length; i++) {
+      if (_hasEventPoap(users[i], eventId)) {
+        return true;
+      }
+    }
+    return false;
+  }
+
+  function schema() public pure returns (string memory) {
+    return "uint256 eventId";
```
clemire, line 32: For a parameter list with multiple params, would this be comma-separated?
e.g.

"uint256 eventId, address someContractAddress"

clemire (commented):
comment inline

contracts/src/spaces/entitlements/poap/PoapEntitlement.sol
giuseppecrj: right, here's an example: https://sepolia.easscan.org/schema/view/0x4f532d4a635375689656f9fe828c8baaa7d475c90e2dbcd381580d65598244f8

contracts/src/spaces/entitlements/poap/PoapEntitlement.sol
```diff
@@ -0,0 +1,62 @@
...
+
+    for (uint256 i = 0; i < users.length; i++) {
+      if (_hasEventPoap(users[i], eventId)) {
+        return true;
+      }
+    }
+    return false;
+  }
+
+  function schema() public pure returns (string memory) {
```
brianathere, line 31: suggest getParamsSchema

contracts/src/spaces/entitlements/poap/PoapEntitlement.sol
```diff
@@ -0,0 +1,62 @@
...
+contract PoapEntitlement {
+  IPOAP immutable poapContract;
+
+  constructor(address poapContractAddress) {
+    poapContract = IPOAP(poapContractAddress);
+  }
+
+  function isEntitled(
+    address[] calldata users,
+    bytes calldata data
```
brianathere, line 19: aligned with getParamsSchema, I'd name this params

contracts/src/spaces/entitlements/poap/PoapEntitlement.sol
brianathere: Is this an EIP based way of doing this? It's basically just using method selector encoding while retaining the variable name. If it's a standard, then I guess we can start with this, though I'd like to also have a description. Let's hop on a call.

giuseppecrj:
Updated with new changes @clemire

contracts/src/spaces/entitlements/ICrossChainEntitlement.sol
```diff
@@ -0,0 +1,22 @@
...
+    string primitive;
+    string description;
+  }
+
+  function isEntitled(
+    address[] calldata users,
+    bytes calldata data
+  ) external view returns (bool);
+
+  function properties() external view returns (Property[] memory);
```
brianathere, line 21: I believe parameters is more appropriate, as this is describing parameters that must be passed, encoded in this way, as parameters to the isEntitled call, and not actual properties of the ICrossChainEntitlement as the name properties implies.

brianathere (commented):
properties -> parameters

contracts/src/spaces/entitlements/ICrossChainEntitlement.sol
giuseppecrj: sounds good, short term memory thought you said properties and not params during our cal. Will fix soon!

PR #616, opened 2024-08-04
encryption: import / export encryption device

miguel-nascimento:
This PR allows the import / export of the encryption device, enables this feature in the sdk and does some changes to use this feature in the stress client.

PR #1030, opened 2024-09-10
fetch e2ekeys _after_ generating fallback keys

erikolsson:
i believe that this _could_ be the source of the `BAD_MESSAGE_KEY_ID` errors we've seen in prod lately.
The `e2ekeys` depend on the account identity keys, which in turn depend on the result of `generateFallbackKeyIfNeeded()`.

image

I didn't actually run this, but this looks suspicious to me!

if correct, this was introduced in #616 on Aug 6.

### DIFF
--- a/packages/encryption/src/encryptionDevice.ts
+++ b/packages/encryption/src/encryptionDevice.ts
@@ -140,8 +140,8 @@ export class EncryptionDevice {
                 }
                 await this.initializeAccount(account)
             }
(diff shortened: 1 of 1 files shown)


miguel-nascimento (approved):
This looks suspicious for me too. And it’s the only change related in the keys right? I that’s it’s.

Thank you for the catch 🙇‍♂️

texuf (approved):

PR #1033, opened 2024-09-10
Kerem/stream metadata opensea race condition

mechanical-turk:
- fixing a race condition between opensea metadata refresh and cloudfront invalidations
- restructuring the cloudfront logic into a class


packages/stream-metadata/src/aws.ts
```diff
@@ -1,37 +1,106 @@
...
+					invalidation: currentInvalidationCommand,
+					path,
+				},
+				'Waiting for CloudFront cache invalidation to complete...',
+			)
+			if (!currentInvalidationCommand.Invalidation) {
+				throw new Error('Invalidation not found')
+			}
+			await new Promise((resolve) =>
+				setTimeout(resolve, this.config.invalidationConfirmationWait),
```
miguel-nascimento, line 71: would be nice to do this without the sleep, but even [aws sdk does that](https://github.com/aws/aws-sdk-js-v3/blob/e6cba699ea7ca7d9c928e1eae17fc4089e985676/clients/client-s3/src/waiters/waitForObjectExists.ts#L30) (check the [waiter impl](https://github.com/smithy-lang/smithy-typescript/blob/main/packages/util-waiter/src/poller.ts#L25))

miguel-nascimento (approved):

miguel-nascimento:
i think the tests are missing some env vars

texuf (approved):

PR #1035, opened 2024-09-10
chore(contract): use node modules for libs

shuhuiluo:


brianathere (commented):
This was removed to avoid every time a developer does a build resulting in a full rebuild of all of the contracts which a lot of time.

brianathere (approved):

PR #1043, opened 2024-09-11
[BREAKING] Move XChainBlockchains to be on-chain

clemire:
This PR removes XChainBlockchains from the node's static / environment config and moves it to be fetched as an on-chain configuration value. In order to deploy this without borking the network, we would need to manually set the config value for all production environments (alpha, gamma, omega).


core/cmd/run_cmd.go
```diff
@@ -44,7 +44,7 @@ func runServices(ctx context.Context, cfg *config.Config, stream bool, xchain bo
 	var xchainService server.XChain
 	var wg sync.WaitGroup
 	if xchain {
-		xchainService, err = server.New(ctx, cmdConfig, nil, 1, metricsRegistry)
+		xchainService, err = server.New(ctx, cfg, nil, nil, 1, metricsRegistry)
```
clemire, line 47: minor logical fix - cmdConfig is the global config, not what is passed into this method. The cfgs refer to the same object, so there is no change in behavior.

sergekh2 (approved):

bas-vk (approved):

texuf (approved):

texuf:
what's the plan to make the necessary changes on chain?

PR #1047, opened 2024-09-11
feat(contract) update membership metadata to return space owner token uri

giuseppecrj:


contracts/src/spaces/facets/membership/metadata/MembershipMetadata.sol
```diff
@@ -2,59 +2,19 @@
...
-          )
-        )
-      );
-  }
-
+contract MembershipMetadata is ERC721ABase, Facet {
   function tokenURI(uint256 tokenId) public view returns (string memory) {
-    return _render(tokenId);
+    if (!_exists(tokenId)) revert URIQueryForNonexistentToken();
+    TokenOwnableStorage.Layout memory ds = TokenOwnableStorage.layout();
```
shuhuiluo, line 17:
```suggestion
    TokenOwnableStorage.Layout storage ds = TokenOwnableStorage.layout();
```
nit

shuhuiluo (approved):

contracts/src/spaces/facets/membership/metadata/MembershipMetadata.sol
```diff
@@ -2,59 +2,19 @@
...
-    return
-      string(
-        abi.encodePacked(
-          "data:application/json;base64,",
-          Base64.encode(
-            abi.encodePacked(
-              '{"name":"',
-              ds._name,
-              '","image":"',
-              _getMembershipImage(),
```
texuf, line 41: do we want this other info on chain? If so we could just change the _getMembershipImage() part of this function, otherwise we can return this from the stream metadata service no problem.

brianathere: Earlier conversations were that we'd want metadata about the space in this that would be compute from the streams (maybe cached), similar to what is on the towns landing page, and making that available to the OpenSea market so people could get a sense of the activity of the town would be valuable. That is a tradeoff, as it's not composible into other contracts, but it is more representative  about what's happening in the town. If an Owner wanted to have a return value that was composable, they could set the URI to be their own Base64 encoded JSON and do anything they want onchain, not using the metadata service and image store we provide.

PR #1050, opened 2024-09-11
stream-metadata: opensea refresh all member nft

miguel-nascimento:
should close TOWNS-12029

example: [member nft](https://testnets.opensea.io/collection/member-505)


packages/stream-metadata/src/opensea.ts
```diff
@@ -5,17 +5,123 @@ import { FastifyBaseLogger } from 'fastify'
 import { config } from './environment'
 import { spaceDapp } from './contract-utils'
 
-const getOpenSeaAPIUrl = (logger: FastifyBaseLogger, space: SpaceInfo) => {
+type GetNFTs = {
+	nfts: { identifier: string }[]
+	next?: string
+}
+
+const getAllMemberTokenIds = async (
```
mechanical-turk, line 13: @miguel-nascimento does this function actually "get all member token ids"? do we need to talk to opensea to get a string of all our member token ids? isn't this an array where each element counts up from 0 to totalSupply?

packages/stream-metadata/src/opensea.ts
```diff
@@ -45,25 +151,13 @@ export const refreshOpenSea = async ({
 		throw new Error('Space not found')
 	}
 
-	const url = getOpenSeaAPIUrl(logger, space)
-	logger.info({ url, spaceAddress }, 'refreshing openSea')
+	const allMemberTokenIds = await getAllMemberTokenIds(logger, space)
+	const promises = [
+		refreshSpaceOwnerNft(logger, space),
+		...allMemberTokenIds.map((tokenId) => refreshMemberNft(logger, space, tokenId)),
+	]
```
mechanical-turk, line 158: if a town has 1000 members, are we sending 1000 invalidations? i'm curious to see if opensea offers an alternative. could we maybe refresh the entire contract with one request?

packages/stream-metadata/src/opensea.ts
miguel-nascimento: Thanks for the suggestion, you're right, we can get the members in the space and count 0..members.length

texuf: members in the space !== the number of tokens. they are different and almost guaranteed to get out of sync

packages/stream-metadata/src/opensea.ts
texuf: i'm worried that 1) this could time out and 2) we get rate limited.
open sea provides an on chain even to refresh a collection. seems like they should allow us to do it at the api level too

packages/stream-metadata/src/opensea.ts
mechanical-turk: Should that matter? We're just trying to reset all member nfts. Besides just like we both pointed out in the other comment, resetting all N tokens with N requests seems fragile. So hopefully none of this will be necessary and we can bulk-reset the entire contract with one request 🤞

miguel-nascimento:
We gonna do this on-chain.

PR #1052, opened 2024-09-12
Add memory and cpu usage statistics to debug endpoint

clemire:
Moved existing /debug/memory to /debug/stats and updated like so:

Screenshot 2024-09-17 at 4 37 54 PM


bas-vk:
There is already `/debug/memory` for mem usage?

Can't we rename that to `/debug/stats` and add the cpu usage to it? Add benefit is that the stats endpoint is already using a template that is rendered so we can follow the same approach as the current debug endpoints do.

clemire:
> There is already `/debug/memory` for mem usage?
> 
> Can't we rename that to `/debug/stats` and add the cpu usage to it? Add benefit is that the stats endpoint is already using a template that is rendered so we can follow the same approach as the current debug endpoints do.

So I took a look at that, and the /debug/memory endpoint invokes garbage collection, stops and restarts all go subroutines, and also collects a more specific set of metrics that are related to the server process.

The above endpoint I added collects system-wide metrics and does not affect the state of garbage collection or other go routines on the service, so they do seem different to me and that's what I understood the issue was filed for.

Does that square with you @bas-vk?

bas-vk:
> > There is already `/debug/memory` for mem usage?
> > Can't we rename that to `/debug/stats` and add the cpu usage to it? Add benefit is that the stats endpoint is already using a template that is rendered so we can follow the same approach as the current debug endpoints do.
> 
> So I took a look at that, and the /debug/memory endpoint invokes garbage collection, stops and restarts all go subroutines, and also collects a more specific set of metrics that are related to the server process itself and how go organizes memory usage (as far as I understand it!)
> 
> The above endpoint I added collects system-wide metrics and also does not affect the state of garbage collection or other go routines on the service, so they do seem different to me and that's what I understood the issue was filed for.
> 
> Does that make sense, @bas-vk?

That only happens when `gcnow` is part of the query params? By default it retrieves memory allocator stats only. I do think it is fine to add these metrics to that page and rename it to `/debug/stats`. That page already renders the output through a template which is what the other endpoints also use.

bas-vk (approved):

Issue #507, opened 2024-07-29
fix GetPaginatedStreams not to return empty records


sergekh2:
### Description

```markdown
Right now GetPaginatedStreams returns exact number of records requested, if less records are available, last returned records are empty.
```

PR #1053, opened 2024-09-12
Refactor StreamRegistry.getPaginatedStreams to never return empty streams

clemire:
ref #507 

getPaginatedStreams would return an oversized array in cases where the stop argument was > # of streams. This PR repairs the contract response and adds test coverage for the getPaginatedStreams method.

Once this is deployed in all environments, we can remove the special case code in the stream node that ignores empty stream responses.


contracts/src/river/registry/facets/stream/StreamRegistry.sol
```diff
@@ -236,15 +236,15 @@ contract StreamRegistry is IStreamRegistry, RegistryModifiers {
...
-    for (
-      uint256 i = 0;
-      ((start + i) < streamCount) && ((start + i) < stop);
-      ++i
-    ) {
+    uint256 maxStreamIndex = stop;
+    if (streamCount < stop) {
+      maxStreamIndex = streamCount;
+    }
+    StreamWithId[] memory streams = new StreamWithId[](maxStreamIndex - start);
```
brianathere, line 245: @giuseppecrj @shuhuiluo cant this have been done without a copy using assembly

shuhuiluo: Return data has to be copied to memory.

PR #1057, opened 2024-09-12
set up space factory init for upgrades on diamond cut

giuseppecrj:
This PR helps in using a deployed SpaceFactoryInit contract during diamond cut proposal. Helpful in the event that we need to set state parameters right after creating the diamond cut during governance. For example in this case we need to set a SpaceProxyInitializer contract address right after the cut in order to not break createSpace in production.


contracts/scripts/deployments/diamonds/DeploySpaceFactory.s.sol
```diff
@@ -184,6 +191,10 @@ contract DeploySpaceFactory is DiamondHelper, Deployer {
     // legacy
     legacyArchitect = deployMockLegacyArchitect.deploy(deployer);
     spaceProxyInitializer = deploySpaceProxyInitializer.deploy(deployer);
+    spaceFactoryInit = deploySpaceFactoryInit.deploy(deployer);
+    spaceFactoryInitData = deploySpaceFactoryInit.makeInitData(
```
jterzis, line 195: Is the idea that if we needed to change spaceProxyInitializer or another proxy param after a diamond cut, we'd do so with an action ? Or how would we do so given this pr sets it once as init for every diamond deployment ?

giuseppecrj: the idea is to have a way to set state during the upgrade, if done through an action we would deploy spaceFactoryInit and fill in the init field, we would also be able to set the spaceFactoryInitData in the initPayload field

Screenshot 2024-09-12 at 8 12 00 PM

Since we're not deploying new diamonds anymore and only facets, the current DeploySpaceFactory and other diamond deployment scripts are limited by not being able to set new state. This setup allows for that and works for both alpha, gamma and omega deployments.

PR #1067, opened 2024-09-13
using 307 on space metadata redirect

mechanical-turk:


brianathere (approved):

packages/stream-metadata/src/routes/spaceMetadata.ts
```diff
@@ -90,7 +90,7 @@ export async function fetchSpaceMetadata(request: FastifyRequest, reply: Fastify
 
 	// Not using the default space image service
 	// redirect to the space contract's uri
-	return reply.header('Cache-Control', CACHE_CONTROL[307]).redirect(spaceInfo.uri)
```
texuf, line 93: what was it doing before?
also i forgot that we didn't remove this code.

texuf (approved):

packages/stream-metadata/src/routes/spaceMetadata.ts
mechanical-turk: @texuf i haven't observed the behavior here. I just noticed it was a default 302 which the cdn doesn't cache. But i too noticed this line as being a bit off. If our server is being hit, it shouldn't really matter who the intended server was. We should probably remove it to reduce complexity, because it seems like an edge case that we don't really need to cover.

PR #1068, opened 2024-09-13
Fix memory leak

brianathere:
After a chat with Kerem I decided to look at the code, and sure enough, there is a "leak". The earlier caching code was only partially removed, and the decoded content was being kept in memory.

While at it I added code to remove the rpcClient from the mapping on error to avoid getting into a state where a bad client is a zombie and degrades the service.


mechanical-turk (approved):

texuf (approved):
thank you

PR #1075, opened 2024-09-15
Run local and CI setup through taskfile.yaml

sergekh2:
This enables faster parallel setup and adds flexibility of re-running individual step.
This change doesn't drop old way of running yet.


texuf (approved):
this is neat... we can run this locally too?

Issue #1046, opened 2024-09-11
Update custom entitlement check to take byte-encoded parameters


clemire:

sergekh2:
pre-work for POAP support

Issue #1081, opened 2024-09-16
Xchain cleanup: remove client simulator run command, test contract address from config


clemire:
### Description

The client simulator tool is no longer used to test the xchain service and the config entry for TestEntitlementContract as well as TestCustomEntitlementContract could be removed. This would be nice because we would no longer be exposing test configuration in our production configuration API.

clemire:
This is almost done and I'm just waiting for an entitlement change to land before I rip out the cs command from the server.

PR #1083, opened 2024-09-16
CustomEntitlement -> CrossChainEntitlement implementation in go

clemire:
The `ICrossChainEntitlement` type replaces `ICustomEntitlement` and takes an abi-encoded byte array of parameters that is decoded by the contract according to it's own application-specific logic.

This PR migrates the ISENTITLED check type to expect ICrossChainEntitlement interfaces for ISENTITLED contracts and uses the v2 check type to encode the parameter byte array for passing directly to the contract.

A follow-up PR will be released that implements the typescript change and re-enables e2e tests to complete #1046.


core/contracts/types/entitlement_data_test.go
```diff
@@ -131,24 +131,6 @@ func TestConvertV1RuleDataToV2(t *testing.T) {
 				},
 			},
 		},
-		"Custom": {
```
clemire, line 134: ISENTITLED checks are now v2 only, we don't need to test converting them to v1 rule data.

core/xchain/client_simulator/client_simulator.go
```diff
@@ -17,110 +16,15 @@ import (
...
 
 	"github.com/ethereum/go-ethereum/accounts/abi"
 	"github.com/ethereum/go-ethereum/accounts/abi/bind"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/types"
-	"github.com/ethereum/go-ethereum/crypto"
-	"github.com/ethereum/go-ethereum/ethclient"
 )
 
-var isEntitled = false
```
clemire, line 33: I plan to remove most of what's in this file (see https://github.com/river-build/river/issues/1081) but for now, to keep the PR appropriately scoped, I am just removing ISENTITLED functionality.

bas-vk (approved):

PR #1087, opened 2024-09-17
remove the fee recipient and duration from settings

giuseppecrj:


salzbrenner:
This means the parameters would change for createSpace, correct?
Does that warrant a new method?

PR #1095, opened 2024-09-17
Kerem/cloudfront invalidation optimization

mechanical-turk:
This PR introduces `onResponse` hooks to the stream metadata service, which allows us avoid having the client wait cf invalidations and opensea refreshes.


miguel-nascimento (approved):
🙌

texuf (approved):
neat

PR #1101, opened 2024-09-18
core/node: Add RPC endpoint to add multiple streams at once to sync operation

bas-vk:
Add RPC endpoint that accepts a list of streams to add/remove to/from a pending sync subscription and add them one by one as fast as possible and report any failures back to the client.

### DIFF
--- a/protocol/protocol.proto
+++ b/protocol/protocol.proto
@@ -768,6 +768,34 @@ message RemoveStreamFromSyncRequest {
 
 message RemoveStreamFromSyncResponse {}
 
+// ModifySyncRequest adds or removes streams from an existing sync session.
+message ModifySyncRequest {
+  // sync_id is the id of the sync session.
+  string sync_id = 1;
+  // add_streams contains the stream cookies for streams to add to the stream sync session.
+  repeated SyncCookie add_streams = 2;
+  // remove_streams contains the stream id's to remove from the sync session.
+  repeated bytes remove_streams = 3;
+}
+
+// SyncStreamOpStatus contains details why adding or removing a stream to/from a sync session failed.
+message SyncStreamOpStatus {
+  // stream_id that failed to add/remove to/from the sync session.
+  bytes stream_id = 1;
+  // code represents the error code
+  int32 code = 2;
+  // message describing what happened
+  string message = 3;
+}
+
+// ModifySyncResponse contains the details of a stream sync modification request.
+message ModifySyncResponse {
+  // List with streams that failed to add to the sync operation.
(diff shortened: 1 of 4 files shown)


protocol/protocol.proto
```diff
@@ -805,6 +825,10 @@ service StreamService {
     rpc AddEvent(AddEventRequest) returns (AddEventResponse);
     rpc SyncStreams(SyncStreamsRequest) returns (stream SyncStreamsResponse);
     rpc AddStreamToSync(AddStreamToSyncRequest) returns (AddStreamToSyncResponse);
+    // AddStreamsToSyncRequest adds multiple streams to an in progress sync operation.
+    // The client must check AddStreamsToSyncResponse.errors to determine if some streams failed to add.
+    // Note that it is possible that the client receives stream updates before the calls returned.
+    rpc AddStreamsToSync(AddStreamsToSyncRequest) returns (AddStreamsToSyncResponse);
```
sergekh2, line 831: Lets do `ModifySync` which also can remove streams (this way it will cover all use cases for batch updates).

protocol/protocol.proto
```diff
@@ -758,6 +758,26 @@ message AddStreamToSyncRequest {
...
 message AddStreamToSyncResponse {}
 
+message AddStreamsToSyncRequest {
+  // sync_id is the id of the sync session.
+  string sync_id = 1;
+  // sync_pos is the list of streams and positions in those streams to receive updates from.
+  repeated AddStreamToSyncRequest streams = 2;
+}
+
+message AddStreamsToSyncResponseError {
```
sergekh2, line 768: To match comment below something like this:
```suggestion
message StreamOpStatus {
```
or `StreamOperationStatus`

protocol/protocol.proto
```diff
@@ -758,6 +758,26 @@ message AddStreamToSyncRequest {
 
 message AddStreamToSyncResponse {}
 
+message AddStreamsToSyncRequest {
+  // sync_id is the id of the sync session.
+  string sync_id = 1;
+  // sync_pos is the list of streams and positions in those streams to receive updates from.
+  repeated AddStreamToSyncRequest streams = 2;
```
sergekh2, line 765: It seems a bit redundant to reuse entire proto here (i.e. sync_id is not used in this case). Maybe something like this:
```suggestion
  repeated SyncCookie streams_to_add = 2;
  repeated bytes streams_to_remove = 3;
```

sergekh2 (approved):

texuf (approved):

PR #1106, opened 2024-09-18
WalletLink: use getRootKeyForWallet to check if wallet is linked to any root

salzbrenner:
User A links wallet X.
User B tries to link wallet X.

Previously, this client check would pass b/c it's comparing X to root key B, but X is linked to A, and then the transaction would go on to revert.

Now the client check will fail b/c X is already linked, so no transaction attempt will be made.


giuseppecrj (approved):

texuf (approved):

clemire (approved):

packages/web3/src/v3/WalletLink.ts
```diff
@@ -42,11 +44,7 @@ export class WalletLink {
 
     private async assertAlreadyLinked(rootKey: ethers.Signer, walletAddress: string) {
         const rootKeyAddress = await rootKey.getAddress()
-        const isLinkedAlready = await this.walletLinkShim.read.checkIfLinked(
-            rootKeyAddress,
-            walletAddress,
-        )
-        if (!isLinkedAlready) {
+        if (!(await this.isLinkedAlready(walletAddress))) {
```
clemire, line 47: So this check no longer asserts that the wallet is linked to the specified root key, is that important?

salzbrenner: @clemire thanks for this note.
Correct, I don't think we need to check if a wallet is specifically linked to the root key - we could just check if it's already linked to _any_ wallet (b/c the tx will ultimately fail if this is the case).
I realized the `rootKey` param was still being passed around and is unnecessary, so I updated my PR to remove it where applicable.

packages/web3/src/v3/WalletLink.ts
```diff
@@ -25,31 +25,27 @@ export class WalletLink {
...
-            walletAddress,
-        )
 
-        if (isLinkedAlready) {
+        if (await this.isLinkedAlready(walletAddress)) {
             throw new WalletAlreadyLinkedError()
         }
 
-        return { rootKeyAddress, walletAddress }
+        return { walletAddress }
```
texuf, line 41: kinda weird that these return the wallet address
And I think we can drop the "already" from these function names

texuf (approved):

PR #1116, opened 2024-09-19
feat(contract): add `getSpaceByTokenId` to `SpaceOwner`

shuhuiluo:
Implement `getSpaceByTokenId` to retrieve space addresses by token ID and update corresponding tests. This enhances the `SpaceOwner` facet to allow querying of space addresses directly by token ID, improving data accessibility.


giuseppecrj:
we already have this in space factory

shuhuiluo:
> we already have this in space factory

It doesn't make `SpaceOwner` self-contained. If someone wants to fetch the space info for all spaces, they have to get the total supply from `SpaceOwner`, call `getSpaceByTokenId` on the `SpaceFactory` and come back to `SpaceOwner` for `getSpaceInfo`. That's not intuitive.

giuseppecrj (approved):

PR #1127, opened 2024-09-20
Report some important and safe informaiton about the Postgres connection pool and server in the status page

brianathere:
With this we can detect if the connection pool is approaching exhaustion, the version of PG, the availability of the actual database, and if two nodes are on a common PG database server.


mechanical-turk (approved):
Approved but i want to draw attention to something i noticed while working with blockchain rpc status as reported on the status page. I think this page is at risk for bloat and it's converging to a mini /metrics now. We should probably report these data points via openmetrics anyway. And that should give us the same level of usability as the /status page would - if not more

sergekh2 (approved):

PR #1134, opened 2024-09-20
test(contract): refactor token tests with better fuzzing and bounds

shuhuiluo:
Renamed test functions to indicate fuzz testing and replaced assume checks with bound constraints for parameter values. This ensures more robust and dynamic testing by covering a broader range of input scenarios. Removed unnecessary `console.log` statements to clean up test output.

### DIFF
--- a/contracts/test/river/token/base/RiverBase.t.sol
+++ b/contracts/test/river/token/base/RiverBase.t.sol
@@ -10,7 +10,6 @@ import {IOwnableBase} from "contracts/src/diamond/facets/ownable/IERC173.sol";
 import {IVotes} from "contracts/src/diamond/facets/governance/votes/IVotes.sol";
 
 //libraries
-import {console} from "forge-std/console.sol";
 
 //contracts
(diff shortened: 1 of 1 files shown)


contracts/test/river/token/base/RiverBase.t.sol
```diff
@@ -65,11 +64,19 @@ contract RiverBaseTest is BaseSetup, ILockBase, IOwnableBase {
...
+  function test_fuzz_permit(
+    uint256 alicePrivateKey,
+    uint256 amount,
+    address bob
+  ) external {
+    alicePrivateKey = bound(
+      alicePrivateKey,
+      1,
+      0xfffffffffffffffffffffffffffffffebaaedce6af48a03bbfd25e8cd0364140
+    );
```
shuhuiluo, line 76: https://ethereum.stackexchange.com/questions/164194/is-there-really-2256-private-key

jterzis (approved):

PR #1141, opened 2024-09-23
Prototype of a custom entitlement contract we could deploy to Polygon to allow creation of Market position gated permissions

brianathere:
Uses the Polymarket mainnet addresses. We can deploy and test. Would want to wire up better UI in the client to make it easy to setup and review requirements on Roles


contracts/src/custom/PolymarketEntitlement.sol
```diff
@@ -0,0 +1,83 @@
+// SPDX-License-Identifier: MIT
```
giuseppecrj, line 1: let's move file to `spaces/entitlements/polymarket`

contracts/test/custom/PolymarketEntitlement.t.sol
```diff
@@ -0,0 +1,171 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.23;
+
+import "forge-std/Test.sol";
+import {PolymarketEntitlement, IGnosisSafeProxyFactory} from "contracts/src/custom/PolymarketEntitlement.sol";
+import {IERC1155} from "@openzeppelin/contracts/token/ERC1155/IERC1155.sol";
+
+contract CrossChainEntitlementTest is Test {
```
giuseppecrj, line 8: let's rename to `PolyMarketEntitlementTest`

contracts/test/custom/PolymarketEntitlement.t.sol
```diff
@@ -0,0 +1,171 @@
...
+      address(0x4D97DCd97eC945f40cF65F87097ACe5EA0476045)
+    );
+
+    // Mock the proxyFactory and tokenContract for testing
+    entitlement = new PolymarketEntitlement();
+  }
+
+  function testSingleUserWithSufficientBalance() public {
+    // Mock proxyFactory.computeProxyAddress
+    vm.mockCall(
```
giuseppecrj, line 40: should we just create mocks of these 2 contracts rather than mocking the calls

giuseppecrj (commented):
small nits, can probably make this a fork test instead by adding `onlyForked` to test and calling with polygon

PR #1142, opened 2024-09-23
update xchain support chains list to exclude river chains

jterzis:


clemire (approved):
I'm noting that we're leaving the gnosis chain id here, and assuming we want to add gnosis support on gamma at some future point in order to test POAP.

PR #1154, opened 2024-09-24
Update AND and OR operation evaluation to properly propagate errors

clemire:
Logical AND and ORs were dropping errors and causing false negatives for entitlement evaluations. This PR makes changes Logical operand evaluation to ignore errors only when we definitely know the user was or was not entitled.

specifically, for AND operations:
- if a single child evaluates to false, return false
- otherwise, if either errors, return an error

For OR operations:
- if a single child evaluates to true, return true
- otherwise, if either child operation errors, return an error

### DIFF
--- a/core/xchain/entitlement/check_operation.go
+++ b/core/xchain/entitlement/check_operation.go
@@ -179,11 +179,14 @@ func (e *Evaluator) evaluateMockOperation(
 	if result != nil {
 		return false, result
 	}
-	if op.ChainID.Sign() != 0 {
-		return true, nil
-	} else {
-		return false, nil
+
+	if (op.ContractAddress != common.Address{}) {
+		// Grab last two digits of contract address as a unique identifier of which check
+		// caused the error, for ease of reading.
+		code := fmt.Sprintf("%v", op.ContractAddress)[40:]
+		return false, fmt.Errorf("intentional failure (%v)", code)
 	}
+	return op.ChainID.Sign() != 0, nil
 }
 
 func (e *Evaluator) evaluateIsEntitledOperation(
--- a/core/xchain/entitlement/check_operation.go
+++ b/core/xchain/entitlement/check_operation.go
@@ -182,7 +182,7 @@ func (e *Evaluator) evaluateMockOperation(
 
 	if (op.ContractAddress != common.Address{}) {
 		// Grab last two digits of contract address as a unique identifier of which check
-		// caused the error, for ease of reading.
+		// caused the error, for ease of debugging test cases.
 		code := fmt.Sprintf("%v", op.ContractAddress)[40:]
 		return false, fmt.Errorf("intentional failure (%v)", code)
 	}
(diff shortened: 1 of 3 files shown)


core/xchain/entitlement/check_operation.go
```diff
@@ -179,11 +179,14 @@ func (e *Evaluator) evaluateMockOperation(
...
-	if op.ChainID.Sign() != 0 {
-		return true, nil
-	} else {
-		return false, nil
+
+	if (op.ContractAddress != common.Address{}) {
+		// Grab last two digits of contract address as a unique identifier of which check
+		// caused the error, for ease of debugging test cases.
+		code := fmt.Sprintf("%v", op.ContractAddress)[40:]
+		return false, fmt.Errorf("intentional failure (%v)", code)
```
bas-vk, line 187: You can also do `fmt.Errorf("intentional failure (%x%x)", op.ContractAddress[18], op.ContractAddress[19])` to save on mem allocations.

core/xchain/entitlement/entitlement.go
```diff
@@ -26,6 +26,21 @@ func (e *Evaluator) EvaluateRuleData(
 	return e.evaluateOp(ctx, opTree, linkedWallets)
 }
 
+func isNilOrCancelled(err error) bool {
+	return err == nil || err == context.Canceled
```
bas-vk, line 30: Consider using `errors.Is(err, context.Cancelled)`. This also verifies is err is wrapped around a canceled err which is probably what you want to check for.

core/xchain/entitlement/entitlement.go
```diff
@@ -88,26 +130,39 @@ func (e *Evaluator) evaluateOrOperation(
...
 		}
 		wg.Done()
 	}()
 
 	wg.Wait()
-	return leftResult || rightResult, nil
+	// If at least one child evaluates as entitled, log any errors and return a true result.
+	if leftResult || rightResult {
+		return true, nil
+	} else {
```
bas-vk, line 155: To improve readability you can remove the else part and reduce nesting 1 lvl.

core/xchain/entitlement/entitlement.go
```diff
@@ -88,26 +130,39 @@ func (e *Evaluator) evaluateOrOperation(
...
+		if leftErr != nil && rightErr != nil {
+			return false, fmt.Errorf("%w; %w", leftErr, rightErr)
+		} else {
+			finalErr := leftErr
+			if rightErr != nil {
+				finalErr = rightErr
+			}
+			return false, finalErr
```
bas-vk, line 163: The and operation logic checks if an error was the result of the context cancelled. Should that be checked here as well?

core/xchain/entitlement/entitlement.go
clemire: Nice, thank you!

core/xchain/entitlement/entitlement.go
clemire: Yes, I think so. Probably I just didn't catch this because it didn't occur when I was running tests.

texuf (approved):

PR #1158, opened 2024-09-24
Add additional cache control headers to prevent client flickers, and unnecessary content re-validation

brianathere:
…


mechanical-turk (commented):
@brianathere immutable makes sense. i had to look up stale-while-validate. it seems like we need to set a duration on this directive, e.g:

Cache-Control: max-age=600, stale-while-revalidate=300

brianathere:
> @brianathere immutable makes sense. i had to look up stale-while-validate. it seems like we need to set a duration on this directive, e.g:
> 
> Cache-Control: max-age=600, stale-while-revalidate=300

Good catch. I don't know how to reason about this. I want the client to not flicker when I come back, but to update when something new is found.

mechanical-turk:
@brianathere i think the duration on the cache-while-validate directive sets the time frame for when it's okay to display stale content. 

So max-age determines when the content becomes stale, and cache-while-validate determines when we can no longer serve stale content and must force a fresh load.

brianathere:
> @brianathere i think the duration on the stale-while-validate directive sets the time frame for when it's okay to display stale content.
> 
> So max-age determines when the content becomes stale, and stale-while-validate determines when we can no longer serve stale content and must force a fresh load.
> 
> So I'd set a high value for this.
> 
> But I'm also curious about what you mean by flicker.

When the page loads, the browser needs to check if the /user and /space content has changed. What does it render until the 307 comes back? Wihtout this, it renders nothing, then the 307 comes back and it renders the content. WIth this it renders the content, and if the 307 comes back different, it updates the content. That flicker.

packages/stream-metadata/src/routes/spaceMetadata.ts
```diff
@@ -19,8 +19,8 @@ export interface SpaceMetadataResponse {
 }
 
 const CACHE_CONTROL = {
-	200: 'public, max-age=30, s-maxage=3600',
-	307: 'public, max-age=30, s-max-age=3600', // NOTE: this is called when the space uses a different image service than ours, but a client is requesting the image from our service.
+	200: 'public, max-age=30, s-maxage=3600, stale-while-revalidate=3600',
+	307: 'public, max-age=30, s-max-age=3600, stale-while-revalidate=3600', // NOTE: this is called when the space uses a different image service than ours, but a client is requesting the image from our service.
```
texuf, line 23: I don't think these comments are still valid.

PR #1162, opened 2024-09-25
Add commands to node binary to get and set on-chain config

sergekh2:
Also use it to set local and CI on-chain configs


core/cmd/on_chain_config_cmd.go
```diff
@@ -0,0 +1,399 @@
...
+	opts := &bind.CallOpts{
+		Context: ctx,
+	}
+	settings, err := caller.GetConfiguration(opts, crypto.HashSettingName(key))
+	if err != nil {
+		return err
+	}
+
+	if len(settings) == 0 {
+		return RiverError(Err_INTERNAL, "returned seetings are empty")
```
bas-vk, line 105: typo in seetings

core/cmd/on_chain_config_cmd.go
```diff
@@ -0,0 +1,399 @@
...
+	}
+
+	for _, s := range settings {
+		fmt.Printf("block: %d\n", s.BlockNumber)
+		switch valueType {
+		case "":
+			fmt.Printf("%s\n", hex.EncodeToString(s.Value))
+		case "uint":
+			num, err := crypto.ABIDecodeUint64(s.Value)
```
bas-vk, line 114: `crypto.ABIDecodeUint256` supports bigger numbers

core/cmd/on_chain_config_cmd.go
```diff
@@ -0,0 +1,399 @@
...
+		txs = append(txs, tx)
+	}
+
+	for _, tx := range txs {
+		receipt, err := tx.Wait(ctx)
+		if err != nil {
+			return err
+		}
+		if receipt.Status != types.ReceiptStatusSuccessful {
+			return RiverError(Err_INTERNAL, "transaction failed", "tx", receipt.TxHash)
```
bas-vk, line 285: Maybe capture all transaction results and return a summary instead of returning immediately so there is an overview which transaction did fail and which pass.

core/node/crypto/config.go
```diff
@@ -48,6 +48,34 @@ const (
 	XChainBlockchainsConfigKey                      = "xchain.blockchains"
 )
 
+var (
+	knownOnChainSettingKeys map[string]string
+	onceInitKeys            sync.Once
+)
+
+func AllKnownOnChainSettingKeys() map[string]string {
```
bas-vk, line 56: Should we call this from a package init method?

core/cmd/on_chain_config_cmd.go
sergekh2: but there is a max size that can fit in the actual Config struct, I changed types to ABI type names to avoid confusion, they are now generated directly from the struct.

core/node/crypto/config.go
sergekh2: then it will be always initialized, even if unused.

PR #1170, opened 2024-09-25
add new sync entitlement feature during create space

giuseppecrj:
### Refactor: Modular Entitlement Creation
- `_createEntitlementForRole`: A generalized function for creating entitlements for a role based on given requirements.
- `_createEveryoneEntitlement`: A specialized function for creating an "everyone" entitlement.

### New Feature: Synced Entitlements
- Introduced a `syncEntitlements` boolean in the `MembershipRequirements` struct.
- When `syncEntitlements` is true, member roles are created with the same entitlement requirements as the minter role.
- When false, member roles are created with an "everyone" entitlement.

### Proposals
- Gamma: https://testnets.llama.xyz/orgs/river/base-sepolia/actions/224


brianathere (approved):

giuseppecrj:
Don't think there's "future proof" way to do any of this since we'll be making changes to these functions throughout the lifecycle of the project. We can allow old ways to work by not removing the selectors when doing the diamond cut until iOS catches up with required changes

giuseppecrj:
Synced with @erikolsson on backwards compatible way of allowing this PR to get merged. We won't remove the old createSpace functions from omega diamond and come up with a cadence for releases.

brianathere:
> Synced with @erikolsson on backwards compatible way of allowing this PR to get merged. We won't remove the old createSpace functions from omega diamond and come up with a cadence for releases.

Does this mean that the old selector will report to the old facet, or are you, hopefully, keeping the old implementation in the new facet and adding the new function also to the new facet?

giuseppecrj:
> > Synced with @erikolsson on backwards compatible way of allowing this PR to get merged. We won't remove the old createSpace functions from omega diamond and come up with a cadence for releases.
> 
> Does this mean that the old selector will report to the old facet, or are you, hopefully, keeping the old implementation in the new facet and adding the new function also to the new facet?

old selector should talk to the old facet

contracts/test/spaces/membership/unit/MembershipJoinSpace.t.sol
```diff
@@ -17,7 +17,6 @@ import {Vm} from "forge-std/Test.sol";
 import {MembershipFacet} from "contracts/src/spaces/facets/membership/MembershipFacet.sol";
 
 // debuggging
```
texuf, line 19: you probably also don't need the //debugging comment now

texuf (commented):
Thanks for jumping on this so fast

brianathere:
> > > Synced with @erikolsson on backwards compatible way of allowing this PR to get merged. We won't remove the old createSpace functions from omega diamond and come up with a cadence for releases.
> 
> > 
> 
> > Does this mean that the old selector will report to the old facet, or are you, hopefully, keeping the old implementation in the new facet and adding the new function also to the new facet?
> 
> 
> 
> old selector should talk to the old facet

Please just put the old function on the new facet. Happy to hop on a call and explain why.

giuseppecrj:
> > > > Synced with @erikolsson on backwards compatible way of allowing this PR to get merged. We won't remove the old createSpace functions from omega diamond and come up with a cadence for releases.
> > 
> > 
> > > 
> > 
> > 
> > > Does this mean that the old selector will report to the old facet, or are you, hopefully, keeping the old implementation in the new facet and adding the new function also to the new facet?
> > 
> > 
> > old selector should talk to the old facet
> 
> Please just put the old function on the new facet. Happy to hop on a call and explain why.

we already have the old function on the diamond

PR #1173, opened 2024-09-26
stream-metadata: add retry to rpc client

miguel-nascimento:
Wait a bit and retry if we get ~~any error~~ a bad connection error from the rpc node. 
This will introduce a overhead in the unhappy path, but hopefully we can convert some possible unhappy response to happy responses :D (especially around the `getStream` errors)


texuf (approved):

texuf:
*nit - it's not "any error" it's just known bad connection errors

PR #1177, opened 2024-09-26
stream-metadata: remove cache for 404 (stream not found/image not found)

miguel-nascimento:
We create the space on-chain before creating the stream.
With the 404 cache, we could accidentally hit `/space/{spaceAddress}/image` before the stream creation, causing and error and caching it. We don't want this for now.

### DIFF
--- a/packages/stream-metadata/src/routes/spaceImage.ts
+++ b/packages/stream-metadata/src/routes/spaceImage.ts
@@ -20,7 +20,8 @@ const paramsSchema = z.object({
 
 const CACHE_CONTROL = {
 	307: 'public, max-age=30, s-maxage=3600, stale-while-revalidate=3600',
-	'4xx': 'public, max-age=30, s-maxage=3600',
+	400: 'public, max-age=30, s-maxage=3600',
+	422: 'public, max-age=30, s-maxage=3600',
 }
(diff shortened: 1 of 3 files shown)


packages/stream-metadata/src/routes/spaceImage.ts
```diff
@@ -52,19 +53,14 @@ export async function fetchSpaceImage(request: FastifyRequest, reply: FastifyRep
...
 
 	// get the image metatdata from the stream
 	const spaceImage = await getSpaceImage(stream)
 	if (!spaceImage) {
-		return reply
-			.code(404)
-			.header('Cache-Control', CACHE_CONTROL['4xx'])
-			.send('spaceImage not found')
+		logger.error({ spaceAddress, streamId: stream.streamId }, 'spaceImage not found')
+		return reply.code(404).send('spaceImage not found')
```
texuf, line 63: I feel like this one should be a small cache time, but we can try this for now.

texuf (approved):
I think we’re going to see warnings in datadog for the responses without headers.

miguel-nascimento:
Let’s ignore the warnings for now, let’s check the failed to get image logs too

PR #1179, opened 2024-09-27
Implement stream scrubbing on add_event, get_stream, and stream sync for local streams

clemire:
This PR includes a simple E2E test case and a suite of go tests.

No metrics or profiling as of yet, but I plan to add metrics in a fast follow PR.


core/node/rpc/server.go
```diff
@@ -581,6 +589,49 @@ func (s *Service) initCacheAndSync() error {
...
+		return err
+	}
+
+	go func() {
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			case event := <-s.scrubEventQueue:
+				err := s.addEventPayload(ctx, event.StreamId, event.Payload)
```
sergekh2, line 616: It seems it's possible to remove this channel:
rename addEventPayload to AddEventPayload.
instead of scrubEventQueue pass s.
On the scrub side define something like 
```go
type EventAdder interface {
  AddEventPayload
}
```
and use it instead of channel.

core/node/scrub/stream_scrub_task.go
```diff
@@ -0,0 +1,199 @@
...
+			"Unable to scrub stream; could not fetch stream view",
+			"streamId",
+			task.channelId,
+			"error",
+			err,
+		)
+		return
+	}
+
+	members, err := view.(events.JoinableStreamView).GetChannelMembers()
```
sergekh2, line 80: it's better to defensively check if view implements this interface:
```go
joinable, ok := view.(events.JoinableStreamView)
```

sergekh2 (commented):
great change!
it might a bit challenging to add go test, but on the other hand it seems unfortunate not to have it all, please take a look on how hard to do it.

core/node/events/stream.go
```diff
@@ -25,9 +26,15 @@ type MiniblockStream interface {
 	GetMiniblocks(ctx context.Context, fromInclusive int64, ToExclusive int64) ([]*Miniblock, bool, error)
 }
 
+type ScrubbableStream interface {
```
bas-vk, line 29: Please add some documentation what it means if a stream supports this interface.

core/node/rpc/add_event.go
```diff
@@ -62,6 +63,7 @@ func (s *Service) addParsedEvent(
 	if err != nil {
 		return err
 	}
+	_, _ = s.scrubTaskProcessor.TryScheduleScrub(ctx, streamId)
```
bas-vk, line 66: Is this the right moment to schedule a scrub task? Should it not be done only after the add succeeded?

bas-vk:
In general there seem to be multiple levels of caching to prevent over checking streams when a user doesn't meet the entitlement criteria anymore. Is there an estimate of how many extra RPC calls this will lead?

core/node/scrub/stream_scrub_task.go
```diff
@@ -0,0 +1,199 @@
...
+// - the stream is a channel stream,
+// - the stream has not been recently scrubbed, and
+// - there is no pending scrub for the given stream.
+// If the worker pool is full, the method will not block but will return an error.
+// This is so that we don't affect ability to post updates to channels with stale scrubs
+// whenever the node falls behind due to high scrubbing volume.
+func (tp *streamScrubTaskProcessorImpl) TryScheduleScrub(
+	ctx context.Context,
+	streamId shared.StreamId,
+) (bool, error) {
```
bas-vk, line 171: The return values are ignored everywhere should we not just drop them and do any logging/future metrics in this function?

core/node/events/stream.go
bas-vk: If a stream is joinable does that always imply it is scrubbable? If that is the case it is possible to either move these 2 functions to the `JoinableStreamView` interface or extend the `JoinableStreamView` interface with this interface.

core/node/scrub/stream_scrub_task.go
clemire: I added an unused return value to make this forward compatible with future uses of scrubbing that we may want to error on, say if we expose an RPC endpoint in our API. We're not implementing that at this time, but we've discussed it as an option, so I kept that in mind as a future possibility when designing.

At this time, my strategy was to *also* aggressively log errors from `TryScheduleScrub` since we don't want to raise an error from any current call sites if this fails.

clemire: Metrics yes! I really want those but wanted to do that in a follow up PR.

core/node/events/stream.go
clemire: I moved the timestamp to the view and copy it on copyAndAddEvent and copyAndApplyMiniblock. I also updated the interface name to ScrubTrackable and added some comments. PTAL!

clemire:
> In general there seem to be multiple levels of caching to prevent over checking streams when a user doesn't meet the entitlement criteria anymore. Is there an estimate of how many extra RPC calls this will lead?

No, I don't think we did any back of the napkin work here.

A single entitlement check for a channel can involve quite a view calls, here is a possibly non-exhaustive list:
- Check against space contract that channel is not disabled (cached)
- Fetch all of user's linked wallets (single space contract call, not cached)
- Fetch all entitlements for a channel (cached)
- Check against contract that user is not an owner (call to space to get owner, not cached)
- Check if user is banned (cached)
- Evaluation of each entitlement for each linked wallet (for a rule data with N check operations and M wallets, NxM calls.)

The aggregate result of all of these computations is cached.

This is the rough flow behind the ChainAuth interface and the entire flow is also evaluated every time adding an event requires that entitlements be checked.

We are doing once per hour in production environments to limit the amount of computation we have to do to support booting users on entitlement loss.

clemire:
> great change! it might a bit challenging to add go test, but on the other hand it seems unfortunate not to have it all, please take a look on how hard to do it.

Messaged you on towns!

For sure it would be really great to have some go tests here. I think the big limitation is that entitlement checks require a deployed space contract to evaluate, and our existing test harness for that is the e2e framework expose via the sdk, which has the (admittedly sparse) test I added. This is a bit of a testing gap for our authorization code in general, but it **is** covered by sdk tests. I would personally be interested in talking about building out some support for deploying space contracts from go for integration testing the river node in isolation and triaging that, since I'd love to be able to isolate the cause of entitlement regressions... but I think it's just too much work to fold into this task.

core/node/scrub/stream_scrub_task.go
```diff
@@ -0,0 +1,210 @@
...
+// This is so that we don't affect ability to post updates to channels with stale scrubs
+// whenever the node falls behind due to high scrubbing volume.
+func (tp *streamScrubTaskProcessorImpl) TryScheduleScrub(
+	ctx context.Context,
+	streamId StreamId,
+) (bool, error) {
+	log := dlog.FromCtx(ctx).With("Func", "TryScheduleScrub").With("streamId", streamId)
+	if !ValidChannelStreamId(&streamId) {
+		return false, nil
+	}
```
texuf, line 186: Just a note - if we ever start scrubbing spaces we'll need to make sure we kick the user from all the channels in the space before kicking from the space.

clemire: I added a note here in the method as a reminder.

texuf: ty

sergekh2 (approved):

PR #1194, opened 2024-09-30
stream-metadata: add cache for user bio

miguel-nascimento:
Should close TOWNS-12552.

### DIFF
--- a/packages/stream-metadata/src/aws.ts
+++ b/packages/stream-metadata/src/aws.ts
@@ -20,24 +20,24 @@ export class CloudfrontManager {
 		})
 	}
 
-	private async invalidate(params: { path: string[]; waitUntilFinished?: boolean }) {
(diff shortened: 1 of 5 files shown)


packages/stream-metadata/src/aws.ts
```diff
@@ -78,14 +80,14 @@ export class CloudfrontManager {
...
-				path,
+				paths,
 			},
 			'CloudFront cache invalidation completed',
 		)
 	}
 
 	static async createCloudfrontInvalidation(params: {
-		path: string
+		path: string[]
```
texuf, line 90: why not rename this one to paths?

texuf (approved):

miguel-nascimento:
re-requesting review - increased the cache duration to 1 year

packages/stream-metadata/src/aws.ts
miguel-nascimento: fixed

texuf (approved):

PR #1197, opened 2024-09-30
Increase runner configuration to ubuntu-x64-16core for all CI jobs

sergekh2:


mechanical-turk (approved):
@sergekh2 approved as long as we observe a noticeable difference in time to completion. Let's not blow up the bill for no reason

jterzis (approved):

PR #1198, opened 2024-09-30
Increase CI workflow concurrency by switching to 32-core machines.

sergekh2:


mechanical-turk:
@sergekh2 same as 16 core. approved as long as we observe a noticeable difference in time to completion

jterzis (approved):

PR #1203, opened 2024-10-01
Add tags message to StreamEvent, populate with channel message data

texuf:
i added a react/reply indicator
a group mention type, which at the moment is just @channel but could support @everyone or @here or @role later
a mentions array for users mentioned
a participating array that grabs people who have replied to a thread or emoji'd the original message

work i expect the notification service to do
- figure out who gets notified for @channel
- figure out if this is a dm or gdm and notify accordingly

things i just skipped
- there's all this weird code about attachment notifications, i.e. "attachmentUsersTagged" that seems like it should just never happen, so i've ignored all that. 
- attachment mime types... seems like you should just decrypt things to figure that out.

### DIFF
--- a/protocol/protocol.proto
+++ b/protocol/protocol.proto
@@ -662,7 +662,7 @@ message Tags {
     MessageInteractionType message_interaction_type = 1;
     repeated GroupMentionType group_mention_types = 2;
     repeated bytes mentioned_user_addresses = 3;
-    repeated bytes participating_user_ids = 4;
+    repeated bytes participating_user_addresses = 4;
 }
 
 // GetStreamExResponse is a stream of raw data that represents the current state of the requested stream.
--- a/protocol/protocol.proto
+++ b/protocol/protocol.proto
@@ -85,6 +85,9 @@ message StreamEvent {
     /** DelegateExpiry is the time when the delegate signature expires. */
     int64 delegate_expiry_epoch_ms = 6;
 
+    /** Tags are optional, used for external systems to deliver notifications. */
+    Tags tags = 7;
+
     /** Variable-type payload. 
       * Payloads should obey the following rules:
       * - payloads should have their own unique type
@@ -137,9 +140,6 @@ message MiniblockHeader {
 
     // pointer to block with previous snapshot
     int64 prev_snapshot_miniblock_num = 7;
-
-    // optional metadata tags for external systems
-    Tags tags = 8;
 
     // stream payloads are required to have a content field
     oneof content {
@@ -659,7 +659,10 @@ message Minipool {
 }
 
 message Tags {
-    map<string, string> tags = 1;
+    MessageInteractionType message_interaction_type = 1;
+    GroupMentionType group_mention_type = 2;
+    repeated bytes mentioned_user_addresses = 3;
+    repeated bytes participating_user_ids = 4;
 }
 
 // GetStreamExResponse is a stream of raw data that represents the current state of the requested stream.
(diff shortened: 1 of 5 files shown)


protocol/protocol.proto
```diff
@@ -85,6 +85,9 @@ message StreamEvent {
     /** DelegateExpiry is the time when the delegate signature expires. */
     int64 delegate_expiry_epoch_ms = 6;
 
+    /** Tags are optional, used for external systems to deliver notifications. */
+    Tags tags = 7;
```
bas-vk, line 89: If these are optional should we mark them as optional in the protobuf file?

protocol/protocol.proto
```diff
@@ -659,7 +659,10 @@ message Minipool {
 }
 
 message Tags {
-    map<string, string> tags = 1;
+    MessageInteractionType message_interaction_type = 1;
+    GroupMentionType group_mention_type = 2;
+    repeated bytes mentioned_user_addresses = 3;
+    repeated bytes participating_user_ids = 4;
```
bas-vk, line 665: What is the difference between an `address` and `user_id`?

sergekh2: (rest of the file uses _address to refer to the user)

protocol/protocol.proto
```diff
@@ -659,7 +659,10 @@ message Minipool {
 }
 
 message Tags {
-    map<string, string> tags = 1;
+    MessageInteractionType message_interaction_type = 1;
+    GroupMentionType group_mention_type = 2;
```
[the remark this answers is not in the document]
sergekh2: my take is that @role need to be resolved at the client first and converted to user arrays?

protocol/protocol.proto
sergekh2: no, in proto3 all fields are optional by default and `optional` keyword is reused for other functionality

protocol/protocol.proto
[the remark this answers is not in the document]
texuf: @person is represented in the mentioned_user_addresses
@role - roles are complicated, it's not really feasible for the client to calculate roles because they have to do a nasty lookup on every wallet for every user in a channel... I think roles will end up getting cached in the stream-metadata service. I think i will remove them from the payloads.proto file just to force us to think about it in the future
@channel - that's here, it's currently the only supported GroupMentionType

Ultimately you are right though - this should be a `repeated GroupMentionType group_mention_type`, i'll fix it.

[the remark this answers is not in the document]
texuf: if we need role mentions and we don't implement them as regular mentions, we can add `repeated RoleMention role_mentions = 5` to capture the role id and anything else needed to broadcast to the role.

protocol/protocol.proto
texuf: shoot - yeah i'll fix that.

tak-hntlabs:
> 
> * there's all this weird code about attachment notifications, i.e. "attachmentUsersTagged" that seems like it should just never happen, so i've ignored all that.
> * attachment mime types... seems like you should just decrypt things to figure that out.

The attachment kind was added so that the Service Worker is able to show some text to the user: https://linear.app/hnt-labs/issue/TOWNS-5716/notification-when-someone-sends-only-an-image-or-file-should-have-text

Is there enough info in an added event for the SW to craft the text message "@who sent a gif" ?

protocol/protocol.proto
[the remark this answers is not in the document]
tak-hntlabs: if the sender tags multiple @channel, @person, -> the SW should only show the most specific tag. NS should send all the tags, in my opinion, and let the client decide based on its UX design.

@role is very tricky, especially roles with token entitlements. Users can sell or buy assets at any time, and may be expecting or not expecting notifications.

tak-hntlabs (approved):

PR #1216, opened 2024-10-03
wait for cache invalidation to complete

erikolsson:
Would this have any negative side effects? It's currently tricky to invalidate local cache and reload images since we have no idea whether the invalidation has completed on Cloudfront's end.

### DIFF
--- a/packages/stream-metadata/src/routes/userRefresh.ts
+++ b/packages/stream-metadata/src/routes/userRefresh.ts
@@ -39,7 +39,11 @@ export async function userRefreshOnResponse(
 
 	try {
 		const path = `/user/${userId}/image`
-		await CloudfrontManager.createCloudfrontInvalidation({ path, logger })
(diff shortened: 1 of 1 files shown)


brianathere (commented):
Kerem I believe noticed this can take a significant amount of time.

What specifically is the issue you're trying to solve, that the user doesn't see the image they just set?

tak-hntlabs:
https://repost.aws/knowledge-center/cloudfront-serving-outdated-content-s3 

"Object invalidations **typically** take from 10 to 100 seconds to complete. "

Almost 2 minutes is quite a long wait for a user. Is this for the user who has just updated his / her own profile pic ?

texuf:
@erikolsson you know the id of the media stream for the current user. You can directly fetch the image via /media/{streamId}?key=key&iv=iv that you grab from the users metadata stream.

PR #1218, opened 2024-10-03
[TOWNS-12637] Update RuleEntitlementV2 to revert when zero length rule data is passed in

clemire:
Previously setEntitlement RuleEntitlementV2 ignored empty rule datas, but this created a bit of confusion with createSpace when the architect assumed a rule data was set when creating the minter role.

Here we update the contract to just always revert when setEntitlement doesn't actually set an entitlement as an extra guard to avoid creating spaces with an invalid view of entitlements on the minter role.

### DIFF
--- a/contracts/src/spaces/entitlements/rule/RuleEntitlementV2.sol
+++ b/contracts/src/spaces/entitlements/rule/RuleEntitlementV2.sol
@@ -117,7 +117,11 @@ contract RuleEntitlementV2 is
   ) external onlySpace {
     _removeRuleDataV1(roleId);
 
-    if (entitlementData.length == 0) return;
+    // We should never allow the setting of empty rule datas because it can cause the xchain
+    // architecture to be invoked when by default a user is not entitled.
+    if (entitlementData.length == 0) {
+      revert Entitlement__InvalidValue();
+    }
 
     // equivalent: abi.decode(entitlementData, (RuleDataV2))
     RuleDataV2 calldata data;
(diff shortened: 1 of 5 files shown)


contracts/src/spaces/entitlements/rule/RuleEntitlementV2.sol
```diff
@@ -127,7 +131,11 @@ contract RuleEntitlementV2 is
       data := add(entitlementData.offset, calldataload(entitlementData.offset))
     }
 
-    if (data.operations.length == 0) return;
+    // We should never allow the setting of empty rule datas because it can cause the xchain
```
clemire, line 134: This is the part that would have caused a revert before, but I'm adding reverts everywhere we don't set an entitlement to create an invariant that setEntitlement always sets an entitlement. This is already true for UserEntitlement.

giuseppecrj:
Note to self: Need to deploy and update Architect to use this new module for new spaces

jterzis (approved):

contracts/test/crosschain/RuleEntitlementV2.t.sol
```diff
@@ -75,6 +75,27 @@ contract RuleEntitlementV2Test is RuleEntitlementTest {
     assertEq(vm.getMappingLength(entitlement, bytes32(ENTITLEMENTS_SLOT)), 0);
   }
 
+  function test_setRuleEntitlement_revertOnEmptyRuleData() public {
+    test_upgradeToRuleV2();
+
+    vm.expectRevert(abi.encodeWithSelector(Entitlement__InvalidValue.selector));
```
shuhuiluo, line 81:
```suggestion
    vm.expectRevert(Entitlement__InvalidValue.selector);
```
nit

contracts/test/crosschain/RuleEntitlementV2.t.sol
```diff
@@ -75,6 +75,27 @@ contract RuleEntitlementV2Test is RuleEntitlementTest {
...
+
+    vm.prank(space);
+
+    ruleEntitlementV2.setEntitlement(roleId, "");
+  }
+
+  function test_setRuleEntitlement_revertOnZeroLengthRuleData() public {
+    test_upgradeToRuleV2();
+
+    vm.expectRevert(abi.encodeWithSelector(Entitlement__InvalidValue.selector));
```
shuhuiluo, line 91:
```suggestion
    vm.expectRevert(Entitlement__InvalidValue.selector);
```
nit

shuhuiluo (approved):

PR #1222, opened 2024-10-04
sdk(sync-agent): handler for token expiry

miguel-nascimento:
Should close TOWNS-11629


packages/react-sdk/src/useRiverConnection.ts
```diff
@@ -4,18 +4,26 @@ import type { ethers } from 'ethers'
...
             if (river?.syncAgent) {
                 return
             }
-
+            const mergedConfig = {
+                ...config,
+                ...river?.config,
+                onTokenExpired: () => {
+                    river?.config?.onTokenExpired?.()
+                    river?.setSyncAgent(undefined)
```
miguel-nascimento, line 22: Unsure if we want this to the default case - lets see

packages/sdk/src/rpcInterceptors.ts
```diff
@@ -78,6 +78,20 @@ export const retryInterceptor: (retryParams: RetryParams) => Interceptor = (
...
 
+export const expiryInterceptor = (opts: { onTokenExpired?: () => void }): Interceptor => {
+    return (next) => async (req) => {
+        try {
+            const res = await next(req)
+            return res
+        } catch (e) {
+            if (errorContains(e, Err.PERMISSION_DENIED) && req.method.name === 'AddEvent') {
+                opts.onTokenExpired?.()
+            }
```
texuf, line 89: I really like the interceptor. This condition isn't correct though. There are more places than just AddEvent that throw... can you check for the string "event delegate has expired"? if not then we need a new error on the nodes.

packages/react-sdk/src/useRiverConnection.ts
texuf: my token expired in marbella, and the app was still usable for reads, you just can't post or create streams. 
i think this will be jarring because it will seem like your app is working, then when you go to post a message everything will just reset to nothing yeah? A modal would probably be better, but we can improve it as we go.

miguel-nascimento: >  A modal would probably be better

the react-sdk will expose the `onTokenExpired` callback, so you can pop a modal or other UI element to indicate this.

But now we have two situations:
1. let the dev control when the sync agent should reset
2. we control when the sync agent should reset

I need to think more about this

packages/sdk/src/rpcInterceptors.ts
miguel-nascimento: Nice, event delegate has expired string matches what we want here. Would be great to have the error on the node itself, but I would like to model this first https://github.com/river-build/river/pull/1222#discussion_r1801194160 and then clean the remaining things. wdyt?

texuf (approved):

PR #1225, opened 2024-10-04
sdk(sync-agent): add gdm collection and gdm model

miguel-nascimento:
GDM is similar to a channel, but it doesnt belong to a space.

should close: TOWNS-12732


packages/sdk/src/sync-agent/gdms/models/gdm.ts
```diff
@@ -0,0 +1,133 @@
...
+            replyId?: string
+            mentions?: PlainMessage<ChannelMessage_Post_Mention>[]
+            attachments?: PlainMessage<ChannelMessage_Post_Attachment>[]
+        },
+    ): Promise<{ eventId: string }> {
+        const channelId = this.data.id
+        const result = await this.riverConnection.withStream(channelId).call((client) => {
+            return client.sendChannelMessage_Text(channelId, {
+                threadId: options?.threadId,
+                threadPreview: options?.threadId ? '🙉' : undefined,
```
clemire, line 66: nit: Do we want this monkey emoji to be the official threadPreview metadata for this message?

clemire (commented):
I'm not super familiar with this code, but... looks good!

packages/sdk/src/sync-agent/gdms/models/gdm.ts
texuf: no, we don't. but we already have this code somewhere else. I'll make a linear ticket to track.

texuf (approved):

PR #1227, opened 2024-10-07
stress: logs to graph test results

miguel-nascimento:
Closes: TOWNS-12692

I still need to play more with this to check if it's enough.
Usually, if something fails, it will be on the join chat step.

### DIFF
--- a/packages/stress/src/mode/chat/statsReporter.ts
+++ b/packages/stress/src/mode/chat/statsReporter.ts
@@ -1,8 +1,5 @@
 import type { StressClient } from '../../utils/stressClient'
 import { ChatConfig } from '../common/types'
-import { getLogger } from '../../utils/logger'
-
-const logger = getLogger('stress:statsReporter')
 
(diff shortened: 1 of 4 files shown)


texuf (approved):

packages/stress/src/mode/chat/statsReporter.ts
```diff
@@ -1,57 +1,74 @@
...
+            step: string,
+            isSuccess: boolean,
+            metadata?: Record<string, unknown>,
+        ) => {
+            client.logger.info(
+                {
+                    sequence: 'STRESS_RESULT',
+                    step,
+                    result: isSuccess ? 'PASS' : 'FAIL',
+                    sessionId: client.logId,
```
texuf, line 17: this is the wrong session id - we want all processes to report the same session Id

mechanical-turk (approved):

texuf (approved):

PR #1228, opened 2024-10-07
fix(contract): remove token delegation threshold and fix lock

shuhuiluo:
Eliminated the token threshold mechanism from the `River` contract, including the removal of the `MIN_TOKEN_THRESHOLD` variable and the `setTokenThreshold` function. Updated related tests to reflect these changes, ensuring consistent functionality throughout the contract.

Replaced `IntrospectionBase` and `LockBase` with `IntrospectionFacet` and `LockFacet` for better modularization. Removed redundant lock methods and corrected minor typos in comments for improved code clarity and maintainability.

Update logic in `_setDelegators` to handle edge cases for delegate assignment more efficiently. Simplify `River._delegate` by reusing the current delegate storage.

Fix delegation lock and add tests.

### DIFF
--- a/contracts/src/tokens/lock/LockBase.sol
+++ b/contracts/src/tokens/lock/LockBase.sol
@@ -24,10 +24,6 @@ abstract contract LockBase is ILockBase {
   function _enableLock(address caller) internal {
     LockStorage.Layout storage ds = LockStorage.layout();
 
-    if (ds.enabledByAddress[caller]) {
-      revert LockAlreadyEnabled();
-    }
-
     ds.enabledByAddress[caller] = true;
 
     emit LockUpdated(caller, true, 0, block.timestamp);
@@ -35,10 +31,6 @@ abstract contract LockBase is ILockBase {
 
   function _disableLock(address caller) internal {
     LockStorage.Layout storage ds = LockStorage.layout();
-
-    if (ds.enabledByAddress[caller] == false) {
-      revert LockAlreadyDisabled();
-    }
 
     ds.enabledByAddress[caller] = false;
     ds.cooldownByAddress[caller] = block.timestamp + ds.defaultCooldown;
--- a/contracts/src/tokens/lock/LockBase.sol
+++ b/contracts/src/tokens/lock/LockBase.sol
@@ -30,7 +30,7 @@ abstract contract LockBase is ILockBase {
 
     ds.enabledByAddress[caller] = true;
 
-    emit LockUpdated(caller, true, 0, block.timestamp);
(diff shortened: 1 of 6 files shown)


contracts/src/tokens/lock/LockBase.sol
```diff
@@ -36,10 +32,6 @@ abstract contract LockBase is ILockBase {
   function _disableLock(address caller) internal {
     LockStorage.Layout storage ds = LockStorage.layout();
 
-    if (ds.enabledByAddress[caller] == false) {
```
giuseppecrj, line 39: Why are we removing this?

shuhuiluo: Can leave it. But there's no code path using it.

giuseppecrj (approved):

contracts/src/tokens/lock/LockBase.sol
```diff
@@ -24,31 +25,25 @@ abstract contract LockBase is ILockBase {
...
-    if (ds.enabledByAddress[caller]) {
-      revert LockAlreadyEnabled();
-    }
+    if (ds.enabledByAddress[caller])
+      CustomRevert.revertWith(LockAlreadyEnabled.selector);
 
     ds.enabledByAddress[caller] = true;
 
-    emit LockUpdated(caller, true, 0, block.timestamp);
+    emit LockUpdated(caller, true, 0);
```
jterzis, line 33: Why remove block.timestamp ? Also, are we sure nothing is depending on this current event signature that will break after this pr ?

jterzis:
@shuhuiluo How do you intend to deploy the changes to the River contracts, which are not setup as diamonds in base, mainnet given the token has already been created ? A short runbook of how these changes will make it to production and effects on current token distribution should be added to this PR.

contracts/src/tokens/lock/LockBase.sol
shuhuiluo: Events can only be accessed offchain. When queried offchain, event logs come with meta data such as block number and contract address already.
https://github.com/wevm/viem/blob/afd1f787e2f69f599cc5cd943e28b0c39fe016c9/src/types/log.ts#L15

Nothing is using the event in the codebase.

shuhuiluo:
> @shuhuiluo How do you intend to deploy the changes to the River contracts, which are not setup as diamonds in base, mainnet given the token has already been created ? A short runbook of how these changes will make it to production and effects on current token distribution should be added to this PR.

The token contract will need to be redeployed on Base. @pfives will make a migration plan.

jterzis (approved):

PR #1234, opened 2024-10-09
Reduce overhead of cache of spaces in SpaceRegistrar

texuf:
we don't really need to cache spaces, but they instantiate a lot of objects for the contracts and it's worth not wasting memory if we need to access the same space multiple times this code is also used on the server so we don't want to cache spaces for too long

Use lru cache and limit to 100 in memory at any one time


tak-hntlabs (approved):

packages/web3/src/v3/SpaceRegistrar.ts
```diff
@@ -41,13 +41,30 @@ export class SpaceRegistrar {
...
+        this.spaces[spaceId] = { lastSeen: new Date().getTime(), space }
+        return space
+    }
+
+    private pruneSpaces(): void {
+        // clear out spaces that haven't been seen in 5 seconds
+        const fiveSecondsAgo = new Date().getTime() - 1000 * 5
+        for (const spaceId in this.spaces) {
+            if (this.spaces[spaceId].lastSeen < fiveSecondsAgo) {
+                delete this.spaces[spaceId]
```
mechanical-turk, line 58: @texuf this turns each `get` operation from O(1) to O(n) - though n approaches to zero every 5 seconds which is okay i guess. still, i think we'd be better off using something like this: https://www.npmjs.com/package/lru-cache

texuf: that's a very good point...

texuf: fun tidbit from the documentation:
```
If you truly wish to use a cache that is bound only by TTL expiration, consider using a Map object, and calling setTimeout to delete entries when they expire. It will perform much better than an LRU cache.
```
there's even an example

mechanical-turk: ha! i like that a lot

mechanical-turk: @texuf though technically we need to bound primarily by space and not necessarily by time. as long as the cache size does not exceed a certain limit i wouldn't care when the cache gets cleared but i would definitely care about not exceeding a max space limit regardless of ttl

texuf: that's a good point. LRU would be better in that regard.

texuf: updated to lru-cache

PR #1236, opened 2024-10-09
Reduce overhead of cache of spaces in SpaceRegistrar

texuf:
we don't really need to cache spaces, but they instantiate a lot of objects for the contracts and it's worth not wasting memory if we need to access the same space multiple times this code is also used on the server so we don't want to cache spaces for too long

### DIFF
--- a/packages/encryption/src/groupEncryption.ts
+++ b/packages/encryption/src/groupEncryption.ts
@@ -42,6 +42,15 @@ export class GroupEncryption extends EncryptionAlgorithm {
 
             if (opts?.awaitInitialShareSession === true) {
                 await promise
+            } else {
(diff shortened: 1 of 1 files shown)


packages/encryption/src/groupEncryption.ts
```diff
@@ -42,6 +42,15 @@ export class GroupEncryption extends EncryptionAlgorithm {
 
             if (opts?.awaitInitialShareSession === true) {
                 await promise
+            } else {
+                // await the promise but timeout after 15 seconds
```
miguel-nascimento, line 46: nit: 15s in comment, 30s in code

miguel-nascimento (approved):

PR #1237, opened 2024-10-09
stress-tests: use pino for logs

miguel-nascimento:
Moving out from `dlog` since we need a better structured log solution.
Should close: TOWNS-12820

### DIFF
--- a/packages/stress/src/demo.ts
+++ b/packages/stress/src/demo.ts
@@ -35,22 +35,22 @@ async function spamInfo(count: number) {
         riverRegistry.getOperationalNodeUrls(),
     )
     for (let i = 0; i < count; i++) {
-        logger.info(`iteration ${i}`)
+        logger.debug(i, 'iteration')
         const info = await rpcClient.info(new InfoRequest({}), {
             timeoutMs: 10000,
         })
         logger.info(getSystemInfo(), 'system info')
-        logger.info(info, `info ${i}`)
+        logger.info({ info, iteration: i }, `info`)
     }
 }
 
 async function sendAMessage() {
-    logger.info('=======================send a message - start =======================')
+    logger.debug('=======================send a message - start =======================')
     const bob = await makeStressClient(config, 0, getRootWallet(), undefined)
     const { spaceId, defaultChannelId } = await bob.createSpace("bob's space")
     await bob.sendMessage(defaultChannelId, 'hello')
 
-    logger.info('=======================send a message - make alice =======================')
+    logger.debug('=======================send a message - make alice =======================')
     const alice = await makeStressClient(config, 1, undefined, undefined)
     await bob.spaceDapp.joinSpace(
         spaceId,
@@ -58,22 +58,22 @@ async function sendAMessage() {
         bob.baseProvider.wallet,
     )
     await alice.joinSpace(spaceId, { skipMintMembership: true })
-    logger.info('=======================send a message - alice join space =======================')
+    logger.debug('=======================send a message - alice join space =======================')
     const channel = await alice.streamsClient.waitForStream(defaultChannelId)
-    logger.info('=======================send a message - alice wait =======================')
+    logger.debug('=======================send a message - alice wait =======================')
     await waitFor(() => channel.view.timeline.filter(isDecryptedEvent).length > 0)
-    logger.info('alices sees: ', channel.view.timeline.filter(isDecryptedEvent))
-    logger.info('=======================send a message - alice sends =======================')
+    logger.debug('alices sees: ', channel.view.timeline.filter(isDecryptedEvent))
+    logger.debug('=======================send a message - alice sends =======================')
     await alice.sendMessage(defaultChannelId, 'hi bob')
-    logger.info('=======================send a message - alice sent =======================')
+    logger.debug('=======================send a message - alice sent =======================')
     const bobChannel = await bob.streamsClient.waitForStream(defaultChannelId)
-    logger.info('=======================send a message - bob wait =======================')
+    logger.debug('=======================send a message - bob wait =======================')
     await waitFor(() => bobChannel.view.timeline.filter(isDecryptedEvent).length > 0) // bob doesn't decrypt his own messages
-    logger.info('bob sees: ', bobChannel.view.timeline.filter(isDecryptedEvent))
+    logger.debug(bobChannel.view.timeline.filter(isDecryptedEvent), 'bob sees')
 
     await bob.stop()
     await alice.stop()
-    logger.info('=======================send a message - done =======================')
+    logger.debug('=======================send a message - done =======================')
 }
 
 async function encryptDecrypt() {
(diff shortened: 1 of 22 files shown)


packages/stress/src/utils/storage.ts
```diff
@@ -1,7 +1,7 @@
-import { dlogger } from '@river-build/dlog'
 import Redis from 'ioredis'
+import { getLogger } from './logger'
 
-const logger = dlogger('stress:storage')
+const logger = getLogger('storage')
```
mechanical-turk, line 4: @miguel-nascimento we may want to keep the `stress:x` namespacing. this was an artifact from dlog which internally depends on the `debug` npm library, which allowed us to filter logs not just based on log levels but also log namespaces. it's not urgent to implement this functionality for stress tests with pino, but we should probably at least leave the namespaces there. see screenshot for references to this namespace:

image

packages/stress/src/utils/stressClient.ts
```diff
@@ -80,7 +82,7 @@ export class StressClient {
         public globalPersistedStore: IStorage | undefined,
         public storageKey: string,
     ) {
-        logger.log('StressClient', {
+        logger.info('StressClient', {
             clientIndex,
```
mechanical-turk, line 86: @miguel-nascimento we should ideally have all logs coming from the same client get annotated by the same clientId. i think we can set a class property on `StressClient.logger = getLogger('stress:stressClient', { clientId: this.clientId }` and have all class methods use `this.logger`

packages/stress/src/start.ts
```diff
@@ -1,27 +1,31 @@
...
 check(isSet(process.env.PROCESS_INDEX), 'process.env.PROCESS_INDEX')
 const processIndex = parseInt(process.env.PROCESS_INDEX)
 
 const config = makeRiverConfig(process.env.RIVER_ENV)
-const logger = dlogger(`stress:run:${processIndex}`)
-logger.log('======================= run =======================')
+const logger = getLogger(`stress:run`, { processIndex })
+logger.info('======================= run =======================')
 
 if (processIndex === 0) {
```
mechanical-turk, line 19: @texuf can you remind me why we only log this when the process index is zero?

packages/stress/src/utils/logger.ts
```diff
@@ -0,0 +1,37 @@
...
+        },
+    },
+}
+
+const baseLogger = pino(pinoOptions)
+
+export function getLogger(name: string, meta: Record<string, unknown> = {}) {
+    return baseLogger.child({
+        name,
+        ...meta,
```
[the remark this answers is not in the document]
mechanical-turk: we should also add `env.PROCESS_INDEX`

packages/stress/src/mode/chat/sumarizeChat.ts
```diff
@@ -1,19 +1,20 @@
...
+const logger = getLogger('stress:summarizeChat')
 export async function sumarizeChat(
     localClients: StressClient[],
     cfg: ChatConfig,
     errors: unknown[],
 ) {
-    const logger = dlogger('stress:sumarizeChat')
     const processLeadClient = localClients[0]
-    logger.log('sumarizeChat', processLeadClient.userId)
+    logger.debug('sumarizeChat', processLeadClient.userId)
```
mechanical-turk, line 17: `logger.debug(processLeadClient, 'sumarizeChat')`

packages/stress/src/mode/chat/kickoffChat.ts
```diff
@@ -66,33 +67,33 @@ export async function kickoffChat(rootClient: StressClient, cfg: ChatConfig) {
...
-        logger.log('minting membership for', i, wallet.address, 'has', hasSpaceMembership)
+        logger.debug('minting membership for', i, wallet.address, 'has', hasSpaceMembership)
         if (!hasSpaceMembership) {
             const result = await rootClient.spaceDapp.joinSpace(
                 spaceId,
                 wallet.address,
                 rootClient.baseProvider.wallet,
             )
-            logger.log('minted membership', result)
+            logger.debug('minted membership', result)
```
mechanical-turk, line 77: logger.debug(result, 'minted membership')

packages/stress/src/mode/chat/kickoffChat.ts
```diff
@@ -66,33 +67,33 @@ export async function kickoffChat(rootClient: StressClient, cfg: ChatConfig) {
             spaceId,
             wallet.address,
         )
-        logger.log('minting membership for', i, wallet.address, 'has', hasSpaceMembership)
+        logger.debug('minting membership for', i, wallet.address, 'has', hasSpaceMembership)
```
mechanical-turk, line 70: logger.debug({
  index: i,
  walletAddress: wallet.address,
  hasSpaceMembership,
}, 'minting membership')

packages/stress/src/mode/chat/joinChat.ts
```diff
@@ -1,18 +1,18 @@
...
 
 export async function joinChat(client: StressClient, cfg: ChatConfig) {
-    const logger = dlogger(`stress:joinChat:${client.logId}}`)
+    const logger = getLogger('stress:joinChat', { logId: client.logId })
     // is user a member of all the channels?
     // is user a member of the space?
     // does user exist on the stream node?
 
-    logger.log('joinChat', client.userId)
+    logger.info(client.userId, 'joinChat')
```
mechanical-turk, line 15: logger.info({userId: client.userId}, 'joinChat')

packages/stress/src/demo.ts
```diff
@@ -34,45 +35,45 @@ async function spamInfo(count: number) {
         riverRegistry.getOperationalNodeUrls(),
     )
     for (let i = 0; i < count; i++) {
-        logger.log(`iteration ${i}`)
+        logger.info(`iteration ${i}`)
```
mechanical-turk, line 38: @miguel-nascimento we generally want to avoid using template strings in the log message and reserve all variation in the metadata object. so something like:

`logger.info({ spamInfoLoopIndex: i }, 'spamInfo loop iteration')`

packages/stress/src/demo.ts
```diff
@@ -34,45 +35,45 @@ async function spamInfo(count: number) {
...
         const info = await rpcClient.info(new InfoRequest({}), {
             timeoutMs: 10000,
         })
-        printSystemInfo(logger)
-        logger.log(`info ${i}`, info)
+        logger.info(getSystemInfo(), 'system info')
+        logger.info(info, `info ${i}`)
```
mechanical-turk, line 43: same here:

logger.info({info, iteration: i}, 'info')

miguel-nascimento:
Thanks for the review! 🙏

packages/stress/src/demo.ts
mechanical-turk: in case you're curious, the reason we want this is to make log querying and log-derived metrics as simple as possible. it allows us to search for a simple static string

miguel-nascimento:
I pushed a commit fixing the comments. 
I feel that we need to abstract a bit the steps and logging around steps (`joinSlowChat`, `chitChat`, `summarizeChat`, etc), since they're not following a pattern __yet__

I'll address this in another PR (#1227 - need to update to use pino and I would like to create a "lifecycle" for steps)

packages/stress/src/demo.ts
```diff
@@ -35,45 +35,45 @@ async function spamInfo(count: number) {
         riverRegistry.getOperationalNodeUrls(),
     )
     for (let i = 0; i < count; i++) {
-        logger.info(`iteration ${i}`)
+        logger.debug(i, 'iteration')
```
mechanical-turk, line 38: @miguel-nascimento if we care about seeing the "i" value in log metadata we need to make it a key-value pair e.g { iteration: i }

miguel-nascimento: ooh I forgot about this one

PR #1239, opened 2024-10-10
stress-test: create gdm chats each run

miguel-nascimento:
Should close TOWNS-12690


packages/stress/src/mode/chat/root_chat.ts
```diff
@@ -126,6 +131,57 @@ export async function startStressChat(opts: {
...
+                    })
+                    .catch((e) => {
+                        errors.push(e)
+                        logStep(client, 'GDM_CHAT', false, { reason: e })
+                        throw e
+                    }),
+            )
+            return Promise.allSettled(promises)
+        }),
+    )
```
texuf, line 183: this only creates gdms with local clients... but, that's fine for this iteration i guess.

PR #1240, opened 2024-10-10
stream-metadata: granular space refresh

miguel-nascimento:
Should close: TOWNS-12688

If you update the space name, the space member nft has a space name field, so we need to refresh both space owner nft metadata (/space/spaceAddress) and member nft (/space/spaceAddress/token/tokenId)

If you update the space image, we only need to update the /space/spaceAddress/image
The image field on the metadata for space & space member will be redirected to the right one


packages/stream-metadata/src/routes/spaceRefresh.ts
```diff
@@ -41,9 +45,17 @@ export async function spaceRefreshOnResponse(
 
 	logger.info({ spaceAddress }, 'Refreshing space')
 
+	const { target } = querySchema.parse(request.query)
+	const paths =
+		target === 'metadata'
+			? [`/space/${spaceAddress}`, `/space/${spaceAddress}/token/*`]
+			: target === 'image'
+			? [`/space/${spaceAddress}/image`]
+			: [`/space/${spaceAddress}/*`]
```
texuf, line 54: for "all" will `/space/${spaceAddress}/*` match `/space/${spaceAddress}`? I am superstitious about trailing slashes.

miguel-nascimento: https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/invalidation-specifying-objects.html#:~:text=If%20you%20use%20the%20AWS,a%20path%20is%204%2C000%20characters.

to include the `/space/spaceAddress` we could use `/space/spaceAdress*`

miguel-nascimento: `/*` only match things inside the slash

texuf (approved):

PR #1258, opened 2024-10-15
Add missing 404 cache control headers

texuf:


brianathere (approved):

texuf:
> #1213

I'll wait to put this in. If the service is fine then maybe we don't actually need these, but it seems like we should be doing something here. maybe a shorter timeout.

mechanical-turk:
> > #1213
> 
> I'll wait to put this in. If the service is fine then maybe we don't actually need these, but it seems like we should be doing something here. maybe a shorter timeout.

@texuf the client-side cache max-age for these are set to 5 seconds. that's a good timeout in my opinion. i think we can merge this

PR #1269, opened 2024-10-16
Prototype of checking signatures in a worker

brianathere:
So as to not block the event loop thread, which in turn eventually blocks the health check coming in on another request


packages/stream-metadata/src/riverStreamRpcClient.ts
```diff
@@ -149,6 +155,38 @@ function stripHexPrefix(hexString: string): string {
 	return hexString
 }
 
+const workerPath = path.join(__dirname, 'unpackStreamWorker.cjs')
```
mechanical-turk, line 158: @brianathere we should avoid source references to build if we can

brianathere: This is how nodejs workers are loaded. If you have another way to make the source available I'm interested in learning.

mechanical-turk: from [nodejs docs](https://nodejs.org/api/worker_threads.html). They put the main thread code and the worker code in the same file:

```js
const {
  Worker, isMainThread, parentPort, workerData,
} = require('node:worker_threads');

if (isMainThread) {
  module.exports = function parseJSAsync(script) {
    return new Promise((resolve, reject) => {
      const worker = new Worker(__filename, {
        workerData: script,
      });
      worker.on('message', resolve);
      worker.on('error', reject);
      worker.on('exit', (code) => {
        if (code !== 0)
          reject(new Error(`Worker stopped with exit code ${code}`));
      });
    });
  };
} else {
  const { parse } = require('some-js-parsing-library');
  const script = workerData;
  parentPort.postMessage(parse(script));
}
```

We could do something along these lines:

```ts
import {
  Worker, isMainThread, parentPort, workerData,
} from 'node:worker_threads';

export function parseJSAsync(script) {
    return new Promise((resolve, reject) => {
      const worker = new Worker(__filename, {
        workerData: script,
      });
      worker.on('message', resolve);
      worker.on('error', reject);
      worker.on('exit', (code) => {
        if (code !== 0)
          reject(new Error(`Worker stopped with exit code ${code}`));
      });
    });
  };

if (!isMainThread) {
  const { parse } = require('some-js-parsing-library');
  const script = workerData;
  parentPort.postMessage(parse(script));
}
```

mechanical-turk: something else i saw while reading the docs:

> The above example spawns a Worker thread for each parseJSAsync() call. In practice, use a pool of Workers for these kinds of tasks. Otherwise, the overhead of creating Workers would likely exceed their benefit.

In our case there will always be a benefit since we avoid blocking the http server, with or without pools. But pooling does sound like something we could benefit from

mechanical-turk (approved):

texuf:
Want to throw another potential solution out there. Python has gunicorn which allows you to run multiple processes of the same app, brian do you have any experience with the node cluster api? https://medium.com/deno-the-complete-reference/the-benefits-of-clustering-fastify-app-in-node-js-hello-world-case-8a99127b9951

brianathere:
> Want to throw another potential solution out there. Python has gunicorn which allows you to run multiple processes of the same app, brian do you have any experience with the node cluster api? https://medium.com/deno-the-complete-reference/the-benefits-of-clustering-fastify-app-in-node-js-hello-world-case-8a99127b9951

I don't. I just read that article, and unless I'm misreading the numbers, it doesn't sound very promising. It would still also put the /health endpoint at the mercy of the sync unpack operation, in the case where all cluster workers are unpacking, health would still block.

I'll take a look at it though, always interesting to learn new things.

PR #1271, opened 2024-10-16
Add miniblock num to event miniblock ref, and MINIBLOCK_TOO_NEW error code

sergekh2:
MINIBLOCK_TOO_NEW is returned instead of BAD_PREV_MINIBLOCK_HASH if local replica of the stream hasn't reached miniblock with the given number yet.


texuf (commented):
Are you going to update the typescript?

protocol/protocol.proto
```diff
@@ -1005,4 +1008,7 @@ enum Err {
     MINIPOOL_MISSING_EVENTS = 60;
     STREAM_LAST_BLOCK_MISMATCH = 61;
     DOWNSTREAM_NETWORK_ERROR = 62;
+
+    // Stream replica on the local node does not have miniblock with given number yet.
```
bas-vk, line 1012: We can mention that this is expected to be a temporary error and that the client can retry it.

PR #1273, opened 2024-10-16
Notification service

bas-vk:


protocol/protocol.proto
```diff
@@ -22,556 +22,556 @@ message Miniblock {
...
-            * on ack or new session, if any device’s lower bound < N generations ago, delete the deviceKey entry from the record
-            */
-            int64 lower_bound = 1;
-            int64 upper_bound = 2;
-        }
-
-        Inception inception = 1;
-        // deviceKey: miniblockNum that the ack was snapshotted
-        map<string, DeviceSummary> device_summary = 2;
+  message Snapshot {
```
texuf, line 402: can you make this protocol.proto formatting change in a different commit? Is this hooked up to a linter somewhere? Pretty sure vscode is just going to undo this.

protocol/notifications.proto
```diff
@@ -0,0 +1,320 @@
+syntax = "proto3";
+package river;
+option go_package = "github.com/river-build/river/core/node/protocol";
+
+import "google/protobuf/timestamp.proto";
+import "google/protobuf/empty.proto";
+
+service AuthenticationService {
```
texuf, line 8: the AuthenticationService service isn't namespaced. Is that intentional? Will it serve as auth for every service going forward or is it just for the NotificationService?

protocol/notifications.proto
```diff
@@ -0,0 +1,320 @@
...
+  // challenge random nonce that was used to sign
+  bytes challenge = 2;
+  // signature calculated with:
+  // ecdsa_sign(client_key, sha256(PREFIX || user_id || expiration || challenge)
+  // with
+  // PREFIX=NS_AUTH: (string with length 8)
+  // user_id raw user id bytes (160 bits)
+  // expiration is a 64 bits unix timestamp encoded in big endian
+  // || means concatenation
+  bytes signature = 3;
```
texuf, line 43: Can we add an optional delegate_sig here like we do in the StreamEvent? It's not vitally important, we can sign things with privy without throwing a modal, but it would be nice to have parity.

protocol/notifications.proto
```diff
@@ -0,0 +1,320 @@
...
+  bytes user_id = 1;
+  // challenge that must be signed.
+  // clients must verify before signing that the challenge is >= 128 bits
+  bytes challenge = 2;
+  // indication until when challenge is valid
+  google.protobuf.Timestamp expiration = 3;
+}
+
+// note, challenge expiration is stored in the notification service and therefore doesn't have to be provided
+// in the request. Expiration is looked up through the nonce.
```
texuf, line 30: you're using both the word nonce and the word challenge to mean the same thing. Can we call it challenge everywhere?

texuf (commented):
are you planning on adding just commands? I can start it with `just RUN_ENV=multi run-notifications` but it's not obvious to me how to config it.
It would be nice if i could run locally. It would be amazing if I could register a callback and start up a tiny server in a jest test to see it working end to end... :)

protocol/notifications.proto
bas-vk: It is just for the nameservice although it's generic by nature. The session token is a JWT with the account id as the subject and the issuer and audience set to `ns`. But that can be changed if there is another service that requires authentication. Including switching over to a asymmetric signing schema that allows any service to validate the JWT.

protocol/notifications.proto
bas-vk: We likely can but I have to look into the delegate signing part and figure out how it exactly works.

texuf:
> are you planning on adding just commands? I can start it with `just RUN_ENV=multi run-notifications` but it's not obvious to me how to config it. It would be nice if i could run locally. It would be amazing if I could register a callback and start up a tiny server in a jest test to see it working end to end... :)

This was incomplete, i edited the justfile to add 
```
run-notifications:
    just run-go notifications "@"
```
what's the proper way to config it?

protocol/notifications.proto
```diff
@@ -0,0 +1,334 @@
...
+  // and the UI should not render any feedback that a message was received.
+  GDM_NO_MESSAGES_AND_MUTE = 2;
+  // GDM_ONLY_MENTIONS_REPLIES_REACTIONS indicates that the user receives notifications for messages added to
+  // GDM channels that are either a direct reply or a reaction to his own messages.
+  GDM_ONLY_MENTIONS_REPLIES_REACTIONS = 3;
+  // GDM_MESSAGES_ALL indicates that the user receives notifications for all messages in GDM channels.
+  // This is the default.
+  GDM_MESSAGES_ALL = 4;
+}
+
```
texuf, line 129: Can we fix these enums to be consistent? We've got `GDM_NO_MESSAGES` with no first, and `DM_MESSAGES_NO` with no last...

I wasn't going to bring this up, but do we really need a new enum for each type? I think the ui can just not show the mentions option in dms... seems like `unspecified, all, mentions_replies_reactions, none, muted` would work... if not it's fine.

bas-vk:
> > are you planning on adding just commands? I can start it with `just RUN_ENV=multi run-notifications` but it's not obvious to me how to config it. It would be nice if i could run locally. It would be amazing if I could register a callback and start up a tiny server in a jest test to see it working end to end... :)
> 
> This was incomplete, i edited the justfile to add
> 
> ```
> run-notifications:
>     just run-go notifications "@"
> ```
> 
> what's the proper way to config it?

I just updated the `README.md` in the notifications package that describes the required configuration. If that isn't complete/clear let me know.

protocol/notifications.proto
bas-vk: > Can we fix these enums to be consistent? We've got `GDM_NO_MESSAGES` with no first, and `DM_MESSAGES_NO` with no last...
> 
> I wasn't going to bring this up, but do we really need a new enum for each type? I think the ui can just not show the mentions option in dms... seems like `unspecified, all, mentions_replies_reactions, none, muted` would work... if not it's fine.

The definitions are now renamed for consistency. Personally I prefer to have separate enums for them.

texuf: I trust you, but can you tell me more about your thought process so that I can be on the same page. Also, heads up - we will probably add a new notification type for enabling disabling notifications about tipping in the very near future.

bas-vk: My reasoning is that having a single enum for a single purpose is easier to understand instead of sharing them and filtering out the ones that aren't supported in the context.

texuf (approved):

protocol/notifications.proto
texuf: I don't think you need to filter on the server. I think you can treat every stream the same. On the client we have made the UX decision that replies and mentions don't make sense in DMs. but that is purely a UX decision and frankly, might change one day in the future.

core/node/events/events.go
```diff
@@ -70,6 +70,33 @@ func MakeStreamEvent(
...
+		return nil, AsRiverError(err, Err_INTERNAL).
+			Message("Failed to create random salt").
+			Func("MakeStreamEvent")
+	}
+	epochMillis := time.Now().UnixNano() / int64(time.Millisecond)
+
+	event := &StreamEvent{
+		CreatorAddress:    wallet.Address.Bytes(),
+		Salt:              salt,
+		PrevMiniblockHash: prevMiniblockHash,
```
sergekh2, line 91:
```suggestion
		PrevMiniblockHash: prevMiniblock.Hash[:],
		PrevMiniblockNum:  prevMiniblock.Num,
```
to match prev comment

core/node/events/events.go
```diff
@@ -70,6 +70,33 @@ func MakeStreamEvent(
 	return event, nil
 }
 
+func MakeStreamEventWithTags(
+	wallet *crypto.Wallet,
+	payload IsStreamEvent_Payload,
+	prevMiniblockHash []byte,
```
sergekh2, line 76:
```suggestion
	prevMiniblock *MiniblockRef,
```
Event has PrevMiniblockNum now - it helps to detect cases when local node is behind to hand it correctly during event add. (see comment below)

core/node/events/miniblock.go
```diff
@@ -141,11 +140,13 @@ func (b *MiniblockInfo) asStorageMbWithData(bytes []byte) *storage.WriteMinibloc
 func (b *MiniblockInfo) forEachEvent(op func(e *ParsedEvent, minibockNum int64, eventNum int64) (bool, error)) error {
 	blockNum := b.header().MiniblockNum
 	eventNum := b.header().EventNumOffset
-	for _, event := range b.events() {
-		c, err := op(event, blockNum, eventNum)
-		eventNum++
-		if !c {
-			return err
+	if b.Proto != nil {
```
sergekh2, line 143: this is not correct:

if MiniblockInfo was parsed with DontParseEvents, forEachEvent should fail (`events()` panics in this case, but it's possible to return Err_INTERNAL if this is more desirable).

in other words, it shouldn't silently pretend that events were processed correctly.

core/node/events/parsed_event.go
```diff
@@ -16,6 +16,7 @@ import (
 type ParsedEvent struct {
 	Event         *StreamEvent
 	Envelope      *Envelope
+	Tags          *Tags
```
sergekh2, line 19: This seems a bit redundant: this field is readily available as `Event.Tags`, if brevity is desired, it's possible to add `Tags()` method to `ParsedEvent`

sergekh2 (approved):

PR #1280, opened 2024-10-17
Stress tests with redis-encoded test plans

clemire:
Thus far:

- I created a test plan schema with a short list of commands that can be used to specify the order of operations in testing
  - each command can optionally specify clients for which it should execute, and/or the command itself can branch it's logic based on the client id
  - success is currently execution without error, I have not thought about more nuanced ways to report success yet

 - the `mintMemberships` command has an example of orchestrating around redis - all clients wait until the root client finishes minting.
 - there is an execution loop that looks for failure from another node and terminates early. To avoid this logic you can catch errors in your command definition and discard them.

test case that runs locally:
- mint memberships for all clients
- all clients join space and announcement channel
- root client joins stress0 and sends a message
- all clients wait for above message to render


packages/stress/scripts/localhost_schemachat.sh
```diff
@@ -0,0 +1,54 @@
+#!/bin/bash
+set -euo pipefail
+cd -P -- "$(dirname -- "${BASH_SOURCE[0]}")"
+cd ..
+
+# run scripts/localhost_chat_setup.sh to set up the environment variables
+
+# List of environment files to source
+ENV_FILES=(
+    "./scripts/.env.storage"
```
clemire, line 10: These tests don't use environment variables besides storage. All test-specific config (space id, channel ids) I plan to move into the test schema, or make available via orchestration.

packages/stress/src/mode/chat/schema/commands/ExecutionError.ts
clemire, line 1: For clarity when throwing nested errors for ease of debugging from console

packages/stress/src/mode/chat/schema/commands/expectChannelMessageCommand.ts
```diff
@@ -0,0 +1,121 @@
...
+
+export const expectChannelMessageCommand = baseCommand.extend({
+    name: z.literal('expectChannelMessage'),
+    params: paramsSchema,
+})
+
+export type ExpectChannelMessageParams = z.infer<typeof paramsSchema>
+export type ExpectChannelMessageCommand = z.infer<typeof expectChannelMessageCommand>
+
+function escapeMessage(template: string, client: StressClient, cfg: ChatConfig): string {
```
clemire, line 29: send and expect commands allow for escapes to make observing stress tests and detecting messages easier.

packages/stress/src/mode/chat/schema/executeCommand.ts
```diff
@@ -0,0 +1,80 @@
...
+            break
+        default: {
+            logger.error({ command }, 'unrecognized command type')
+        }
+    }
+
+    check(!!execute, 'Unimplemented command type')
+    await Promise.all(
+        targetClients.map(async (client) => {
+            // Should not be an issue due to the check line above, but sometimes
```
clemire, line 74:
```suggestion
            // A null `execute` should not be an issue due to the check line above, but sometimes
```

packages/stress/src/mode/chat/schema/commands/joinSpaceCommand.ts
```diff
@@ -0,0 +1,62 @@
...
+export const joinSpaceCommand = baseCommand.extend({
+    name: z.literal('joinSpace'),
+    params: joinSpaceParamsSchema,
+})
+
+export type JoinSpaceParams = z.infer<typeof joinSpaceParamsSchema>
+export type JoinSpaceCommand = z.infer<typeof joinSpaceCommand>
+
+// joins space and announcement channel
+// idempotent, fine to run if user is already a space member
```
clemire, line 21:
```suggestion
// this command is idempotent, fine to run if user is already a space member
```

packages/stress/src/mode/chat/schema/executeCommand.ts
```diff
@@ -0,0 +1,80 @@
...
+    command: Command,
+    commandId: string, // For logging / error reporting
+    chatConfig: ChatConfig,
+    clients: StressClient[],
+) {
+    const logger = getLogger('stress:run', { function: 'executeCommand' })
+    logger.info({ command }, '--------------- executing command ---------------')
+
+    const range = (start: number, end: number) =>
+        Array.from({ length: end - start }, (v, k) => k + start)
```
sergekh2, line 21: endIndex inclusive? i.e. it should be end-start + 1 ?

packages/stress/src/mode/chat/schema/executeCommand.ts
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    logger.debug({ commandId, targetClientIndices }, 'Executing command for clients', 'command')
+    type execFn = (client: StressClient, cfg: ChatConfig) => Promise<void>
+    let execute: execFn | undefined = undefined
+
+    switch (command.name) {
+        case 'mintMemberships':
+            {
+                execute = async (client: StressClient, cfg: ChatConfig) => {
```
sergekh2, line 42: if execFn is updated to have params as 3rd argument, then it's possible to assign it directly. And then it's a bit easier to map even without switch like this:

```typescript
const commandFns : { [key: string]: execFn } = {
  mintMemberships: mintMemberships,
  joinSpace: joinSpace,
  fooBar: fooBar,
};
const execute = commandFns[command.name]
```

packages/stress/src/mode/chat/schema/commands/expectChannelMessageCommand.ts
```diff
@@ -0,0 +1,121 @@
...
+    client: StressClient,
+    cfg: ChatConfig,
+    channel: Stream,
+    text: string,
+    logger: Logger,
+) {
+    let count = 0
+    const message = await client.waitFor(
+        () => {
+            if (count % 3 === 0) {
```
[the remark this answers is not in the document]
clemire: yes, this was copy-pasta, I kept for that reason. chatty logs make debugging more difficult.

packages/stress/src/mode/chat/schema/executeCommand.ts
clemire: Non-inclusive! clientEndIndex is not included in the indices of clients running on this process.

packages/stress/src/mode/chat/schema/executeCommand.ts
clemire: The param shape actually varies by command type - does that work for your suggestion?

clemire: @texuf also ^ - we're running up against the limitations of my typescript type foo

PR #1309, opened 2024-10-21
Add miniblock headers to view when initializing streams

texuf:
1) in cases where event signature validation is turned off, i want to be able to validate events later, so i need to keep the miniblock header event around so that i can check the signature
2) we have bookkeeping on miniblock numbers that we need to keep up to date
3) i think this is causing issues with membership confirmation


texuf:
ah nevermind, the header is in there already

texuf:
i think this is the second time i fell for this

PR #1312, opened 2024-10-21
Key solicitation correctness fix - key local and dispatched solicitations in sync

texuf:
I don’t know if this caused any actual issues, but I’m going to be modifying this structure and it seems weird that we didn’t preserve the sort here.

### DIFF
--- a/packages/sdk/src/streamStateView_Members_Solicitations.ts
+++ b/packages/sdk/src/streamStateView_Members_Solicitations.ts
@@ -32,19 +32,19 @@ export class StreamStateView_Members_Solicitations {
         user.solicitations = user.solicitations.filter(
             (x) => x.deviceKey !== solicitation.deviceKey,
         )
-        user.solicitations.push({
+        const newSolicitation = {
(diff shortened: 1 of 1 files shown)


packages/sdk/src/streamStateView_Members_Solicitations.ts
```diff
@@ -32,19 +32,19 @@ export class StreamStateView_Members_Solicitations {
         user.solicitations = user.solicitations.filter(
             (x) => x.deviceKey !== solicitation.deviceKey,
         )
-        user.solicitations.push({
+        const newSolicitation = {
             deviceKey: solicitation.deviceKey,
             fallbackKey: solicitation.fallbackKey,
             isNewDevice: solicitation.isNewDevice,
             sessionIds: [...solicitation.sessionIds.sort()],
```
ttillberg, line 39: sort mutates the original array, [...solicitation.sessionIds].sort() ?

texuf: i guess why this wasn't a problem, i'll fix

Issue #1307, opened 2024-10-21
Make stream reconciliation worker pool configurable


sergekh2:
### Description

https://github.com/river-build/river/blob/bcb65fa243a743f98bd49d70499586c4a1185669/core/node/events/stream_sync_task.go#L24

1. replace ants with github.com/gammazero/workerpool
2. make number of worker configurable, set default to 8 (to mitigate DB load)

PR #1315, opened 2024-10-21
Stream reconciliation to use workerpool with configurable size, defaulting to 8

clemire:
Configuration logic was tested locally.

### DIFF
--- a/core/config/config.go
+++ b/core/config/config.go
@@ -37,6 +37,9 @@ func GetDefaultConfig() *Config {
 		},
 		// TODO: ArchitectContract: ContractConfig{},
 		// TODO: RegistryContract:  ContractConfig{},
+		StreamReconciliation: StreamReconciliationConfig{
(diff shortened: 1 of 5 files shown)


clemire:
#1307

core/config/config.go
```diff
@@ -401,6 +404,17 @@ type ScrubbingConfig struct {
 	ScrubEligibleDuration time.Duration
 }
 
+type StreamReconciliationConfig struct {
+	WorkerPoolSize int // If 0, default to 8.
+}
+
+func (s *StreamReconciliationConfig) GetWorkerPoolSize() int {
```
sergekh2, line 411: no need for this function:
set it to 8 in GetDefaultConfig();
if setting is set, it will overwritten.

sergekh2 (approved):

bas-vk (approved):

PR #1318, opened 2024-10-21
Add ClientInitStatus flag for isHighPriorityDataLoaded

texuf:
so that we can render data faster


texuf:
moved to https://github.com/river-build/river/pull/1319

PR #1325, opened 2024-10-22
Postgres mitigation: Migrate queries to target specific partitions.

clemire:


core/node/storage/pg_stream_store.go
```diff
@@ -602,6 +642,7 @@ func (s *PostgresStreamStore) readMiniblocksTx(
 ) ([][]byte, error) {
 	miniblocksRow, err := tx.Query(
 		ctx,
+		// Do not escape this query as the specific partition may not exist
```
sergekh2, line 645: For all "Do not escape" comments, does it make sense to handle "table not found" error in the same way as no results?

clemire: I updated the handling for each specific case, ptal at the comments where I did not change things.

sergekh2 (approved):

PR #1328, opened 2024-10-22
Optimize stream metadata service by skipping validation

texuf:
long term we should figure out a way to turn this back on

### DIFF
--- a/packages/stream-metadata/src/riverStreamRpcClient.ts
+++ b/packages/stream-metadata/src/riverStreamRpcClient.ts
@@ -17,6 +17,11 @@ import { FastifyBaseLogger } from 'fastify'
 import { MediaContent, StreamIdHex } from './types'
 import { getNodeForStream } from './streamRegistry'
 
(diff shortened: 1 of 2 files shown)


packages/stream-metadata/src/riverStreamRpcClient.ts
```diff
@@ -180,7 +185,10 @@ export async function getStream(
 			'getStream finished',
 		)
 
-		const unpackedResponse = await unpackStream(response.stream, opts)
+		const unpackedResponse = await unpackStream(
```
mechanical-turk, line 188: @texuf i think we should remove the optional opts both from the function and from the rest of the codebase if its used. lgtm otherwise

mechanical-turk (approved):

packages/stream-metadata/src/riverStreamRpcClient.ts
texuf: I did it this way so that when this code gets copypasta'd somewhere else it's obvious to the dev what's supposed to happen. 
Seeing that this code is itself a copypasta, it will probably happen. 
And I want to at least leave the door open to implementations where we verify everything. I will move the constant up to the function definition though, that will read better.

PR #1345, opened 2024-10-24
adding an lru cache for incoming getStream calls

mechanical-turk:


texuf (approved):

packages/stream-metadata/src/routes/spaceImage.ts
```diff
@@ -47,7 +47,9 @@ export async function fetchSpaceImage(request: FastifyRequest, reply: FastifyRep
 	let stream: StreamStateView
 	try {
 		const streamId = makeStreamId(StreamPrefix.Space, spaceAddress)
-		stream = await getStream(logger, streamId)
+		stream = await getStream(logger, streamId, {
+			skipCache: !!eventId, // when an event id is provided, we skip the cache
+		})
```
texuf, line 52: sorry I was worried about cache busting on the `fetchSpaceMetadata` and `fetchSpaceMemberMetadata`
@miguel-nascimento is polling it with a cacheBust=nanoid url param to get around the browser cache, i don't want him to then hit this cache too.

I think this would work better if it was part of the cache key

texuf:
I landed this https://github.com/river-build/river/pull/1519. Lets see if it makes a difference.

PR #1351, opened 2024-10-25
stream-metadata: add support for `/refreshStatus/:invalidationId`

miguel-nascimento:
We need to wait for refresh to finish before emitting `refreshMetadata` for space members nft.
Previously we tried polling both space and space member metadata and waiting eventId of the image to be equal
The previous attempt didn’t worked and also increased the load on the service. 

This PR solves it, adding a `/refreshStatus/:invalidationId` route. 
Now after a refresh we get the invalidationId, which we can use to poll `/refreshStatus/:invalidationId` on the client until the refresh is completed.

We dont need `/event` in the space image route anymore. I'll create a PR later

### DIFF
--- a/packages/stream-metadata/src/environment.ts
+++ b/packages/stream-metadata/src/environment.ts
@@ -79,9 +79,6 @@ function makeConfig() {
 		healthCheck: {
 			timeout: 5000, // 5 seconds
 		},
-		headers: {
-			invalidationId: 'X-StreamMetadata-Invalidation-Id',
-		},
 	}
 }
 
--- a/packages/stream-metadata/src/environment.ts
+++ b/packages/stream-metadata/src/environment.ts
@@ -79,6 +79,9 @@ function makeConfig() {
 		healthCheck: {
 			timeout: 5000, // 5 seconds
 		},
+		headers: {
+			invalidationId: 'X-StreamMetadata-Invalidation-Id',
+		},
 	}
 }
 
--- a/packages/stream-metadata/src/node.ts
+++ b/packages/stream-metadata/src/node.ts
@@ -16,9 +16,10 @@ import { fetchUserProfileImage } from './routes/profileImage'
 import { fetchUserBio } from './routes/userBio'
 import { fetchMedia } from './routes/media'
 import { spaceRefresh, spaceRefreshOnResponse } from './routes/spaceRefresh'
-import { userRefresh, userRefreshOnResponse } from './routes/userRefresh'
+import { userRefresh } from './routes/userRefresh'
 import { addCacheControlCheck } from './check-cache-control'
 import { fetchSpaceMemberMetadata } from './routes/spaceMemberMetadata'
+import { fetchRefreshStatus } from './routes/refreshStatus'
 
 // Set the process title to 'stream-metadata' so it can be easily identified
 // or killed with `pkill stream-metadata`
@@ -116,10 +117,11 @@ export function setupRoutes(srv: Server) {
 
 	// not cached
 	srv.get('/health', checkHealth)
+	srv.get('/refreshStatus/:invalidationId', fetchRefreshStatus)
 
 	// should be rate-limited, but not yet
 	srv.get('/space/:spaceAddress/refresh', { onResponse: spaceRefreshOnResponse }, spaceRefresh)
-	srv.get('/user/:userId/refresh', { onResponse: userRefreshOnResponse }, userRefresh)
+	srv.get('/user/:userId/refresh', userRefresh)
 
 	// Fastify will return 404 for any unmatched routes
 }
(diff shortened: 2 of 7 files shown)


packages/stream-metadata/src/node.ts
```diff
@@ -116,10 +117,11 @@ export function setupRoutes(srv: Server) {
 
 	// not cached
 	srv.get('/health', checkHealth)
+	srv.get('/refreshStatus/:invalidationId', fetchRefreshStatus)
 
 	// should be rate-limited, but not yet
 	srv.get('/space/:spaceAddress/refresh', { onResponse: spaceRefreshOnResponse }, spaceRefresh)
-	srv.get('/user/:userId/refresh', { onResponse: userRefreshOnResponse }, userRefresh)
```
texuf, line 122: why did this change ^

texuf: Oh, i see. Nice.

texuf (approved):
Please wait for kerem to review, but this is awesome thanks.

packages/stream-metadata/src/node.ts
miguel-nascimento: We have the onResponse hook on the space refresh because we want to wait it to finish before invalidating opensea cache.
We don’t need to do this in the user refresh endpoint.

miguel-nascimento: Also, now that we want to return the invalidationId after calling user refresh, I had to move the invalidation from the hook to the endpoint itself. The hook was empty after it

packages/stream-metadata/src/environment.ts
```diff
@@ -79,6 +79,9 @@ function makeConfig() {
 		healthCheck: {
 			timeout: 5000, // 5 seconds
 		},
+		headers: {
+			invalidationId: 'X-StreamMetadata-Invalidation-Id',
+		},
```
mechanical-turk, line 84: nit: @miguel-nascimento i think we should move this out to elsewhere because it's more of a constant and less of a config. it will never change. you can create a utils/consts.ts for this maybe?

packages/stream-metadata/src/routes/spaceRefresh.ts
```diff
@@ -14,7 +15,7 @@ const paramsSchema = z.object({
 		}),
 })
 
-// This route handler validates the refresh request and quickly returns a 200 response.
+// This route handler validates the refresh request, initiates the CloudFront invalidation, and returns a 200 response.
```
mechanical-turk, line 18: @miguel-nascimento i like what you're doing here. good job!

question: does the client need to be aware of the invalidation id? if not, i suppose you're using this to communicate the id between the route and the onResponse hook. this is not a big deal but using the context pattern might be better here

packages/stream-metadata/src/routes/userRefresh.ts
```diff
@@ -57,7 +46,12 @@ export async function userRefreshOnResponse(
 			: [`/user/${userId}/*`]
 
 	try {
-		await CloudfrontManager.createCloudfrontInvalidation({ paths, logger })
+		const invalidation = await CloudfrontManager.createCloudfrontInvalidation({ paths, logger })
+		const invalidationId = invalidation?.Invalidation?.Id
+		return reply
+			.code(200)
+			.header(config.headers.invalidationId, invalidationId)
```
mechanical-turk, line 53: @miguel-nascimento same question here: is there a reason it's on the response header and the body as well?

mechanical-turk: @miguel-nascimento and a final comment: cf invalidations have rate limits and i'm seeing a bunch of error logs about exceeding our limit. this is why i was putting them in the onResponse hooks instead of the actual response. 

while i'm talking about error logs, @texuf the reason this error is not front and center right now is because we have a lot of error log fatigue in this service. things that are not errors (i.e stream not found) are being logged as if they're fatal errors, making it harder to understand the real errors. can we please do an error hygiene round? it's out of control

packages/stream-metadata/src/routes/spaceRefresh.ts
miguel-nascimento: The client needs the `invalidationId` so it can get the `refreshStatus` later with the same id.

I'm passing in the headers just to communicate between route <> onResponse, it was the easiest way (I think the correct way here would be `AsyncLocalStorage`)

packages/stream-metadata/src/routes/userRefresh.ts
miguel-nascimento: If we need to move it to onResponse, we could use PubSub/SNS here, so we can return a quick 200 response and later the client can get the `invalidationId`. Unsure if it's the best approach to the cf invalidation rate limit.

miguel-nascimento: > is there a reason it's on the response header and the body as well?

hack to communicate between response <> onResponse hook 🙈

PR #1353, opened 2024-10-25
dApp: Enable empty metadata when updating channel status only

ttillberg:
When disabling channels, passing names and roles is not required. To ensure the contract accepts this exception, the metadata param should be encoded to 0 bytes.


shuhuiluo (approved):
Lgtm.

PR #1357, opened 2024-10-25
sdk(sync-agent): add dm models

miguel-nascimento:
This PR adds support for DM channels in sdk and in the react-sdk.

Also adds DM support in the playground, for users that already have some DMs.


packages/sdk/src/sync-agent/dms/dms.ts
```diff
@@ -0,0 +1,73 @@
...
+        if (!this.dms[streamId]) {
+            this.dms[streamId] = new Dm(streamId, this.riverConnection, this.store)
+        }
+        return this.dms[streamId]
+    }
+
+    byUserId(userId: string): Dm {
+        const streamId = makeDMStreamId(this.riverConnection.userId, userId)
+        return this.byStreamId(streamId)
+    }
```
texuf, line 52: I think these two functions should be `getDm(streamId)` and `getDmWithUserId(userId)`

texuf (approved):

PR #1377, opened 2024-10-30
stream-metadata: explicit log when space metadata get called

miguel-nascimento:
lets make it easy to inspect

### DIFF
--- a/packages/stream-metadata/src/routes/spaceMetadata.ts
+++ b/packages/stream-metadata/src/routes/spaceMetadata.ts
@@ -11,7 +11,12 @@ import { getStream } from '../riverStreamRpcClient'
 export const spaceMetadataBaseUrl = `${config.streamMetadataBaseUrl}/space`.toLowerCase()
 
 const paramsSchema = z.object({
-	spaceAddress: z.string().min(1, 'spaceAddress parameter is required'),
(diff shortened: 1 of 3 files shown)


packages/stream-metadata/src/routes/spaceMetadata.ts
```diff
@@ -40,15 +45,7 @@ export async function fetchSpaceMetadata(request: FastifyRequest, reply: Fastify
...
-
-	// Validate spaceAddress format using the helper function
-	if (!isValidEthereumAddress(spaceAddress)) {
-		logger.info({ spaceAddress }, 'Invalid spaceAddress format')
-		return reply
-			.code(400)
-			.header('Cache-Control', CACHE_CONTROL['4xx'])
-			.send({ error: 'Bad Request', message: 'Invalid spaceAddress format' })
-	}
+	logger.info({ spaceAddress }, 'Fetching space metadata')
```
texuf, line 48: does this result in us logging every request twice?

miguel-nascimento: Technically yes. We log all incoming requests in a middleware, but since we also have this logging pattern in other endpoints, I would find it useful to have the same here

PR #1387, opened 2024-10-31
react-sdk: get member metadata by userId and streamId

miguel-nascimento:
Make it easier to get member metadata / update our own metadata


packages/react-sdk/src/useMyself.ts
```diff
@@ -0,0 +1,51 @@
+import type { Member, Myself, SyncAgent } from '@river-build/sdk'
+import { useMemo } from 'react'
+import { type ActionConfig, useAction } from './internals/useAction'
+import { type ObservableConfig, useObservable } from './useObservable'
+import { useSyncAgent } from './useSyncAgent'
+import { getRoom } from './utils'
+
+const getMyself = (sync: SyncAgent, streamId: string) => getRoom(sync, streamId).members.myself
```
texuf, line 8: can we call this useMyMember instead of useMyself? useMySelf sounds like it would return everything about myself...

texuf (approved):
approved with changes requested

PR #1393, opened 2024-10-31
Add docker compose pull to launch_storage.sh script

texuf:
to get latest images? is this recommended? I was stuck on postgres version 15

### DIFF
--- a/core/scripts/launch_storage.sh
+++ b/core/scripts/launch_storage.sh
@@ -5,4 +5,5 @@ echo
 echo "Launching Postgres..."
 echo 
 
+docker compose pull
 docker compose --project-name river up --detach --wait


sergekh2 (approved):

texuf:
Because our scripts don't delete the volumes, if we delete the container and restart then we could get a different error, which is that the volume version

```
2024-10-31 16:51:03.723 UTC [1] FATAL:  database files are incompatible with server
2024-10-31 16:51:03.723 UTC [1] DETAIL:  The data directory was initialized by PostgreSQL version 15, which is not compatible with this version 17.0 (Debian 17.0-1.pgdg120+1).
```

texuf:
closing since randomly breaking people seems worse than just leaving them alone.

PR #1406, opened 2024-11-02
Change stream pagination setting from 5000 to 1000 (provider started to return failure for this page size)

sergekh2:


core/node/registries/river_registry_contract.go
```diff
@@ -290,7 +290,7 @@ func (c *RiverRegistryContract) ForAllStreams(
 	cb func(*GetStreamResult) bool,
 ) error {
 	// TODO: setting
-	const pageSize = int64(5000)
+	const pageSize = int64(3000)
```
jterzis, line 293: I also tried a lower page size locally running a bash script of the same function and got mixed results for page size as low as 1000. Maybe we should reduce this by a factor of ten, so 500 to be on the safe side ?

jterzis (approved):

PR #1420, opened 2024-11-04
hot fix for token balance withdrawal resetting

giuseppecrj:


jterzis (approved):

contracts/src/spaces/facets/membership/MembershipBase.sol
```diff
@@ -109,7 +109,20 @@ abstract contract MembershipBase is IMembershipBase {
...
+    MembershipStorage.Layout storage ds = MembershipStorage.layout();
+    uint256 contractBalance = address(this).balance;
+
+    // Rebalance stored balance if it doesn't match actual balance
+    // only if the contract balance is less than the stored balance
+    if (contractBalance < ds.tokenBalance) {
+      return contractBalance;
+    }
+
+    return ds.tokenBalance;
```
shuhuiluo, line 121:
```suggestion
    return FixedPointMathLib.min(contractBalance, ds.tokenBalance);
```

shuhuiluo (approved):

contracts/src/spaces/facets/membership/MembershipBase.sol
```diff
@@ -109,7 +109,20 @@ abstract contract MembershipBase is IMembershipBase {
   }
 
   function _getCreatorBalance() internal view returns (uint256) {
-    return MembershipStorage.layout().tokenBalance;
+    MembershipStorage.Layout storage ds = MembershipStorage.layout();
+    uint256 contractBalance = address(this).balance;
+
+    // Rebalance stored balance if it doesn't match actual balance
+    // only if the contract balance is less than the stored balance
```
shuhuiluo, line 116: What if the contract has surplus ETH? Does it have a `fallback` function?

shuhuiluo (changes requested):

contracts/src/spaces/facets/membership/MembershipBase.sol
giuseppecrj: The only way it would have surplus ETH apart from just sending directly is if there's an entitlement check that's pending

giuseppecrj: no fallback for now

shuhuiluo (approved):

PR #1436, opened 2024-11-05
banning: change to Permission.ModifyBanning and fix entitlement check

salzbrenner:
Banning only worked for space owners
- contract uses Permission.ModifyBanning. Updates both TS and go code to match
- calling removeUser on space recursively calls removeUser for each channel, but the entitlement check for booting from channel was failing for a non-owner user w/ banning permission. Fixes so that in the case of ModifyBanning, the channel entitlement check will check space permissions

### DIFF
--- a/core/node/rules/can_add_event.go
+++ b/core/node/rules/can_add_event.go
@@ -970,6 +970,17 @@ func (ru *aeMembershipRules) channelMembershipEntitlements() (*auth.ChainAuthArg
 		return nil, err
 	}
 
+
+	// ModifyBanning is a space level permission
+	// but users with this entitlement should also be entitled to kick users from the channel
+	if permission == auth.PermissionModifyBanning {
+		return auth.NewChainAuthArgsForSpace(
+			spaceId,
+			permissionUser,
+			permission,
+		), nil
+	}
+
 	chainAuthArgs := auth.NewChainAuthArgsForChannel(
 		spaceId,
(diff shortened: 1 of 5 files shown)


ttillberg (approved):

core/node/rules/can_add_event.go
```diff
@@ -970,6 +970,17 @@ func (ru *aeMembershipRules) channelMembershipEntitlements() (*auth.ChainAuthArg
 		return nil, err
 	}
 
+
+	// ModifyBanning is a space level permission
+	// but users with this entitlement should also be entitled to kick users from the channel
```
clemire, line 975: I'm a little confused by this comment. Did you mean that space-level ModifyBanning also entitles a user to kick users from any channel due to them being banned?

clemire (approved):

core/node/rules/can_add_event.go
salzbrenner: Yes, that's what I meant.
Based on the use of banning in both tests and in app code, we expect users w/ ban permission to call `removeUser` and be able to remove users from space and channels. However, a role w/ a banning permission does not need to be applied to a channel, and this check was failing for the user entitled at space level.
Does that make sense? I think that's the intended behavior, but this is the first time I've worked with banning and so not too familiar with expectation.

clemire: That makes sense and all sounds good! Thank you for explaining!

PR #1441, opened 2024-11-06
Fix archiver race condition

clemire:
for the Continuous test case, streams are added because they are detected in the registry, and the call to ReadMiniblocks on the stream node may result in a stream not found since the stream node may not have yet allocated the stream node in storage.

This PR adds a retry in the archiver when a NOT_FOUND error is returned and extends test waits to accommodate the longer execution time.

### DIFF
--- a/core/node/rpc/archiver.go
+++ b/core/node/rpc/archiver.go
@@ -181,17 +181,42 @@ func (a *Archiver) ArchiveStream(ctx context.Context, stream *ArchiveStream) err
 	}
 
 	for mbsInDb < mbsInContract {
+
 		toBlock := min(mbsInDb+int64(a.config.GetReadMiniblocksSize()), mbsInContract)
-		resp, err := stub.GetMiniblocks(
-			ctx,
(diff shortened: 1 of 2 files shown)


core/node/rpc/archiver.go
```diff
@@ -181,17 +181,42 @@ func (a *Archiver) ArchiveStream(ctx context.Context, stream *ArchiveStream) err
...
+					retries,
+				)
+				stream.nodes.AdvanceStickyPeer(nodeAddr)
+
+				if AsRiverError(err).Code != Err_NOT_FOUND {
+					return err
+				}
+			}
+			// Wait half a second before retrying
+			time.Sleep(time.Millisecond * 500)
```
bas-vk, line 215: I'm not deeply familiar with the archive code. But is this long enough? And if after 3 attempts it still fails is it rescheduled at some time in the future?

clemire: It is actually rescheduled for 1 second later by this loop, I'm refactoring now.

sergekh2 (approved):

PR #1445, opened 2024-11-06
Add DecodedCheckOperationBuilder for validating constructed check ops

clemire:
Add a builder for DecodedCheckOperations that can be used in harmony for constructing rule datas. Eventually I'd like to move all DecodedCheckOperation code to harmony.

This is part of a planned set of PRs to consolidate logic for constructing rule datas and harden them with better library-provided validation.

### DIFF
--- a/packages/web3/src/entitlement.ts
+++ b/packages/web3/src/entitlement.ts
@@ -1,7 +1,7 @@
-import type { ExtractAbiFunction } from 'abitype'
+import { parseAbiParameters, type ExtractAbiFunction } from 'abitype'
 import { IRuleEntitlementBase, IRuleEntitlementAbi } from './v3/IRuleEntitlementShim'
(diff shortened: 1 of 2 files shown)


packages/web3/src/entitlement.ts
```diff
@@ -767,12 +768,133 @@ export function createExternalNFTStruct(
...
+                throw new Error(`Check operation type ${opStr} unrecognized or not used in production`)
+        }
+    }
+}
+
 export type DecodedCheckOperation = {
     type: CheckOperationType
-    chainId: bigint
-    address: Address
+    chainId?: bigint
```
clemire, line 893: chainId and address are actually not required for ETH_BALANCE checks, lets make these optional and ensure they exist when required by constructing DecodedCheckOperations with the builder in harmony.

PR #1448, opened 2024-11-06
Add proof generation route, tests

jterzis:
- add /api/merkle-proof route to generate a merkle proof from conditionId, merkleRoot, claim leaf data.
- update types for enhanced type coverage
- update wrangler.toml for deployment to gamma, omega
- update tests

### DIFF
--- a/packages/workers/merkle-worker/src/router.ts
+++ b/packages/workers/merkle-worker/src/router.ts
@@ -1,18 +1,23 @@
 import { Router } from 'itty-router'
-import { createErrorResponse, ErrorCode } from './createResponse'
+import { createErrorResponse, createSuccessResponse, ErrorCode } from './createResponse'
 import { Env } from '.'
 import { WorkerRequest } from './utils'
 import { StandardMerkleTree } from '@openzeppelin/merkle-tree'
-import { Claim } from './types'
+import { Claim, MerkleData, MerkleTreeDump, MerkleProofResponse } from './types'
 
 const router = Router()
 
 router.post('/admin/api/merkle-root', async (request: WorkerRequest, env: Env) => {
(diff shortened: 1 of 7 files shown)


packages/workers/merkle-worker/src/router.ts
```diff
@@ -1,35 +1,120 @@
...
+        }
+
+        const response: MerkleProofResponse = {
+            proof,
+            leaf: [claim.address, claim.amount],
+        }
+
+        return createSuccessResponse(200, 'Proof generated successfully', response)
+    } catch (error) {
+        return createErrorResponse(500, 'Error processing request', ErrorCode.INTERNAL_SERVER_ERROR)
```
texuf, line 114: can you log the error here? Doesn't look like we're doing anything with it atm.

packages/workers/merkle-worker/src/router.ts
```diff
@@ -1,35 +1,120 @@
...
+import { createErrorResponse, createSuccessResponse, ErrorCode } from './createResponse'
 import { Env } from '.'
 import { WorkerRequest } from './utils'
 import { StandardMerkleTree } from '@openzeppelin/merkle-tree'
-import { Claim } from './types'
+import { Claim, MerkleData, MerkleTreeDump, MerkleProofResponse } from './types'
 
 const router = Router()
 
 router.post('/admin/api/merkle-root', async (request: WorkerRequest, env: Env) => {
```
texuf, line 10: how are we authenticating admins on this admin route?

jterzis: With a separate encrypted bearer token that the public facing client app does not see.

PR #1455, opened 2024-11-07
move to vitest

miguel-nascimento:
since i was struggling with jest and all the transform steps to bundle esm -> cjs, lets try move to vitest and use esm everywhere.


miguel-nascimento:
close - we want to move to vitest before messing with esm exports & bundling

PR #1458, opened 2024-11-07
Refactor storage stats to have it's own debug endpoint and add stream-related metadata

clemire:
The PostgresStatusResult for the local node was incorrectly being reported for all nodes in the network on the debug/multi and debug/multi/json endpoints. This PR does the following:
- remove local postgres status result from external nodes on the json page
- remove the pg status column from the debug/multi table, since we can't compute it for external nodes
- add a separate storage debug endpoint for inspecting these metrics

We also add the following stream-related metadata to our postgres metrics:
- MigratedStreams
- UnmigratedStreams
- NumPartitions

### DIFF
--- a/core/node/rpc/debug.go
+++ b/core/node/rpc/debug.go
@@ -63,6 +63,7 @@ func (s *Service) registerDebugHandlers(enableDebugEndpoints bool, cfg config.De
 	mux := s.mux
 	handler := debugHandler{}
 	mux.HandleFunc("/debug", handler.ServeHTTP)
+	handler.HandleFunc(mux, "/debug/storage", s.handleDebugStorage)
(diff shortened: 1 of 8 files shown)


core/node/rpc/debug.go
```diff
@@ -63,6 +63,7 @@ func (s *Service) registerDebugHandlers(enableDebugEndpoints bool, cfg config.De
 	mux := s.mux
 	handler := debugHandler{}
 	mux.HandleFunc("/debug", handler.ServeHTTP)
+	handler.HandleFunc(mux, "/debug/storage", s.handleDebugStorage)
```
sergekh2, line 66: Also please add an option to enable/disable it here (and set it to true in GetDefaultConfig())

https://github.com/river-build/river/blob/main/core/config/config.go#L377

sergekh2 (approved):

PR #1462, opened 2024-11-07
Switch to Vitest

miguel-nascimento:
https://vitest.dev/

Why?
- modern: support ESM out of the box (jest requires ton of plugins and additional effort)
- support for HMR - writing tests in watch mode works nice

Most changes of this PR includes: 
- `.toResolve()` -> `.resolves.not.toThrow()`
- `.toReject()` -> `.rejects.toThrow()`
- `.toBeTrue()` -> `.toBe(true)`
- `.toBeFalse()` -> `.toBe(false)`
- `test (inside describe)` -> `it`
- `jsdom` -> `happy-dom`
- `util.test.ts` -> `test-utils.ts`
- `describe` -> `describe.concurrent` (Simple unit tests can be run in concurrent mode. SDK is more complex)

This PR moves Jest to Vitest from:
- sdk
- web3
- encryption
- dlog
- stress

Missing:
- stream-metadata (https://github.com/river-build/river/pull/1468)
- merkle-worker

Vitest has support for workspaces, running a single instance across all packages, but since we have plans for the future to use more of turborepo caching, so it's better to have Vitest in each package.

Should close TOWNS-14617


packages/sdk/package.json
```diff
@@ -9,16 +9,14 @@
...
-        "test:ci": "yarn test:ci:multi:ne",
-        "test:ci:multi:ent:legacy": "USE_LEGACY_SPACES=true RIVER_ENV=local_multi yarn my-jest --group=with-entitlements",
-        "test:ci:multi:ent": "RIVER_ENV=local_multi yarn my-jest --group=with-entitlements --group=with-v2-entitlements",
-        "test:ci:multi:ne": "RIVER_ENV=local_multi_ne yarn my-jest --group=main",
+        "test": "yarn test:multi:ne",
+        "test:watch": "vitest --watch",
+        "test:ci": "yarn test:multi:ne",
+        "test:multi:ent-legacy": "vitest run --config vitest.config.entitlements-legacy.ts",
+        "test:multi:ent": "vitest run --config vitest.config.entitlements.ts",
+        "test:multi:ne": "vitest run --config vitest.config.ts",
```
texuf, line 19: 1) why did you rename these jobs? can we leave the names the same at least?
2) we have --groups that we were selecting the tests that had a ` * @group main` or ``* @group with-entitlements` I guess the filename convention is okay for this? but i don't see that you renamed the tests in the syncAgent folder so maybe you missed those? And if we're going to have custom names I would rather make a custom extension for NO Entitlements than for entitlements, entitlements should be the default moving forward.
`include: ['./src/**/*.test.entitlements.ts', './src/**/*.test.entitlements-v2.ts'],`

texuf: myTest.test.unentitled.ts

miguel-nascimento: Vitest doesn't have `--group` support, that's why we need to rename the test itself or make a separate folder like `/sdk/src/tests/no-entitlements/*.test.ts`

I'll revert the test job name back to the original.
Also thanks for the catch about sync agent tests, I'll add it later.

miguel-nascimento: is that right?
- `group main` -> `myTest.test.noentitlements.ts`
- `group entitlement` -> `myTest.test.ts`
- `group v2-entitlement` -> `myTest.test.entitlement-v2.ts`

texuf: the names are wrong, but it's directionally correct. However, it's ugly. What if we do this
-`no group` -> `myTest.test.ts` runs with `test:unit` (we need to add a new scripts entry `test:unit` for this in sdk)
- `@group main` -> `myTest.test.multi_ne.ts`
- `@group with-entitlements` -> `myTest.test.multi.ts`
- `@group with-v2-entitlements` => `myTest.test.multi.v2.ts`

now they match the environment names
I think @clemire had a plan for the v2 entitlements test? maybe she has an opinion.

miguel-nascimento: done, lmk if it matches the expectation

.vscode/launch.json
```diff
@@ -13,6 +13,43 @@
...
+      "type": "node",
+      "request": "launch",
+      "env": { "RIVER_ENV": "local_multi_ne", "DEBUG": "csb:*,test:*",  "DEBUG_DEPTH":"10" },
+      "program": "${workspaceFolder}/node_modules/.bin/vitest",
+      "args": ["${file}"],
+      "console": "integratedTerminal",
+    },
+    
     {
       "name": "Jest: current file in 'packages/dlog/'",
```
texuf, line 54: 1) how do these launch commands grab the right config?
2) can you clean up the jest commands here?

packages/sdk/package.json
sergekh2: I like folder for test group idea: clear what's going on, easy to follow.

miguel-nascimento:
closed - we're tracking this in the small PRs

PR #1466, opened 2024-11-08
Docs: Initial Commit

texuf:
Migrate Docs from https://github.com/river-build/docs

I went with packages/docs since this has a package.json and installs node packages

I stuck the spellcheck and link linting in with common ci since it took less than a second and didn't seem worth spinning up a whole new job for it.


jterzis (approved):

packages/docs/mint.json
```diff
@@ -0,0 +1,174 @@
...
+    },
+    {
+      "name": "Github",
+      "url": "https://github.com/river-build"
+    }
+  ],
+  "topbarCtaButton": {
+    "name": "Dashboard",
+    "url": "https://dashboard.mintlify.com"
+  },
```
miguel-nascimento, line 43: can we remove this?

packages/docs/package-lock.json
miguel-nascimento, line 1: we want to remove this - we're using the monorepo root yarn.lock

packages/docs/mint.json
texuf: i don't know... we can mess with the config in a follow up commit.

PR #1474, opened 2024-11-09
Notification Service: Fix bugs in challenge signing

texuf:
1) return errors everywhere so it’s possible to debug things
2) make sure we check `len(delegateSig) > 0` before checking delegateExpiryEpochMs
3) use crypto.RecoverEthereumMessageSignerPublicKey which does magic bit toggling to support ethereum signing spec

### DIFF
--- a/core/node/notifications/auth.go
+++ b/core/node/notifications/auth.go
@@ -44,7 +44,7 @@ func (c authenticationChallenge) Verify(
 	// ensure that the auth challenge nor the delegateExpiryEpoch hasn't expired.
 	now := time.Now()
 	if now.After(c.expires) || (len(delegateSig) > 0 && now.After(time.Unix(delegateExpiryEpochMs/1000, 0))) {
-		return RiverError(Err_UNAUTHENTICATED, "authentication expired").Tag("expires", c.expires).Tag("delegateExpiryEpochMs", delegateExpiryEpochMs)
+		return RiverError(Err_UNAUTHENTICATED, "authentication expired", "expires", c.expires, "delegateExpiryEpochMs", delegateExpiryEpochMs)
 	}
 
 	// ensure that the signature that was calculated with:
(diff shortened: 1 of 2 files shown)


core/node/notifications/auth.go
```diff
@@ -36,15 +35,16 @@ type (
...
 	delegateSig []byte,
 	delegateExpiryEpochMs int64,
-) bool {
+) error {
 	// ensure that the auth challenge nor the delegateExpiryEpoch hasn't expired.
 	now := time.Now()
-	if now.After(c.expires) { // ignored for now || now.After(time.Unix(delegateExpiryEpochMs/1000, 0)) {
-		return false // expired
+	if now.After(c.expires) || (len(delegateSig) > 0 && now.After(time.Unix(delegateExpiryEpochMs/1000, 0))) {
+		return RiverError(Err_UNAUTHENTICATED, "authentication expired").Tag("expires", c.expires).Tag("delegateExpiryEpochMs", delegateExpiryEpochMs)
```
bas-vk, line 47: Nitpick, RiverError accepts key/value pairs as tags. No need to chain Tag calls.

core/node/notifications/auth.go
```diff
@@ -60,22 +60,25 @@ func (c authenticationChallenge) Verify(
...
 	if err != nil {
-		return false
+		return RiverError(Err_UNAUTHENTICATED, "error recovering signer public key").Tag("user", c.userID).Tag("error", err)
 	}
 
 	signerAddress := crypto.PublicKeyToAddress(signerPubKey)
 
 	if len(delegateSig) == 0 {
-		return c.userID == signerAddress
+		if ( c.userID == signerAddress) {
```
bas-vk, line 73: Could you format this file?

core/node/notifications/auth.go
```diff
@@ -136,8 +139,9 @@ func (s *Service) FinishAuthentication(
 
 	// make sure that the caller has access to the private key from which user id was derived
 	chal := raw.(*authenticationChallenge)
-	if !chal.Verify(challenge, msg.GetSignature(), msg.GetDelegateSig(), msg.GetDelegateExpiryEpochMs()) {
-		return nil, RiverError(Err_PERMISSION_DENIED, "bad signature").Tag("user", userID)
+	err := chal.Verify(ctx, challenge, msg.GetSignature(), msg.GetDelegateSig(), msg.GetDelegateExpiryEpochMs())
+	if err != nil {
+		return nil, RiverError(Err_PERMISSION_DENIED, "bad signature").Tag("user", userID).Tag("error", err)
```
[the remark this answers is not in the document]
texuf: I don't think that's a good idea because clients will making this request and need handle errors in the response. 
The api should provide a consistent error message for the case where the challenge failed to verify, so that we can properly handle it and accurately retry or show an error to the user. 
If we're just forwarding along an error that came from a random library we lose clarity for the client implementation and set ourselves up for breaking changes if the library gets refactored or replaced

core/node/notifications/auth.go
texuf: i was just following the convention in the file. I can replace them all.

PR #1479, opened 2024-11-10
Upgrade to go 1.23.3

sergekh2:
Bring all deps to latest: some older deps are not compatible.


bas-vk:
There is also a Dockfile that needs to be bumped.

texuf (approved):

PR #1485, opened 2024-11-11
docs: react-sdk core, space, channel and members

miguel-nascimento:
This PR adds tsdocs around the Space module in the both react-sdk and core sdk.


packages/web3/src/types.ts
```diff
@@ -1,12 +1,28 @@
...
+    /**
+     * The timestamp of when the space was created.
+     * It is a {@link https://docs.ethers.org/v5/api/utils/bignumber/ ethers.BigNumber} serialized as a `string`.
+     */
+    createdAt: string
+    /** The River `spaceId` of the space. */
     networkId: string
+    /** The name of the space. */
     name: string
+    /** The on-chain address of the space owner. */
```
miguel-nascimento, line 13: i think space creator is the right naming here right?

packages/react-sdk/src/useSendMessage.ts
```diff
@@ -18,6 +24,11 @@ export const useSendMessage = (
     const { action: sendMessage, ...rest } = useAction(room, 'sendMessage', config)
 
     return {
+        /** Sends a message to the stream.
+         * @param message - The message to send.
+         * @param options - Additional options for the message.
+         * @returns The event id of the message.
+         */
```
miguel-nascimento, line 31: I couldn't find a way to reference and show the tsdoc that we have on `Channel.sendMessage` 🥲

packages/web3/src/ContractTypes.ts
```diff
@@ -124,22 +135,31 @@ export interface ChannelMetadata {
...
 export interface RoleEntitlements {
+    /** The id of the role. */
     roleId: number
+    /** The name of the role. */
     name: string
+    /** The permissions that this role has. @see {@link Permission} */
     permissions: Permission[]
+    /** The userIds that are in this role. */
     users: string[]
+    /** @internal The River struct that represents the rule data of the role {@link VersionedRuleData}. */
```
miguel-nascimento, line 162: Thats "internal" right?

packages/sdk/src/sync-agent/spaces/models/channel.ts
```diff
@@ -67,19 +75,31 @@ export class Channel extends PersistedObservable<ChannelModel> {
...
     async sendMessage(
         message: string,
         options?: {
+            /** If set, this message will be linked to a thread with the specified message. */
             threadId?: string
+            /** If set, this message will be linked as a reply to the specified message. */
             replyId?: string
+            // TODO: would be great to improve the usability here. Its not that ergonomic to pass the `ChannelMessage_Post_Mention` payload itself
+            // and its a bit unclear about what are the intetion of the fields.
+            /** The users that are mentioned in the message */
```
miguel-nascimento, line 100: > its a bit unclear about what are the intention of the fields

For example, to create a `ChannelMessage_Post_Mention` you can pass `userId`, `displayName` and type `atChannel` or `atRole` or leave it empty for user mentions

It's not clear for me how to mention a specific channel. e.g: mention `atChannel`. But which channel? There's no `channelId` field. 

For the role mention, `atRole` mention behavior proto expects a `ChannelMessage_Post_RoleMention` which holds the `roleId`

texuf: atRoles aren't currently implemented
we should just accept the PlainMessage<of the proto> and call it a day

texuf (approved):

Issue #1494, opened 2024-11-12
Do not ignore membership check errors on space membership checks during entitlement evaluation


clemire:
### Description

When space memberships are checked, at this time errors are ignored. However if no checks return a positive result, we should not consider that the user has been determined to be un-entitled unless there were no errors encountered when evaluating space membership. This needs to be fixed.

PR #1495, opened 2024-11-12
Capture membership evaluation errors when no wallets evaluate as space members

clemire:
When multiple linked wallets are checked for space membership and none of them are conclusively determined to be members, we ignore any errors in invoked checks and return false. However, if any wallet check failed, we cannot conclusively determine that the user was not a space member and should return an entitlement evaluation error so that the user will not be considered unentitled by the caller.

This PR captures all errors generated when conducting membership checks and returns an aggregate error in the case that all checks evaluated and none of them were positive. The user is consider "not a member" only if there are no errors whatsoever.

#1494


core/node/auth/auth_impl.go
```diff
@@ -686,17 +686,19 @@ func (ca *chainAuth) checkMembership(
...
 	results chan<- bool,
+	errors chan<- error,
 	wg *sync.WaitGroup,
 ) {
 	log := dlog.FromCtx(ctx)
 	defer wg.Done()
 	isMember, err := ca.spaceContract.IsMember(ctx, spaceId, address)
 	if err != nil {
-		log.Warn("Error checking membership", "err", err, "address", address.Hex(), "spaceId", spaceId)
+		log.Error("Error checking membership", "err", err, "address", address.Hex(), "spaceId", spaceId)
```
clemire, line 696: This could actually be downgraded to a warning as it will fire if another wallet evaluates to have membership, because we terminate remaining checks as soon as a single wallet is conclusively determined to be a member.

core/node/auth/auth_impl.go
```diff
@@ -779,9 +800,53 @@ func (ca *chainAuth) checkEntitlement(
 		}
 	}
 
+	// Look for any returned errors. If at least one check was positive, then we ignore any subsequent
+	// errors. Otherwise we will report an error result since we could not conclusively determine that
+	// the user was not a space member.
+	for err := range isMemberError {
```
sergekh2, line 806: reading errors will always wait for all goroutines to complete, even if isMember it true.

core/node/auth/auth_impl.go
```diff
@@ -779,9 +800,53 @@ func (ca *chainAuth) checkEntitlement(
...
+				"userId",
+				args.principal,
+				"spaceId",
+				args.spaceId,
+				"wallets",
+				wallets,
+				"aggregateError",
+				membershipError,
+			)
+			return &boolCacheResult{allowed: false}, membershipError
```
sergekh2, line 835: why not to return nil value in case of the error? would it lead to more robust code when incorrect result can't be cached?

core/node/auth/auth_impl.go
```diff
@@ -755,22 +769,29 @@ func (ca *chainAuth) checkEntitlement(
 
 	isMemberCtx, isMemberCancel := context.WithCancel(ctx)
 	defer isMemberCancel()
+
 	isMemberResults := make(chan bool, 1)
```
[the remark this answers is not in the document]
clemire: Thanks for the catch!

core/node/auth/auth_impl.go
```diff
@@ -779,9 +800,64 @@ func (ca *chainAuth) checkEntitlement(
...
+				Func("checkEntitlement").
+				Tag("principal", args.principal).
+				Tag("permission", args.permission).
+				Tag("wallets", args.linkedWallets).
+				Tag("spaceId", args.spaceId)
+		}
+	}
+
+	if !isMember {
+		if membershipError != nil {
```
sergekh2, line 834:
```suggestion

```
seems like these two pairs of consecutive ifs have same conditions

sergekh2 (approved):

bas-vk (approved):

PR #1509, opened 2024-11-13
sdk(sync-agent): leave space, channel and gdm

miguel-nascimento:
Tests are failing around gdm & channel. I'll investigate why later.

Keeping it here as a draft just so we keep a track of those missing core features


packages/sdk/src/sync-agent/syncAgents.test.ts
```diff
@@ -215,4 +217,59 @@ describe('syncAgents.test.ts', () => {
             { timeoutMS: 10000 },
         )
     })
+
+    test('leave gdm', async () => {
```
texuf, line 221: these tests are run in sequence with the same users, can you leave the existing gdm?

packages/react-sdk/src/useLeaveGdm.ts
```diff
@@ -0,0 +1,15 @@
+'use client'
+
+import type { Gdms } from '@river-build/sdk'
+import { type ActionConfig, useAction } from './internals/useAction'
+import { useSyncAgent } from './useSyncAgent'
+
```
texuf, line 6: there's no doc strings!!!

PR #1514, opened 2024-11-13
generated: publish deployments.json to npm

miguel-nascimento:
While using the sdk from npm I found this error:

```
[ERROR] Could not resolve "@river-build/generated/config/deployments.json"
node_modules/@river-build/web3/dist/IStaticContractsInfo.js:1:28:
  1 │ ...entsJson from '@river-build/generated/config/deployments.json' assert { type...
    ╵                  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
You can mark the path "@river-build/generated/config/deployments.json" as external to  
exclude it from the bundle, which will remove this error and leave the unresolved path in  
the bundle.
```

The publish to npm script run build, but the generated package doesnt have a build script. Which is maybe causing the package to be published without a `config/deployments.json`

I managed to make it work using `npm link`, but I added `.npmignore` following [this trick](https://stackoverflow.com/a/47127547), since npm can use git ignore to ignore the config folder.


miguel-nascimento:
merging so we can test it asap

texuf (commented):
nice. i wonder how this is working in my tools project.

PR #1516, opened 2024-11-14
Parallel read using GetPaginatedStreams

sergekh2:


core/node/registries/river_registry_contract.go
```diff
@@ -393,13 +423,133 @@ func (c *RiverRegistryContract) ForAllStreams(
...
+// If cb returns false forAllStreams returns.
+func (c *RiverRegistryContract) forAllStreamsParallel(
+	ctx context.Context,
+	blockNum crypto.BlockNumber,
+	cb func(*GetStreamResult) bool,
+) error {
+	log := dlog.FromCtx(ctx)
+
+	numWorkers := c.Settings.ParallelReaders
+	if numWorkers <= 1 {
```
jterzis, line 455: I don't think this will ever be hit, b/c of https://github.com/river-build/river/pull/1516/files#diff-15b39610ef19b415cd3d948a79ebe97f7c9fcee7a2cb83316a584289057b6d2fR367.

core/node/registries/river_registry_contract.go
```diff
@@ -393,13 +423,133 @@ func (c *RiverRegistryContract) ForAllStreams(
...
+					break OuterLoop
+				}
+			}
+			if taskCounter.Load() == 0 {
+				break OuterLoop
+			}
+		case receivedErr := <-chErrors:
+			err = receivedErr
+			break OuterLoop
+		case <-ctx.Done():
```
jterzis, line 532: Should we set a reasonable ctx timeout when calling parallel stream get ? If not, then there's the potential for a deadlock if chErrors, chResults get stuck where this just hangs.

core/node/registries/river_registry_contract.go
```diff
@@ -11,6 +11,8 @@ import (
 	"github.com/ethereum/go-ethereum/accounts/abi/bind"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/gammazero/workerpool"
+	"go.uber.org/atomic"
```
jterzis, line 15: Do you need to run `go mod tidy` or something similar to update go.mod file ?

bas-vk: Lets use `sync/atomic.Int64`.

core/node/registries/river_registry_contract.go
sergekh2: this handles the case if setting was set explicitly

core/node/registries/river_registry_contract.go
sergekh2: yes, it sets timeout and number of retries from settings

jterzis (approved):

core/node/registries/river_registry_contract.go
```diff
@@ -393,13 +431,155 @@ func (c *RiverRegistryContract) ForAllStreams(
...
+	startTime := time.Now()
+	lastReport := time.Now()
+	var taskCounter atomic.Int64
+	for i := int64(0); i < numStreams; i += pageSize {
+		taskCounter.Add(1)
+		pool.Submit(func() {
+			streams, _, err := c.callGetPaginatedStreamsWithBackoff(ctx, blockNum, i, i+pageSize)
+			if ctx.Err() == nil {
+				if err == nil {
+					chResults <- streams
```
bas-vk, line 514: On both `chResults` and `chErrors` you can use a write + read select to prevent a deadlock entirely. Something like this:

```
select {
case chResults <- streams:
   break 
case <-ctx.Done():
   break
}
```
This takes away the need to check if the ctx was cancelled with an error.

PR #1522, opened 2024-11-14
react-sdk: docgen script

miguel-nascimento:
Adapted from the [docgen script of wevm/ox](https://github.com/wevm/ox/tree/main/scripts/docgen), which uses ts-morph and api-extractor to extract ts-docs and build definitions.

TypeDoc didn't work as we expected, it wasn't solving external types.
I had an issue with ts-doc `@link` references, decided to skip for now, lets get the docs in a good shape first and then we can add it later.

Missing
- render tsdocs of return type
- make the script readable and simpler

closes TOWNS-14936, TOWNS-14934


packages/workers/merkle-worker/package.json
```diff
@@ -32,7 +32,7 @@
         "miniflare": "^3.20240524.1",
         "prettier": "^2.8.8",
         "ts-jest": "^29.1.1",
-        "tsx": "^4.7.1",
```
texuf, line 35: can we upgrade tsx in its own pr?

texuf (approved):

PR #1531, opened 2024-11-15
Support for in-place migration with migration tool

clemire:
Ran on my local machine on a schema copied as-is from gamma.


sergekh2 (approved):

clemire:
I removed the documentation on the migrate option so that no one will use it, but I'm merging this to get access to the `inspect` command and some improvements to other commands for debugging failed migrations.

PR #1562, opened 2024-11-19
docgen: prune mdx that arent used anymore

miguel-nascimento:
Doc gen is now syncing the entry points modules <> docs api folder.
If a function doesn't exist anymore, the doc api page should not exist as well.


texuf (approved):
good catch.

PR #1580, opened 2024-11-20
move stress package to vitest

miguel-nascimento:
stacked on top of #1576

I'm breaking #1462 into small PRs

Moving the stress package from jest to vitest.


texuf (commented):
i don't see these tests running in ci

texuf (approved):
found them

PR #1589, opened 2024-11-21
notifications: drop 0x prefix in notification payload

bas-vk:


bas-vk:
Succeeded by another PR

PR #1595, opened 2024-11-21
create-river-build-app: playground template

miguel-nascimento:
This PR includes a new template for the `create-river-build-app` script: `playground`
It's on by default.


packages/create-river-build-app/README.md
```diff
@@ -24,3 +40,15 @@ If you want to create a new app in a different directory, you can specify the di
...
+
+```bash
+yarn create river-build-app my-app --template react-ts
+```
+
+Available templates:
+
+- react-ts (default)
+- react
+- playground
```
texuf, line 54: Can we make playground the default?

texuf (commented):
Nice

texuf (approved):

packages/create-river-build-app/README.md
miguel-nascimento: nice catch! it is the default, but it was wrong in the readme 🤦‍♂️

PR #1605, opened 2024-11-22
create-river-build-app(playground): get latest sdk tag to clone

miguel-nascimento:
Clone from main can lead into a broken playground state

Npm publishes our packages once a day, but we can change the package and playground N times a day.
If someone gets playground from main but with outdated npm packages, playground will fail to build/be in a broken state.

In order to fix this, we need to get the playground from the latest sdk published tag.

### DIFF
--- a/packages/create-river-build-app/src/modules/clone-playground.ts
+++ b/packages/create-river-build-app/src/modules/clone-playground.ts
@@ -9,6 +9,7 @@ export const clonePlayground = async (cfg: CreateRiverBuildAppConfig) => {
     console.log(picocolors.blue('\nCloning River Build Playground...'))
 
     const result = await cloneRepo(cfg)
+    if (!result) return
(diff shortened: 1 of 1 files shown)


packages/create-river-build-app/src/modules/clone-playground.ts
```diff
@@ -130,3 +139,45 @@ const fixTsConfig = async (cfg: CreateRiverBuildAppConfig) => {
...
+
+function getLatestSdkTag(): string | null {
+    const tagsResult = spawn.sync(
+        'git',
+        ['ls-remote', '--tags', 'https://github.com/river-build/river.git', 'sdk-*'],
+        { encoding: 'utf8' },
+    )
+
+    if (tagsResult.status !== 0 || !tagsResult.stdout) return null
+
```
texuf, line 151: i don't love this, but i understand why

texuf (approved):

PR #1609, opened 2024-11-22
Update migrate tool readme

jterzis:


core/tools/migrate_db/README.md
```diff
@@ -1,26 +1,18 @@
...
 
-    # Inspect specific stream content
-    ./river_migrate_db source inspect <streamId>
-    ./river_migrate_db target inspect <streamId>
+./river_migrate_db help
+./river_migrate_db source test # test source config
+./river_migrate_db target test # test target config
 
-    # reconfigure container to use target db
+Note: shutdown node process connected to source DB during migration process below.
```
clemire, line 13: You actually don't need to shut down the node for the `target create` and `target init` commands, but if it makes the docs easier to read, this isn't the biggest deal.

clemire (approved):

PR #1614, opened 2024-11-22
playground: create dm feature

miguel-nascimento:
Now if you login with a signer (ie metamask), you can create a dm in the playground


texuf:
if you're already a member of a space right?

miguel-nascimento:
correct

texuf (approved):

Issue #1621, opened 2024-11-23
Cached linked wallets, using the entitlement cache to determine positive / negative ttl


clemire:
### Description

Cache linked wallets, busting the cache whenever an entitlement result for a principal comes back negative. This supports the UX where a user tries to join a channel / space, is denied, adds a linked wallet, and immediately tries again, but hopefully alleviates the high rpc call volume on base contracts for linked wallets during scrubbing.

clemire:
https://github.com/river-build/river/pull/1635

PR #1635, opened 2024-11-25
Linked Wallet Caching

clemire:
Cache linked wallets, busting the cache as soon as any entitlement result comes back negative for the user.

We cannot decide whether to use a positive or negative cache ttl for linked wallets based on the result of the linked wallets call itself, but we want to keep a short linked wallet ttl whenever an entitlement is negative because it gives the user a chance to update their entitlement result by linking wallets whenever they are unentitled.

This may not be super relevant on the stream node at all because the client proactively evaluates channel and space join entitlements and does not allow either unless the user is entitled... it's possible it's best to just go ahead and use the positive cache.


clemire:
#1621

texuf (approved):

core/node/auth/auth_impl.go
```diff
@@ -670,21 +691,64 @@ func (ca *chainAuth) isEntitledToChannel(ctx context.Context, cfg *config.Config
...
 	if ca.walletLinkContract == nil {
 		log.Warn("Wallet link contract is not setup properly, returning root key only")
-		return []common.Address{wallet}, nil
+		return []common.Address{args.principal}, nil
 	}
 
-	wallets, err := entitlement.GetLinkedWallets(ctx, wallet, ca.walletLinkContract, nil, nil, nil)
+	userCacheKey := newArgsForLinkedWallets(args.principal)
+	// We want fresh linked wallets when evaluating space and channel joins, key solicitations,
+	// and user scrubs, all of which request the Read permission.
```
salzbrenner, line 726: just confirming, joining a space checks Permission.Read and not Permission.JoinSpace?

Where does Permission.JoinSpace come into play? Does it at all?
I just looked this up and in Permissions.sol it's `Permissions.JoinSpace` whereas in permissions.go its `PermissionJoin`

clemire: The Read check is actually a bug I'm fixing in another PR, we have a few of those in our auth code still. I'll add you as a reviewer.

The difference in string constant values for Join and JoinSpace probably needs to be fixed.

So we actually ought to be doing something more subtle here. The contract checks the Join permission and if the user qualifies, it issues the token. The node itself then doesn't need to check if the user also qualifies for the join, it just needs to check that the user has a membership token, which was issued by the contract, who is the ultimate decider of who joins the space. That's the PR in the works. I proactively put in logic that would be compatible with this change.

PR #1642, opened 2024-11-26
Add try catch blocks around cached promises in stream-metadata service

texuf:
1. client requests image for stream before it’s registered
2. service returns 404
3. client busts cache, requests again
4. the promise from the previous request is still cached, nothing is retried, 404 is returned again
5. restarting the server makes everything go away and get better

classic blunder

fix: add try catch around all promise caching
future: we should probably have a cached promise abstraction, there are a lot of gotchas here.


mechanical-turk (approved):
fascinating

PR #1645, opened 2024-11-26
sdk: string autocomplete for river environments

miguel-nascimento:
It is a bit tricky when you don't have all the river environment information context to create a river config

this PR adds autocomplete for river environments, I'll add a follow-up adding docs about environments

image

It's a loose autocomplete, this function still accepts any other `string` values.

### DIFF
--- a/packages/sdk/src/riverConfig.ts
+++ b/packages/sdk/src/riverConfig.ts
@@ -8,6 +8,9 @@ import {
 } from '@river-build/web3'
 import { isDefined } from './check'
 import { check } from '@river-build/dlog'
+import type DeploymentsJson from '@river-build/generated/config/deployments.json'
+
+type EnvironmentId = keyof typeof DeploymentsJson
 
 function getEnvironmentId(): string {
     if (typeof process === 'object') {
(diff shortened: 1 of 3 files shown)


packages/sdk/src/riverConfig.ts
```diff
@@ -117,7 +120,9 @@ export function makeBaseChainConfig(environmentId?: string) {
 
 export type RiverConfig = ReturnType<typeof makeRiverConfig>
 
-export function makeRiverConfig(inEnvironmentId?: string) {
+// trick to get autocomplete for environment ids while allowing any string
+// eslint-disable-next-line @typescript-eslint/ban-types
+export function makeRiverConfig(inEnvironmentId?: EnvironmentId | (string & {})) {
```
miguel-nascimento, line 125: more info about `<string-literal-type> | (string & {}`
https://www.totaltypescript.com/typescript-5-3#automatic-loose-autocomplete-on-strings

texuf (commented):
I don't want to do this because we have lots of conflict over how we are ingesting the environments. These should really be environment variables, not source code that is compiled in...

PR #1648, opened 2024-11-26
Turns membership base into lib

giuseppecrj:


shuhuiluo (approved):
I think the library pattern is better suited for primitives whose data structure is tightly coupled with the logic which is unlikely to change.

giuseppecrj:
> I think the library pattern is better suited for primitives whose data structure is tightly coupled with the logic which is unlikely to change.

Yeah, not sure the juice is worth the squeeze with functionality like this. Interesting approach but moving everything into libs will have to be on a case-by-case basis.

PR #1652, opened 2024-11-27
[RIVER-1612] instrument clients with eth_call metrics

clemire:
Local node metrics output:

```
# HELP river_stream_eth_calls Number of eth_calls made by an instrumented client
# TYPE river_stream_eth_calls counter
river_stream_eth_calls{chain_id="31337",method_name="",status="ok"} 27
river_stream_eth_calls{chain_id="31337",method_name="Channels.getChannel",status="ok"} 4
river_stream_eth_calls{chain_id="31337",method_name="EntitlementDataQueryable.getChannelEntitlementDataByPermission",status="ok"} 6
river_stream_eth_calls{chain_id="31337",method_name="EntitlementDataQueryable.getEntitlementDataByPermission",status="ok"} 6
river_stream_eth_calls{chain_id="31337",method_name="WalletLink.getRootKeyForWallet",status="ok"} 26
river_stream_eth_calls{chain_id="31337",method_name="WalletLink.getWalletsByRootKey",status="ok"} 26
river_stream_eth_calls{chain_id="31337",method_name="balanceOf",status="ok"} 19
river_stream_eth_calls{chain_id="31338",method_name="",status="ok"} 3
river_stream_eth_calls{chain_id="31338",method_name="StreamRegistry.getStreamWithGenesis",status="NOT_FOUND"} 32
river_stream_eth_calls{chain_id="31338",method_name="StreamRegistry.getStreamWithGenesis",status="ok"} 65
```


core/node/crypto/client.go
```diff
@@ -6,136 +6,350 @@ import (
...
+	if len(*data) > 4 {
+		selector := hex.EncodeToString((*data)[:4])
+		var defined bool
+		methodName, defined = GetSelectorMethodName(selector)
+		if !defined {
+			methodName = selector
+		}
+	}
+	return methodName
+}
```
sergekh2, line 184: Lets uint32 as key for selectors: this will eliminate the need to allocate in most of the cases.
For slices usually it's not required to pass them by reference, it's assumed passing them as-is is fast enough.

As a general style rule, lets not use named return args (since it is rarely used feature, it makes code less readable).

Something like this:
```suggestion
func getMethodName(data []byte) (string) {
    var methodName string
	if len(data) >= 4 {
		selector := binary.BigEndian.Uint32(data)
		var defined bool
		methodName, defined = GetSelectorMethodName(selector)
		if !defined {
			methodName = hex.EncodeToString(date[:4])
		}
	}
	return methodName
}
```

core/node/crypto/client.go
```diff
@@ -6,136 +6,350 @@ import (
...
+	client *ethclient.Client,
+	metrics infra.MetricsFactory,
+	tracer trace.Tracer,
+) *otelEthClient {
+	var ethCalls *prometheus.CounterVec
+	if metrics != nil {
+		ethCalls = metrics.NewCounterVecEx(
+			"eth_calls",
+			"Number of eth_calls made by an instrumented client",
+			"method_name",
```
sergekh2, line 92: please add `chain_id` as first label.

please take a look into adding `status` as last label.
`ok` on success
revert reason if it's possible to extract it like this:
https://github.com/river-build/river/blob/b22df64b5a4e86ce78e81f3298704f85f938ac4e/core/node/base/error.go#L267
`revert` if error is rpc.DataError but reason can't be extracted
`fail` if this is some other error.

clemire: See updated output above for examples of capturing revert reasons

core/node/crypto/client.go
```diff
@@ -2,140 +2,404 @@ package crypto
...
+				return reason
+			}
+		}
+		return "revert"
 	}
+	return "fail"
+}
 
-	return ic.Client.CallContract(ctx, msg, blockNumber)
+func getMethodName(data *[]byte) string {
```
bas-vk, line 186: Why is data `*[]byte` and not just `[]byte`?

core/node/crypto/client.go
```diff
@@ -2,140 +2,404 @@ package crypto
...
+			status = extractCallErrorStatus(err)
+		}
+
+		if methodName == "" {
+			methodName = getMethodName(&msg.Data)
+		}
+		ic.ethCalls.With(
+			prometheus.Labels{
+				"chain_id":    ic.chainId,
+				"method_name": methodName,
```
bas-vk, line 229: The txpool labels transaction metrics with `func_selector`. If the read operations move to this new convention with method name the txpool metrics should also move to this new convention. And as a result the txpool dashboard in dd needs to be updated as well.

core/node/crypto/client.go
clemire: To avoid copy by value of the array.

core/node/crypto/client.go
```diff
@@ -2,140 +2,404 @@ package crypto
...
+				"status":      status,
```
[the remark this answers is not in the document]
clemire: I think that could potentially be very high cardinaility, on the order of number of spaces (or more), and it's better not to have high cardinality tags on a metric.

core/node/crypto/client.go
```diff
@@ -2,140 +2,404 @@ package crypto
...
+// NewInstrumentedEthClient wraps an Ethereum client and adds open-telemetry tracing and metrics
+// collection.
+func NewInstrumentedEthClient(
+	client *ethclient.Client,
+	chainId uint64,
+	metrics infra.MetricsFactory,
+	tracer trace.Tracer,
+) *otelEthClient {
+	var ethCalls *prometheus.CounterVec
+	if metrics != nil {
```
sergekh2, line 95: in other parts of the code to eliminate these nil checks it's assumed that metrics is always initialized. In test it's empty metrics are initialized like this: `infra.NewMetricsFactory(nil, "", ""),` let's drop nil checks here as well.

core/node/crypto/client.go
```diff
@@ -2,140 +2,404 @@ package crypto
...
+
+		methodName = getMethodName(&msg.Data)
+		span.SetAttributes(attribute.String("method_name", methodName))
+	}
 
-	if len(msg.Data) >= 4 {
-		span.SetAttributes(attribute.String("func_selector", hex.EncodeToString(msg.Data[:4])))
+	data, err := call()
+
+	if ic.ethCalls != nil {
```
sergekh2, line 217: (previous comment) same here - this nil check can be removed and so on.

core/node/crypto/client.go
sergekh2: this a slice, array would be copied, but go slice itself is a reference type, and doesn't copy underlying buffer 

(ironically enough this sometimes leads to the bugs when underlying buffer is modified and `slices.Clone` should be used, but this is not the case here)

sergekh2 (approved):

PR #1669, opened 2024-11-28
local claim condition script

giuseppecrj:
This creates a local claim condition to the airdrop contract for testing, Update the setUp function in InteractClaimCondition to add the wallet you're using locally to set a claim condition. you can run it as many times as you can by doing the following steps:

```
export RIVER_ENV="local_multi"
cd contracts
make interact-any-local context=$RIVER_ENV rpc=base_anvil contract=InteractClaimCondition
```


ttillberg (approved):
Thank you!

jterzis (approved):

shuhuiluo (approved):

PR #1671, opened 2024-11-28
check total claimable supplied balance

giuseppecrj:


jterzis (approved):

contracts/src/tokens/drop/DropFacetBase.sol
```diff
@@ -138,9 +138,19 @@ abstract contract DropFacetBase is IDropFacetBase {
           DropFacet__ClaimConditionsNotInAscendingOrder.selector
         );
       }
+      _verifyEnoughBalance(
+        newCondition.currency,
+        IERC20(newCondition.currency).balanceOf(address(this)) +
+          newCondition.maxClaimableSupply
+      );
```
shuhuiluo, line 145: This will fail.

shuhuiluo (changes requested):

Issue #1701, opened 2024-12-03
Remove old schema from schema migrations


clemire:
### Description

Now that all nodes in omega are running the new schema, let's remove the old schema. Existing nodes in gamma and alpha that are still hosting unmigrated streams will be unaffected, but there's not reason to keep these tables as we have made using the migrated stream access pattern the default for nodes now.

PR #1703, opened 2024-12-03
[RIVER-1701] Remove support for legacy (unmigrated) stream schema

clemire:
#1701


core/node/storage/pg_stream_store.go
```diff
@@ -376,7 +376,7 @@ func (s *PostgresStreamStore) createStreamArchiveStorageTx(
 			`INSERT INTO es (stream_id, latest_snapshot_miniblock, migrated) VALUES ($1, -1, false);
 			CREATE TABLE {{miniblocks}} PARTITION OF miniblocks FOR VALUES IN ($1);`,
 			streamId,
-			false,
+			s.config.MigrateStreamCreation,
```
clemire, line 379: Bugfix: the archiver was always using the legacy schema.

core/node/testutils/dbtestutils/db.go
```diff
@@ -76,14 +76,15 @@ func ConfigureDB(ctx context.Context) (*config.DatabaseConfig, string, func(), e
...
-			NumPartitions: 4,
+			Host:                  "localhost",
+			Port:                  5433,
+			User:                  "postgres",
+			Password:              "postgres",
+			Database:              "river",
+			Extra:                 "?sslmode=disable&pool_max_conns=1000",
+			StartupDelay:          2 * time.Millisecond,
+			NumPartitions:         4,
+			MigrateStreamCreation: true,
```
sergekh2, line 87: after this change it seems this setting is no longer required?

sergekh2 (approved):

bas-vk (approved):

core/node/testutils/dbtestutils/db.go
clemire: Yes, I was going to do in a follow-up PR but went ahead and put in here so the code would not be in a confusing state.

clemire:
@sergekh2 I updated this PR according to your comment to remove the migration setting, and i also updated the migration tool, but I did not test the migration tool since we don't really need it right now.  I saw some flakes locally but after making changes they disappeared - if 5 consecutive CI tests pass I will consider this solid.

clemire:
> @sergekh2 I updated this PR according to your comment to remove the migration setting, and i also updated the migration tool, but I did not test the migration tool since we don't really need it right now. I saw some flakes locally but after making changes they disappeared - if 5 consecutive CI tests pass I will consider this solid.

Ok, all 5 runs after a lint fix are totally green.

clemire:
Note: I can drop the "migrated" column in a follow-up PR, I did not do that here.

sergekh2 (approved):

PR #1710, opened 2024-12-03
check in feature on river points

giuseppecrj:


contracts/src/tokens/points/RiverPoints.sol
```diff
@@ -65,6 +66,52 @@ contract RiverPoints is IERC20Metadata, IRiverPoints, OwnableBase, Facet {
         points = protocolFee * 3_000_000;
       }
     }
+
+    if (action == Action.CheckIn) {
+      (uint256 lastCheckIn, uint256 streak) = abi.decode(
+        data,
+        (uint256, uint256)
+      );
+      (points, ) = CheckIn.getPointsAndStreak(lastCheckIn, streak);
```
shuhuiluo, line 75: Decode the address and read from storage?

shuhuiluo (approved):

contracts/src/tokens/points/RiverPoints.sol
giuseppecrj: block.timestamp

pfives (approved):

PR #1715, opened 2024-12-04
don't send push notifications for edits/redactions

erikolsson:
This PR stops the notification service from sending push notifications for edits and redactions.


andreyz (approved):

bas-vk (approved):
Only remark is that the early return allows for a fast return before allocating and executing the initial part of the function logic.

bas-vk (approved):

bas-vk (approved):

texuf (commented):
Seems good for now

PR #1717, opened 2024-12-04
filter sender from notification recipients field

erikolsson:
I believe this will prevent notifications from displaying "You and Charlie" in the title for DMs.
I'm not 100% sure if the comparison between `member` and `sender.Hex()` is valid?

### DIFF
--- a/core/node/notifications/event_processor.go
+++ b/core/node/notifications/event_processor.go
@@ -78,15 +78,8 @@ func (p *MessageToNotificationsProcessor) OnMessageEvent(
 	}
 	l.Debug("Process event")
 
-	usersToNotify := make(map[common.Address]*types.UserPreferences)
 	kind := "new_message"
-	recipients := mapset.NewSet[common.Address]()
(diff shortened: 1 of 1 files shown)


andreyz (approved):

core/node/notifications/event_processor.go
```diff
@@ -297,7 +298,11 @@ func (p *MessageToNotificationsProcessor) sendNotification(
 
 	var receivers []string
 	if channelID.Type() == shared.STREAM_DM_CHANNEL_BIN || channelID.Type() == shared.STREAM_GDM_CHANNEL_BIN {
-		receivers = members.ToSlice()
+		for _, member := range members.ToSlice() {
+			if member != sender.Hex() {
```
bas-vk, line 302: `sender.Hex()` creates each time a new hex string while it is stable during the iteration. Do something like `senderHex := sender.Hex()` outside of the loop and do the comparison with that inside the loop.

texuf (commented):
We do this on the client on web

PR #1724, opened 2024-12-05
Exclude */tests/* from generate_sdk_index.sh

texuf:
We moved the tests in to a folder, let's not export the helper scripts


miguel-nascimento:
good catch thank you

PR #1732, opened 2024-12-05
Reduce histogram buckets and tune bucket range and intervals

clemire:
I partnered with @mechanical-turk to discuss our highest cardinality metrics and to shave down cost while maintaining observability where possible. Here is a proposal to reduce the granularity of buckets for several histogram metrics.

### DIFF
--- a/core/node/crypto/chain_txpool.go
+++ b/core/node/crypto/chain_txpool.go
@@ -165,8 +165,8 @@ func newPendingTransactionPool(
 		"chain_id", "address", "status",
 	)
 
-	// 2, 6, 10, 14, 18s
-	txPoolHistogramBuckets := prometheus.LinearBuckets(2.0, 4.0, 5)
+	// 1, 3, 5, 7, 9s
+	txPoolHistogramBuckets := prometheus.LinearBuckets(1.0, 2.0, 5)
 	transactionTotalInclusionDuration := metrics.NewHistogramVecEx(
 		"txpool_tx_total_inclusion_duration_sec",
 		"How long it takes before a transaction is included in the chain since first submit",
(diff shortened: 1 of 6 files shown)


core/node/crypto/chain_txpool.go
```diff
@@ -164,15 +164,18 @@ func newPendingTransactionPool(
 		"txpool_processed", "Number of submitted transactions that are included in the chain",
 		"chain_id", "address", "status",
 	)
+
+	// 2, 6, 10, 14, 18s
```
clemire, line 168: @bas-vk this look ok to you? Let me know if you have an alternate preference.

core/node/infra/metrics_publisher.go
```diff
@@ -13,25 +13,26 @@ import (
 	"github.com/river-build/river/core/node/dlog"
 )
 
-var DefaultDurationBucketsSeconds = []float64{
-	0.001,
-	0.002,
-	0.005,
+var DefaultRpcDurationBucketsSeconds = []float64{
```
clemire, line 16: @sergekh the existing bucket setting looks like it was used for two types of metrics:
- things that measure rpc calls, and
- measurement of db transactions.

Each bucket has been updated to better reflect what I was seeing in omega on datadog for both types.

core/node/crypto/chain_txpool.go
```diff
@@ -164,15 +164,18 @@ func newPendingTransactionPool(
 		"txpool_processed", "Number of submitted transactions that are included in the chain",
 		"chain_id", "address", "status",
 	)
+
+	// 1, 3, 5, 7, 9s
```
clemire, line 168: @bas-vk does this look ok to you? This more than covers the range I've seen in omega over the past few days, with most tx inclusion <=5s and total inclusion <= 11s.

sergekh2 (approved):

mechanical-turk (approved):

PR #1737, opened 2024-12-06
feat: Stream miniblocks directly into RPC stream

rbhnt:


clemire (commented):
This looks great! Thanks so much for the refactor. :)

Can you make sure that there is at least one unit test running that uses the getStreamEx endpoint?

There is a `mediaStreamGetStreamEx` test in media.test.ts and of course there is `TestForwardingWithRetries` in go, but these tests were disabled due to a race and I'm also fine if you write your own separate test.

I think the race condition was in finalizing miniblock content for the stream, I can't recall exactly. However, they may actually work fine now, or if there is still a race in the implementation, you could get around this by waiting on the stream until all miniblocks are created and then calling getStreamEx to validate the contents of the response.

Or if you want feel free to investigate and fix the race! Could be follow-up work.

core/node/protocol/helpers.go
```diff
@@ -10,3 +17,21 @@ func (e *StreamEvent) GetStreamSettings() *StreamSettings {
...
+func (m *Miniblock) Scan(src any) error {
+	if src == nil {
+		return nil
+	}
+
+	switch src := src.(type) {
+	case []byte:
+		// Assume the data is serialized protobuf bytes
+		return proto.Unmarshal(src, m)
+	case string:
```
bas-vk, line 31: I'm slightly worried about implementing the `sql.Scanner` interface in a way that it supports multiple serialisation formats. This interface is used by database packages internally and can cause unexpected and hard to find problems. For example, if the result of `json.Marshal(miniblock)` is written it can't be read from the database because it tries to parse it as protobuf. If we implement the scan function we should also implement `database.Value` and be precise and pass miniblocks instead of byte slices.

For now I personally prefer that we keep the current convention and call `proto.Marshal` explicit as we do in other places and make a new issue to implement the scanner and value interfaces and migrate to passing miniblocks instead of byte slices.

core/node/rpc/stream_test.go
```diff
@@ -0,0 +1,54 @@
...
+	channelId, _ := alice.createChannel(spaceId)
+
+	for count := range 100 {
+		alice.say(channelId, fmt.Sprintf("hello from Alice %d", count))
+	}
+
+	time.Sleep(1 * time.Second)
+
+	stream := alice.getStream(channelId)
+	fmt.Println(len(stream.GetEvents())) // Prints 0
```
bas-vk, line 40: Use `require.Eventually` and `t.Logf` instead of sleep and println.

roman-aprio: Yea, this is just for debugging.

rbhnt: Yea, this is just for debugging

core/node/protocol/helpers.go
rbhnt: Agree, removed this part

core/node/rpc/stream_test.go
```diff
@@ -0,0 +1,49 @@
...
+	_ = alice.createUserStream()
+	spaceId, _ := alice.createSpace()
+	channelId, _ := alice.createChannel(spaceId)
+
+	for count := range 100 {
+		alice.say(channelId, fmt.Sprintf("hello from Alice %d", count))
+	}
+
+	require.Eventually(func() bool {
+		mbs := make([]*protocol.Miniblock, 0, 102)
```
clemire, line 37: super-nit: for the 102 number, please comment what the extra two blocks are

core/node/rpc/stream_test.go
```diff
@@ -0,0 +1,53 @@
...
+	_ = alice.createUserStream()
+	spaceId, _ := alice.createSpace()
+	channelId, _ := alice.createChannel(spaceId)
+
+	for count := range 100 {
+		alice.say(channelId, fmt.Sprintf("hello from Alice %d", count))
+	}
+
+	// Expected number of events is 102 because the first event is the channel creation event (inception),
+	// the second event is the joining the channel event (membership), and the rest are the messages.
```
clemire, line 37: 🙏🏻

clemire (approved):
I'm not sure about the changes @sergekh asked for, but LGTM!

(BTW if you cannot merge just make sure all outstanding comments on the PR are closed.)

core/node/storage/pg_stream_store.go
```diff
@@ -863,6 +865,68 @@ func (s *PostgresStreamStore) readMiniblocksTx(
...
+// Client should call Close() on the returned MiniblocksDataStream object when done.
+func (s *PostgresStreamStore) ReadMiniblocksByStream(
+	ctx context.Context,
+	streamId StreamId,
+	onEachMb func(blockdata []byte, seqNum int) error,
+) error {
+	return s.txRunnerWithUUIDCheck(
+		ctx,
+		"ReadMiniblocksByStream",
+		pgx.ReadWrite,
```
sergekh2, line 879:
```suggestion
		pgx.ReadOnly,
```

sergekh2 (approved):

PR #1739, opened 2024-12-06
[RIVER-1738] Update pg store to use advisory locking instead of comparing UUIDs per transaction

clemire:
This PR adds the usage of a session-wide `pg_advisory_lock`s to ensure unique read/write access to the database schema for this node while the node is up and available,  and removes the per-transaction check on the singlenodekey table, improving query performance of the running node at the cost of neglegible amount of additional startup latency to take the schema lock.

In order to maintain the session lock we acquire a connection from the pgx pool, take the lock on that connection, and continuously maintain the connection by polling it with pings. If a ping fails, the node will close the connection and attempt to acquire a new connection and re-establish the lock on the schema. If the node is unable to establish a healthy connection after about 20s of retries, or the node is unable to retake the schema lock, the node will exit.

As before, we still detect the entrance of other nodes vying for the same schema on the network by monitoring updates to the singlenodekey table. The node updates the singlenodekey table on startup before taking the schema lock, and new nodes will trigger old nodes to release their schema lock and exit. When a node detects that a schema lock is unavailable, it will use the existing StartupDelay config parameter to give the exiting node some time to clean up after itself.

Tests:
- an existing unit test confirms that the existing node exits whenever a new node starts, properly responding to the new entry in the singlenodekey table. The success of the test case implies that the first node is properly releasing the schema lock.
- manual local testing, using toxiproxy:
    - the node is able to recover the schema lock after temporary connection interruptions
    - the node will shut down after maximum retries to re-establish the connection and retake the schema lock if connectivity is not restored
    - the node will shut down if it is not able to retake the schema lock if network connectivity to pg is restored before it exhausts reconnection retries.


core/node/events/miniblock_producer.go
```diff
@@ -276,7 +276,10 @@ func (p *miniblockProducer) TestMakeMiniblock(
 
 		err = SleepWithContext(ctx, 10*time.Millisecond)
 		if err != nil {
-			return nil, err
+			return nil, AsRiverError(err, Err_INTERNAL).
```
clemire, line 279: As part of this PR I added some more tags to errors that popped up in testing in order to track down race conditions, and I left them in since they may be helpful in the future.

core/node/rpc/archiver_test.go
```diff
@@ -70,12 +78,10 @@ func createUserSettingsStreamsWithData(
 	streamIds := make([]StreamId, numStreams)
 	errChan := make(chan error, numStreams)
 
-	var wg sync.WaitGroup
-	wg.Add(numStreams)
+	wp := workerpool.New(10)
```
clemire, line 81: CI failures were occurring because the simulated backend would be overwhelmed in this test, so I added some rate limiting here by limiting the maximum number of concurrent go routines creating streams to 10 and this fixed it, although it would probably be helpful to test it tonight when CI seems to come under highest load to confirm.

core/node/storage/pg_stream_store.go
```diff
@@ -141,7 +142,186 @@ func NewPostgresStreamStore(
...
+				err = AsRiverError(fmt.Errorf("schema lock was not available"), Err_RESOURCE_EXHAUSTED).
+					Message("Lost connection and unable to re-acquire schema lock").
+					Func("maintainSchemaLock").
+					Tag("schema", s.schemaName).
+					Tag("lockId", lockId).
+					LogError(dlog.FromCtx(ctx))
+				s.exitSignal <- err
+			}
+
+			time.Sleep(1 * time.Second)
```
sergekh2, line 243: should this be a select on context cancelation/time.After? there is helper function base.SleepWithContext to do that.

core/node/storage/pg_stream_store.go
```diff
@@ -141,7 +142,186 @@ func NewPostgresStreamStore(
...
+				LogError(log)
+		}
+
+		if acquired {
+			log.Info("Schema lock acquired", "lockId", lockId, "nodeUUID", s.nodeUUID)
+			break
+		}
+
+		lockWasUnavailable = true
+		time.Sleep(1 * time.Second)
```
sergekh2, line 288: same

sergekh2 (approved):

PR #1755, opened 2024-12-08
Notifications service, remove channel setting when updating to unspecified

texuf:
I don’t have a good api to “unset” a channel setting.
The RemoveChannelSetting seems unused, wired it up to fire when i post unspecified for a channel.
Does this seem okay? Alternatively I could just keep the change in preferences.go that ignroes the unspecified channel setting.

### DIFF
--- a/core/node/rpc/notification_test.go
+++ b/core/node/rpc/notification_test.go
@@ -17,6 +17,9 @@ import (
 	"github.com/ethereum/go-ethereum/common"
 	eth_crypto "github.com/ethereum/go-ethereum/crypto"
 	"github.com/google/go-cmp/cmp"
+	payload2 "github.com/sideshow/apns2/payload"
+	"github.com/stretchr/testify/require"
+
 	"github.com/river-build/river/core/node/crypto"
 	"github.com/river-build/river/core/node/events"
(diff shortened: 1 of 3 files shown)


andreyz (approved):
I don't have a concept of unset, only reset to default, which you've essentially implemented

texuf:
> I don't have a concept of unset, only reset to default, which you've essentially implemented 

How do you reset to default? Do you reset everything? Reset a single space?

core/node/rpc/notification_test.go
```diff
@@ -1564,4 +1564,28 @@ func spaceChannelSettings(
...
+	test.req.NoError(err, "SetSpaceSettings failed")
+
+	// ensure that the settings are correct applied
+	settingsResp, err = test.notificationClient.GetSettings(ctx, request1)
+	test.req.NoError(err, "GetSettings failed")
+
+	settings = settingsResp.Msg
+
+	space = settings.GetSpace()[0]
+	test.req.Equal(1, len(space.Channels))
```
bas-vk, line 1590: Nitpick, add a check that ensures that the correct channel is dropped. So channel 1 is still in the channel list and channel2 is dropped.

bas-vk (approved):

PR #1757, opened 2024-12-08
Add transaction debugging and tracking to PostgresEventStore.

sergekh2:


clemire (approved):
Looks like this is being used to track what's in flight when a transaction is being retried? If so, please update the description.

PR #1759, opened 2024-12-09
Fix double start of chain monitor in tests

sergekh2:
Was started from both test harness and node start functions

### DIFF
--- a/core/xchain/server/server.go
+++ b/core/xchain/server/server.go
@@ -120,7 +120,6 @@ func New(
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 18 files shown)


core/xchain/server/server.go
```diff
@@ -187,7 +186,6 @@ func New(
 	}
```
clemire, line 186: Is the above setting being properly applied here now that chain monitor start happens on blockchain creation?

sergekh2: no, I'll fix it...

PR #1764, opened 2024-12-09
feat: GetMiniblocks limit of max blocks that can be fetched at once

rhbntt:
https://github.com/river-build/river/issues/1707


core/node/crypto/config.go
```diff
@@ -164,6 +166,8 @@ func DefaultOnChainSettings() *OnChainSettings {
 		StreamCacheExpiration:    5 * time.Minute,
 		StreamCachePollIntterval: 30 * time.Second,
 
+		MbsListLimit: 200,
```
texuf, line 169: can we commit this as disabled by default until we make the client changes?

core/node/crypto/config.go
```diff
@@ -95,6 +95,8 @@ type OnChainSettings struct {
 	StreamCacheExpiration    time.Duration `mapstructure:"stream.cacheExpirationMs"`
 	StreamCachePollIntterval time.Duration `mapstructure:"stream.cacheExpirationPollIntervalMs"`
 
+	MbsListLimit uint64 `mapstructure:"stream.miniblocksListLimit"`
```
sergekh2, line 98: Lets name it explicitly by API name:
```suggestion
	GetMiniblocksMaxPageSize uint64 `mapstructure:"stream.getMiniblocksMaxPageSize"`
```

core/node/rpc/get_miniblocks.go
```diff
@@ -14,14 +14,32 @@ func (s *Service) localGetMiniblocks(
 	req *connect.Request[GetMiniblocksRequest],
 	stream SyncStream,
 ) (*connect.Response[GetMiniblocksResponse], error) {
-	miniblocks, terminus, err := stream.GetMiniblocks(ctx, req.Msg.FromInclusive, req.Msg.ToExclusive)
+	toExclusive := req.Msg.ToExclusive
```
sergekh2, line 17: please return error here if toExlusive <= fromInlcusive (otherwise math below can be incorrect)

sergekh2 (approved):

PR #1770, opened 2024-12-10
Fix reading history on start in xchain 

sergekh2:
I broke it in #1759


core/xchain/server/server.go
```diff
@@ -313,7 +325,7 @@ func (x *xchain) Run(ctx context.Context) {
 
 	// register callback for Base EntitlementCheckRequested events
 	x.baseChain.ChainMonitor.OnContractWithTopicsEvent(
-		x.baseChain.InitialBlockNum,
+		x.baseChainStartBlock,
```
clemire, line 328: I don't know if we have any test cases for this, so I'm not sure this will break anything, but based on a cursory reading of the code it looks like there still might be a bug here.

If we start the chain monitor with a certain initial block number and then ask it to run a callback with a starting block that is << the initial block number, do we ever expect to see any events from those blocks that occur before the initial block number? I think that block number argument is meant for blocks that occur after the chain monitor is running (not exactly sure though.)

PR #1775, opened 2024-12-10
Update publish-to-npm script to include all addresses from the deployments directlry

texuf:
The client uses the diamond addresses which never change, but sometimes (like for rewards distribution) we need the addresses in the sdk. This keeps these folders in sync daily without needing to worry about it.

### DIFF
--- a/scripts/publish-to-npm.sh
+++ b/scripts/publish-to-npm.sh
@@ -45,6 +45,7 @@ if [ $exit_status_yarn -ne 0 ]; then
     exit 1
 fi
 
+# generate docs
 yarn workspace @river-build/react-sdk gen
 exit_status_docgen=$?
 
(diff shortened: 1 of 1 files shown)


scripts/publish-to-npm.sh
```diff
@@ -56,6 +57,14 @@ fi
 git add packages/docs/
 git commit -m "docs for version ${VERSION_PREFIX}"
 
+# copy contracts
+rm -rf packages/generated/deployments
+cp -r contracts/deployments packages/generated/deployments
```
miguel-nascimento, line 62: how about `packages/generated/dev` for abis?

texuf: that should be part of the normal ci build - are those out of sync?

PR #1776, opened 2024-12-10
Delete all facet addresses from the generated package

texuf:
We don’t and shouldn’t use these addresses, everything should be accessed through the diamonds. Keeping them up to date is a pain, so lets not have them at all.

Wrote a script that copies everything over and removes *facet.json files


jterzis (changes requested):
Why are you deleting `space.json` corresponding to space diamond ? This will break diff reports, please don't delete this.

texuf:
> Why are you deleting `space.json` corresponding to space diamond ? This will break diff reports, please don't delete this.

space.json is "an example space deployment" spaceFactory.json and spaceOwner are the diamonds

jterzis:
> > Why are you deleting `space.json` corresponding to space diamond ? This will break diff reports, please don't delete this.
> 
> space.json is "an example space deployment" spaceFactory.json and spaceOwner are the diamonds

It's an example but also stores updated space facets which are canonical. We need this diamond to continue to exist so we can determine whether space facets differ between environments. If you remove the address, we won't be able to reference the corresponding deployed Space facets on alpha, gamma, omega.

texuf:
> > > Why are you deleting `space.json` corresponding to space diamond ? This will break diff reports, please don't delete this.
> > 
> > 
> > space.json is "an example space deployment" spaceFactory.json and spaceOwner are the diamonds
> 
> It's an example but also stores updated space facets which are canonical. We need this diamond to continue to exist so we can determine whether space facets differ between environments. If you remove the address, we won't be able to reference the corresponding deployed Space facets on alpha, gamma, omega.

Thanks for the explanation, I'll add them back. @giuseppecrj ^^

texuf:
> > > > Why are you deleting `space.json` corresponding to space diamond ? This will break diff reports, please don't delete this.
> > > 
> > > 
> > > space.json is "an example space deployment" spaceFactory.json and spaceOwner are the diamonds
> > 
> > 
> > It's an example but also stores updated space facets which are canonical. We need this diamond to continue to exist so we can determine whether space facets differ between environments. If you remove the address, we won't be able to reference the corresponding deployed Space facets on alpha, gamma, omega.
> 
> Thanks for the explanation, I'll add them back. @giuseppecrj ^^

But also - @jterzis - they are still, and will always be, in contracts/deployments... i'm only deleting them from generated/deployments, the package we publish to npm. Do we need them in both places? (okay if the answer is yes)

texuf:
> > > > > Why are you deleting `space.json` corresponding to space diamond ? This will break diff reports, please don't delete this.
> > > > 
> > > > 
> > > > space.json is "an example space deployment" spaceFactory.json and spaceOwner are the diamonds
> > > 
> > > 
> > > It's an example but also stores updated space facets which are canonical. We need this diamond to continue to exist so we can determine whether space facets differ between environments. If you remove the address, we won't be able to reference the corresponding deployed Space facets on alpha, gamma, omega.
> > 
> > 
> > Thanks for the explanation, I'll add them back. @giuseppecrj ^^
> 
> But also - @jterzis - they are still, and will always be, in contracts/deployments... i'm only deleting them from generated/deployments, the package we publish to npm. Do we need them in both places? (okay if the answer is yes)

Ah one more thing, sorry for the never ending replies. Looks like you added a `Drop` class that references a `riverAirdrop` address... but you never hand copied those addresses into the generated/deployments folder so that will never work. @giuseppecrj here's one more. (looks like I approved your initial PR, sorry for not giving you better feedback John).

jterzis:
> > > > > Why are you deleting `space.json` corresponding to space diamond ? This will break diff reports, please don't delete this.
> > > > 
> > > > 
> > > > space.json is "an example space deployment" spaceFactory.json and spaceOwner are the diamonds
> > > 
> > > 
> > > It's an example but also stores updated space facets which are canonical. We need this diamond to continue to exist so we can determine whether space facets differ between environments. If you remove the address, we won't be able to reference the corresponding deployed Space facets on alpha, gamma, omega.
> > 
> > 
> > Thanks for the explanation, I'll add them back. @giuseppecrj ^^
> 
> But also - @jterzis - they are still, and will always be, in contracts/deployments... i'm only deleting them from generated/deployments, the package we publish to npm. Do we need them in both places? (okay if the answer is yes)

@texuf we read them from contracts/deployments for diffing so it is okay I think to remove them from packages/generated.

texuf:
updated this pr to only remove the facet.json files, also include a script to do the copying

PR #1783, opened 2024-12-11
Add logging to quorum sends, refactor a bit

sergekh2:


bas-vk (approved):
It looks ok but there is a test failing. Not sure why.

PR #1787, opened 2024-12-11
Add Blockchain Transaction Event

texuf:
What's going on?
A user can upload any transaction receipt to their user stream, the transaction must
- exist on chain 
- have at least 1 confirmation
- blocknumber, log count, log bytes, topic count, topic bytes, log data bytes, and the to and from addresses must match
- the receipt must be from a wallet that is linked to the user's root key 
- not already uploaded

If a user uploads a Tip, the node will
- verify metadata about the tip (still todo)
- forward a Tip_Received event to the recipient's user stream
- forward the original tip transaction to the channel stream

This will allow us to render tips in real time in the app, and receive push notifications just like we do for emojis.

New types
- blockchain transaction
- blockchain transaction receipt
- blockchain transaction kind
- two user payloads
- one member payload (generic across all streams)

New tech
- i updated the can add event return values to include a "verifyReceipt" param
- i added a new chain auth type that just checks to see if a wallet is linked to the principal.

Still todo in a follow up
- tip receipt validation. I need to parse out the values from the receipt logs and make sure they match what's in the event metadata
- user and channel stream message tests. I have stubs written, but not there yet.
- snapshotting relevant data


texuf:
very much still in progress, will continue plugging away tomorrow.

miguel-nascimento:
that's cool! 🔥

packages/sdk/src/client.ts
```diff
@@ -1877,6 +1882,53 @@ export class Client
         )
     }
 
+    // upload transactions made on the base chain
+    async addTransaction(
+        chainId: number,
+        receipt: ContractReceipt,
```
salzbrenner, line 1888: I wonder if we can keep ethers out of the sdk? Or is it in already elsewhere? I still have hope and plans that we can refactor to viem or upgrade. Based on what's passed to the stream, maybe custom type

```
transactionHash: string`
to: string
from: string
logs: {
  address: string
  data: string
  topics: string[]
}[]
```

salzbrenner:
looking good so far

packages/sdk/src/client.ts
texuf: it is all over, but we can change this!

core/node/auth/auth_impl.go
```diff
@@ -726,7 +861,8 @@ func (ca *chainAuth) getLinkedWallets(
 	// and user scrubs, all of which request the Read permission.
 	// Note: space joins seem to request Read on the space, but they should probably actually
 	// be sending chain auth args with kind set to chainAuthKindIsSpaceMember.
-	if args.permission == PermissionRead || args.kind == chainAuthKindIsSpaceMember {
+	if args.permission == PermissionRead || args.kind == chainAuthKindIsSpaceMember ||
```
clemire, line 864: Can you update the above comment?

core/node/rules/can_add_event.go
```diff
@@ -540,9 +586,209 @@ func (params *aeParams) creatorIsMember() (bool, error) {
...
+			"unknown transaction type",
+			"transactionType",
+			ru.transaction.Kind,
+		)
+	}
+}
+
+func (ru *aeReceivedBlockchainTransactionRules) parentEventForReceivedBlockchainTransaction() (*DerivedEvent, error) {
+	transaction := ru.receivedTransaction.Transaction
+	streamId, err := shared.StreamIdFromBytes(transaction.StreamId)
```
clemire, line 707: if transaction is nil according to the check below shouldn't this line panic? Looks like the order of these two checks should be switched.

core/node/rules/can_add_event.go
```diff
@@ -540,9 +586,209 @@ func (params *aeParams) creatorIsMember() (bool, error) {
...
+	}
+	if transaction == nil {
+		return nil, RiverError(Err_INVALID_ARGUMENT, "transaction is nil")
+	}
+
+	switch ru.receivedTransaction.Kind {
+	case ReceivedBlockchainTransactionKind_RECEIVED_BLOCKCHAIN_TRANSACTION_KIND_UNSPECIFIED:
+		return nil, RiverError(Err_INVALID_ARGUMENT, "transaction kind is unspecified")
+	case ReceivedBlockchainTransactionKind_RECEIVED_BLOCKCHAIN_TRANSACTION_KIND_TIP:
+		// forward the tip to the channel stream of the fromUserAddress
```
clemire, line 719: comment nit

`// forward the tip to the channel stream, attributed to fromUserAddress` maybe?

packages/sdk/src/tests/multi/transactions.test.ts
```diff
@@ -0,0 +1,214 @@
...
+    afterEach(() => {
+        // for speed we share a receipt between tests, don't modify it
+        expect(dummyReceiptCopy).toEqual(dummyReceipt)
+    })
+
+    test('addEvent', async () => {
+        // a user should be able to upload a transaction that
+        // is from their account or one of their linked accounts and
+        // is valid on chain
+        // add the transaction to the river
```
clemire, line 86:
```suggestion
        // add the transaction to the river chain
```

packages/sdk/src/tests/multi/transactions.test.ts
```diff
@@ -0,0 +1,214 @@
...
+    })
+
+    test('cantAddEventTwice', async () => {
+        const transaction = await bobIdentity.web3Provider.mintMockNFT(riverConfig.base.chainConfig)
+        const receipt = await transaction.wait(2)
+        await expect(
+            bob.riverConnection.client!.addTransaction(chainId, receipt),
+        ).resolves.not.toThrow()
+        // can't add the same transaction twice
+        await expect(bob.riverConnection.client!.addTransaction(chainId, receipt)).rejects.toThrow()
```
clemire, line 101: A nice-to-have for these would be a regex matching the thrown error, `/duplicate/` or something, as an argument to toThrow

clemire (approved):

core/node/rpc/add_event.go
```diff
@@ -77,11 +77,11 @@ func (s *Service) addParsedEvent(
...
 
-	if len(chainAuthArgsList) > 0 {
+	if len(verifications.OneOfChainAuths) > 0 {
 		isEntitled := false
 		var err error
 		// Determine if any chainAuthArgs grant entitlement
-		for _, chainAuthArgs := range chainAuthArgsList {
+		for _, chainAuthArgs := range verifications.OneOfChainAuths {
 			isEntitled, err = s.chainAuth.IsEntitled(ctx, s.config, chainAuthArgs)
 			if err != nil {
```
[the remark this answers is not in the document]
texuf: we want to return errors so that clients can retry - but since it's a oneOf, maybe one succeeds and one errors, in which case we could just move forward? It's a good note.

core/node/rules/can_add_event.go
```diff
@@ -540,9 +586,209 @@ func (params *aeParams) creatorIsMember() (bool, error) {
...
+			"unknown transaction kind",
+			"kind",
+			ru.memberTransaction.Transaction.Kind,
+		)
+	}
+}
+
+func (ru *aeMemberBlockchainTransactionRules) validMemberBlockchainTransaction_IsUnique() (bool, error) {
+	// loop over all events in the view, check if the transaction is already in the view
+	userStreamView := ru.params.streamView.(events.JoinableStreamView)
```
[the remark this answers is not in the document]
texuf: yes, copy pasta, i'll update the name thanks.

core/node/rules/can_add_event.go
texuf: good point

packages/sdk/src/tests/multi/transactions.test.ts
texuf: ah yes i missed this one

core/node/rules/can_add_event.go
texuf: `// forward the tip to the stream as a member event, preserving the original sender as the from address`

core/node/auth/auth_impl.go
```diff
@@ -868,6 +1006,17 @@ func (ca *chainAuth) checkEntitlement(
...
+	if args.kind == chainAuthKindIsWalletLinked {
+		for _, wallet := range wallets {
+			if wallet == args.walletAddress {
+				return boolCacheResult(true), nil
+			}
+		}
+		return boolCacheResult(false), nil
+	}
+
+	// populate the cache key for the linked wallets
```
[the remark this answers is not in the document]
texuf: I don't love this because it mutates the args passed into the function. It concatenates the wallets into a string that is used as a cache key for several of the following operations. I guess at some point the string is split and the wallets are then used individually. I can take the comment out though.

PR #1790, opened 2024-12-11
Updated SDK to paginate miniblocks

rhbntt:
GetMiniblocks endpoint has a limit of max items that can be returned per request (https://github.com/river-build/river/pull/1764) - some sort of pagination. This PR updates the client SDK to properly handle the behaviour of the updated endpoint.

### DIFF
--- a/packages/sdk/src/makeStreamRpcClient.ts
+++ b/packages/sdk/src/makeStreamRpcClient.ts
@@ -85,7 +85,7 @@ export async function getMiniblocks(
     let currentFromInclusive = fromInclusive
     let reachedTerminus = false
 
-    while (currentFromInclusive <= toExclusive) {
+    while (currentFromInclusive < toExclusive) {
         const { miniblocks, terminus, nextFromInclusive } = await fetchMiniblocksFromRpc(
             client,
             streamId,
--- a/core/node/rpc/get_miniblocks.go
+++ b/core/node/rpc/get_miniblocks.go
@@ -17,7 +17,7 @@ func (s *Service) localGetMiniblocks(
 ) (*connect.Response[GetMiniblocksResponse], error) {
 	toExclusive := req.Msg.ToExclusive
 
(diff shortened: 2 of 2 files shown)


packages/sdk/src/makeStreamRpcClient.ts
```diff
@@ -70,3 +73,68 @@ export function getMaxTimeoutMs(opts: StreamRpcClientOptions): number {
...
+    streamId: string | Uint8Array,
+    fromInclusive: bigint,
+    toExclusive: bigint,
+    unpackEnvelopeOpts: UnpackEnvelopeOpts | undefined,
+): Promise<{ miniblocks: ParsedMiniblock[]; terminus: boolean }> {
+    const allMiniblocks: ParsedMiniblock[] = []
+    let currentFromInclusive = fromInclusive
+    let reachedTerminus = false
+
+    while (currentFromInclusive < toExclusive && !reachedTerminus) {
```
texuf, line 88: 1. shouldn't this be `currentFromInclusive <= toExclusive`? I would expect the next
2. you can't check both the condition and `reactedTerminus` if you're returning left to right

stream = [0,1,2,3,4,5]
maxMBs = 3
fetchMiniblocksFromRpc(0,6) will return { blocks: [0,1,2], reachedTerminus: true, fromInclusive: 0 }
// and the loop will exit

// for a trimmed stream
trimmedStream = [3,4,5,6,7,8]
maxMBs = 3
fetchMiniblocksFromRpc(0,6) will return { blocks: [3,4,5], reachedTerminus: true, fromInclusive: 3 }
// and the loop will exit

texuf (approved):

packages/sdk/src/makeStreamRpcClient.ts
texuf: 🤦 yeah #1 is right as is, thanks for fixing #2

PR #1795, opened 2024-12-11
proto: move to vitest

miguel-nascimento:


packages/proto/package.json
```diff
@@ -26,14 +26,11 @@
...
-        "@types/jest": "^29.5.3",
         "@types/node": "^20.5.0",
-        "jest": "^29.6.2",
-        "jest-extended": "^4.0.1",
         "lodash.debounce": "^4.0.8",
-        "ts-jest": "^29.1.1",
         "ts-node": "^10.9.1",
-        "typescript": "^5.1.6"
+        "typescript": "^5.1.6",
+        "vitest": "^2.1.8"
```
texuf, line 33: looks like everything else is on vitest 2.1.5? - can we use the same?

miguel-nascimento: good catch!

miguel-nascimento: ill bump all packages later to 2.1.8

PR #1805, opened 2024-12-12
feat(dapp): add river points entry

ttillberg:


packages/web3/src/IStaticContractsInfo.ts
```diff
@@ -9,6 +9,7 @@ export interface BaseChainConfig {
         spaceOwner: Address
         baseRegistry: Address
         riverAirdrop?: Address
+        riverPointsFacet?: Address
```
texuf, line 12: we don't put facet addresses here. @giuseppecrj or @shuhuiluo  is river points on a new diamond? What goes here?

ttillberg: thanks for checking; I'm unfamiliar with this area.

giuseppecrj: river points is on riverAirdrop

texuf: the riverAirdrop addresses were never put in the right folder. I'm going to re-create my pr that copies everything over and get this fixed up today.

ttillberg: Thank you – as far as I understand @texuf is taking a look. I doesn't look like riverAirdrop is actually hooked up to the dApp yet.

texuf:
you can close in favor of https://github.com/river-build/river/pull/1811

PR #1808, opened 2024-12-12
add protocol toggling of MLS on / off on specific streams

erikolsson:
I was flipflopping a bit here — do we want the toggled property to be something more granular than a boolean?
Ideas:
`int32 default_encryption_algorithm`
`string default_encryption_algorithm`

I kind of think a bool is fine but open to suggestions!

### DIFF
--- a/protocol/protocol.proto
+++ b/protocol/protocol.proto
@@ -175,8 +175,8 @@ message MemberPayload {
         
         repeated Member joined = 1;
         repeated SnappedPin pins = 2;
-        bool mlsEnabled = 3;
-        Mls mls = 4;
+        Mls mls = 3;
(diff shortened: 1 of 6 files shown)


jakzale (approved):

protocol/protocol.proto
```diff
@@ -221,6 +222,7 @@ message MemberPayload {
         Nft nft = 7;
         Pin pin = 8;
         Unpin unpin = 9;
+        bool mlsEnabled = 10;
```
texuf, line 225: variable should be snake cased in the protobuf def (because python)
I would have gone with `optional string encryption_algorithm` to give more flexibility in the future.

texuf (approved):

PR #1819, opened 2024-12-13
SpaceDapp: change tip to receive tokenId instead of receiver address

salzbrenner:
I thought it made more sense to be able to pass receiver here. But after writing some tests, I realized the membership nft could be in any linked wallet - and that is where the tip is going to go.

It's confusing to have a parameter called receiver, then internally fetch linked wallets, and have the tip go potentially to another address. So instead callers can pass the tokenId using getTokenIdOfOwner

### DIFF
--- a/packages/web3/src/v3/SpaceDapp.ts
+++ b/packages/web3/src/v3/SpaceDapp.ts
@@ -1646,10 +1646,39 @@ export class SpaceDapp implements ISpaceDapp {
         return space.Membership.listenForMembershipToken(receiver, abortController)
     }
 
+    /**
+     * Get the token id for the owner
+     * Returns the first token id matched from the linked wallets of the owner
+     * @param spaceId - The space id
+     * @param owner - The owner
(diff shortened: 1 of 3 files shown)


packages/web3/src/v3/SpaceDapp.ts
```diff
@@ -1646,26 +1646,52 @@ export class SpaceDapp implements ISpaceDapp {
...
+        const { spaceId, tokenId, currency, amount, messageId, channelId } = args
         const space = this.getSpace(spaceId)
         if (!space) {
             throw new Error(`Space with spaceId "${spaceId}" is not found.`)
         }
-        const tokenId = (await space.getTokenIdsOfOwner(receiver))[0]
-        if (!tokenId) {
-            throw new Error(`TokenId for receiver "${receiver}" is not found.`)
-        }
+
```
texuf, line 1694: Weird but okay

texuf (approved):

Issue #1825, opened 2024-12-14
Archiver metrics, logging cleanup


clemire:
### Description

Archiver should emit metrics so that we can track the state of the archiver. We also should clean up log emission so we can re-capture those logs in datadog.

PR #1826, opened 2024-12-14
[RIVER-1825] Archiver metrics and log reduction

clemire:
#1825 

Example local metrics output for stats when I point a node at omega:

```
# HELP river_archive_stats_failed_ops_count Total failed ArchiveStream executions
# TYPE river_archive_stats_failed_ops_count gauge
river_archive_stats_failed_ops_count 0
# HELP river_archive_stats_miniblocks_processed Total miniblocks downloaded and stored since the last boot
# TYPE river_archive_stats_miniblocks_processed gauge
river_archive_stats_miniblocks_processed 169863
# HELP river_archive_stats_new_stream_allocated Total streams allocated in response to detected stream allocation events
# TYPE river_archive_stats_new_stream_allocated gauge
river_archive_stats_new_stream_allocated 0
# HELP river_archive_stats_stream_last_miniblock_updated Total miniblock update events
# TYPE river_archive_stats_stream_last_miniblock_updated gauge
river_archive_stats_stream_last_miniblock_updated 0
# HELP river_archive_stats_stream_placement_updated Total stream placement changes
# TYPE river_archive_stats_stream_placement_updated gauge
river_archive_stats_stream_placement_updated 0
# HELP river_archive_stats_streams_created Total streams allocated on disk by the archiver since the last boot
# TYPE river_archive_stats_streams_created gauge
river_archive_stats_streams_created 1612
# HELP river_archive_stats_streams_examined Total streams monitored by the archiver
# TYPE river_archive_stats_streams_examined gauge
river_archive_stats_streams_examined 110218
# HELP river_archive_stats_streams_up_to_date Total ArchiveStream executions that did not see stream updates
# TYPE river_archive_stats_streams_up_to_date gauge
river_archive_stats_streams_up_to_date 8580
# HELP river_archive_stats_success_ops_count Total successful ArchiveStream executions
# TYPE river_archive_stats_success_ops_count gauge
river_archive_stats_success_ops_count 10198
```


core/node/rpc/archiver.go
```diff
@@ -10,24 +11,63 @@ import (
...
 	"github.com/river-build/river/core/node/registries"
 	. "github.com/river-build/river/core/node/shared"
 	"github.com/river-build/river/core/node/storage"
 )
 
+// nodeUpdateGracePeriod describes the maximum delay we would expect to see
+// when a stream's miniblocks updates in the registry before we consider a
+// node behind of it does not have the latest miniblock.
+// Two blocks plus change
+var nodeUpdateGracePeriod = 5 * time.Second
```
clemire, line 33: Perhaps these should be on-chain config? So they are common across the network?

core/node/rpc/archiver.go
```diff
@@ -101,6 +146,82 @@ func NewArchiver(
 	return a
 }
 
+func (a *Archiver) setupStatisticsMetrics(factory infra.MetricsFactory) {
```
clemire, line 149: Emit internally tracked statistics as gauge metrics, a la what we do for postgres

core/node/rpc/archiver.go
```diff
@@ -101,6 +144,82 @@ func NewArchiver(
...
+		},
+		{
+			"stream_last_miniblock_updated",
+			"Total miniblock update events",
+			func() float64 { return float64(a.streamLastMiniblockUpdated.Load()) },
+		},
+	}
+
+	for _, metric := range statsMetrics {
+		factory.NewGaugeFunc(
```
sergekh2, line 201: but why this two-step process? just call factory.NewGaugeFunc directly instead of creating array and then iterating. it's same of amount of text, but more readable and to the point. 

(It's also fine to replace these atomics with metrics completely, i.e. inc/dec metrics directly)

also printStats, etc. can be removed - once they are published as metrics, there is no need to log them anymore.

clemire: I think I just copied the code from postgres and didn't fully realize it didn't need to be so complex.

As discussed in towns, I'll look into using a test metrics collector so we can check the output of the metrics collection for test cases.

clemire: eh, I think it's just easier to use the atomics.

texuf (approved):

PR #1828, opened 2024-12-14
Kerem/river ci optimization

mechanical-turk:
- fix turbo cache directory
- fix turbo cache keys
- aggressive caching for most turbo scripts
- remove duplicate test run
- each job now has its own separate cache key, but they will all fall back to the common_ci key, and then to the main branch cache if all else fails

Next steps: don't boot nodes unless we need to test them - i.e if turbo's cache for the test is not stale, there's no need to start the nodes only to realize that the tests don't need to get run. 

Final step: turbofy everything, including "build-contract-types". no need to rerun things unless the source has changed. we can also turbofy golang ci scripts.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -244,6 +244,14 @@ jobs:
             - name: Install node dependencies
               run: yarn install --immutable
 
+            - name: Read Turbo cache
+              uses: actions/cache/restore@v4
+              with:
+                  path: .turbo
+                  key: ${{ runner.os }}-turbo-${{ github.ref_name}}
+                  restore-keys: |
+                      ${{ runner.os }}-turbo-main
+
             - name: Build all packages
               run: yarn build
 
@@ -260,14 +268,6 @@ jobs:
 
             - name: Print Yarn package versions
               run: yarn info --name-only --all
-
-            - name: Read Turbo cache
-              uses: actions/cache/restore@v4
-              with:
-                  path: .turbo
-                  key: ${{ runner.os }}-turbo-${{ github.ref_name}}
-                  restore-keys: |
-                      ${{ runner.os }}-turbo-main
 
             - name: create db
               run: psql -h localhost -p 5433 -U postgres -f scripts/sql/create_db.sql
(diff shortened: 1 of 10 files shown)


texuf (approved):

foundry.toml
```diff
@@ -14,7 +14,7 @@ script = "contracts/scripts" # The path to the contract scripts relative to the
 solc_version = "0.8.24" # Uses a single Solidity compiler version for the project
 src = "contracts" # The path to the contract sources relative to the root of the project.
 test = "contracts/test" # The path to the test contract sources relative to the root of the project.
-cache_path = "contracts/cache" # The path to the cache, relative to the root of the project.
+cache_path = "contracts/contracts-cache" # The path to the cache, relative to the root of the project.
```
mechanical-turk, line 17: turbo needs to be aware of this directory, but once include a directory in turbo cache, you include it on every package in the workspace. and we don't want the `cache` directory of every package in turbo cache, so i'm giving this one a more specific name

miguel-nascimento (approved):

.github/workflows/ci.yml
```diff
@@ -289,17 +290,15 @@ jobs:
...
+              run: yarn turbo run test:ci:multi:ne --filter=@river-build/sdk
 
             - name: Run Stream Metadata Nodes
               run: |
                   yarn workspace @river-build/stream-metadata dev:local_multi_ne &
                   yarn wait-on http://localhost:3003/health --timeout=120000 --i=5000 --verbose
 
             - name: Run Stream Metadata Tests (without entitlements)
-              working-directory: packages/stream-metadata
-              run: yarn test:integration
```
mechanical-turk, line 302: revert this ^. it's necessary to rerun this test

.github/workflows/ci.yml
```diff
@@ -559,11 +560,8 @@ jobs:
               run: just RUN_ENV=multi config-and-start
               working-directory: core
 
-            - name: Build & Test (with entitlements)
-              run: yarn csb:turbo-no-sdk
```
mechanical-turk, line 563: note to self: bring this back, this is necessary

contracts/.gitignore
```diff
@@ -1,5 +1,6 @@
 /node_modules
 cache/
+contracts-cache/
```
mechanical-turk, line 3: undo this

PR #1836, opened 2024-12-16
[DRAFT] Update tool to ignore minipools and miniblock_candidates for archiver

clemire:


clemire:
Used to migrate our omega archiver nodes, this PR doesn't need to be committed.

PR #1848, opened 2024-12-17
add protocol slot for mls events

erikolsson:
identical to #1818 but i forgot to set up signed commits on my new computer and this is faster than trying to repair the old PR


texuf:
👍

PR #1849, opened 2024-12-17
updates to migration token contract

giuseppecrj:


jterzis (approved):

contracts/src/tokens/migration/TokenMigration.sol
```diff
@@ -28,29 +30,34 @@ contract TokenMigrationFacet is
...
+    _validateTokens(oldToken, newToken);
+
     TokenMigrationStorage.Layout storage ds = TokenMigrationStorage.layout();
     (ds.oldToken, ds.newToken) = (oldToken, newToken);
     _pause();
   }
 
   /// @inheritdoc ITokenMigration
-  function migrate(address account) external whenNotPaused {
+  function migrate(address account) external whenNotPaused nonReentrant {
```
shuhuiluo, line 41: There's no arbitrary external call causing reentrancy.

shuhuiluo (approved):

contracts/src/tokens/migration/TokenMigration.sol
giuseppecrj: fixed in latest push

PR #1853, opened 2024-12-17
[DEPRECATED][BUGFIX] Move sleep to implement a 1s delay between pings for schema lock connection

clemire:
Edit: This PR is built off of the schema lock feature, which has since been reverted. The replication bugs unearthed by this PR need to be fixed, then the schema lock reversion needs to be reverted again, and this can be applied on top of that PR.

The sleep was incorrectly placed previously, resulting in non-stop pings on the schema lock connection, which is unnecessary. This PR moves the sleep to the correct location so we ping once very second.

### DIFF
--- a/core/node/storage/pg_stream_store.go
+++ b/core/node/storage/pg_stream_store.go
@@ -163,17 +163,10 @@ func (s *PostgresStreamStore) maintainSchemaLock(
 
 	lockId := s.computeLockIdFromSchema()
 
-	count := 0
 	for {
 		// Check for connection health with a ping. Also, maintain the connection in the
 		// case of idle timeouts.
 		err := conn.Ping(ctx)
-		count++
-
-		if count%100 == 0 {
-			log.Debug("DB Ping!", "error", err)
-		}
-
 		if err != nil {
 			// We expect cancellation only on node shutdown. In this case,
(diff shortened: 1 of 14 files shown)


sergekh2 (approved):

core/node/storage/pg_stream_store.go
```diff
@@ -239,10 +232,10 @@ func (s *PostgresStreamStore) maintainSchemaLock(
 					LogError(dlog.FromCtx(ctx))
 				s.exitSignal <- err
 			}
-
-			if err = SleepWithContext(ctx, 1*time.Second); err != nil {
-				return
-			}
+		}
+		// Wait 1s between db pings.
```
sergekh2, line 236: Comment is incorrect? But anyway it's redundant: it's pretty obvious from the code that this is sleep between pings.

sergekh2 (approved):

core/node/storage/pg_stream_store.go
clemire: I'll take the comment out. In the meantime this PR is in a weird state because I'm trying to solicit more info about how the race condition bug is occurring and so it's regressing away from it's final state until I fix the issue causing the test flake.

PR #1857, opened 2024-12-18
[RIVER-1858] Node and corrupt stream metrics

clemire:
#1858 

Consider a stream corrupt if it was unavailable for greater than 50 consecutive update attempts. The archiver rotates between nodes whenever an update fails, so in the future when replication is deployed we can reduce this number to something like 5 since at that point we will have tried every node. The 50 is in there to accomodate stream unavailability while nodes are updating.

Once a stream is corrupt, it will not attempt to retry updating itself. New miniblocks on the network will trigger the stream, however, to update itself... but it will no longer retry on failures. This is an in-memory setting and will be reset the next time the node starts.


core/node/rpc/archiver.go
```diff
@@ -238,12 +271,11 @@ func (a *Archiver) addNewStream(
 	a.streamsExamined.Add(1)
 }
 
+// Consider this node behind for this stream and advance the node pointer
```
clemire, line 274: update this comment

core/node/rpc/archiver.go
```diff
@@ -219,6 +227,31 @@ func (a *Archiver) setupStatisticsMetrics(factory infra.MetricsFactory) {
 	)
 }
 
+// getStaleStreams iterates over all streams in the in-memory cache and collects stale
```
clemire, line 230: Fix method docs

core/node/rpc/archiver.go
```diff
@@ -403,14 +468,34 @@ func (a *Archiver) ArchiveStream(ctx context.Context, stream *ArchiveStream) err
 		a.miniblocksProcessed.Add(uint64(len(serialized)))
 	}
 
-	// All blocks processed, mark stream as current
-	stream.stale.Store(false)
-
 	return nil
 }
 
+func (a *Archiver) emitPeriodicCorruptStreamReport(ctx context.Context) {
```
clemire, line 474: the debug endpoints will not be available until Kerem gets back in the office and sets it up in AWS, so for now let's periodically print a report to logs

sergekh2 (approved):

PR #1879, opened 2024-12-19
add mls-tools unix socket util

erikolsson:
MLS Tools

A ~command line util~ grpc service for performing hyper specific operations requiring https://github.com/awslabs/mls-rs.
The util reads from stdin and the response is parsed by the caller from stdout.

For now, there are two validation steps, that mainly focus on mls state consistency. We will later be able to add additional checks that focus on security.

- `InitialGroupInfoRequest` validates the group info message for MLS group inception. among other things, check that:
  - ensure that the group info message contains the `ExternalPubExt` (required for external joins)
  - epoch == 0
  - external group snapshot contains tree
- `ExternalJoinRequest` validates an external join. Among other things:
  - roll up external group snapshot and process (apply) non-rolled-up commits
  - `proposedGroupInfoMessage.epoch` == `currentEpoch + 1`
  - `proposedCommit.epoch == currentEpoch`
  - we are still missing checks here, but it's a start.

The rust code still needs more tests — will be addressed as we actually start hooking this up.

The util will also be used for snapshotting the MLS state.

### DIFF
--- a/mls/mls-tools/crates/mls/src/lib.rs
+++ b/mls/mls-tools/crates/mls/src/lib.rs
@@ -69,6 +69,12 @@ pub fn validate_initial_group_info_request(request: InitialGroupInfoRequest) -> 
         };
     }
 
+    if group_info.group_context().group_id() != external_group.group_context().group_id() {
+        return InitialGroupInfoResponse {
+            result: ValidationResult::InvalidGroupInfo.into(),
+        };
+    }
+
     match group_info.extensions().get_as::<ExternalPubExt>() {
         Ok(extensions) => {
             match extensions {


texuf:
> Some thoughts:
> 
> - Am I overengineering the justfile? Can't all the nodes just share the same instance of mls_service when running locally? That way we can just `killall mls_service` and be done with it. Also simplifies the the call site (mls_service.go)
> 
> - I'm fine hardcoding the socket address `"unix:/tmp/mls_service"` in `core/node/mls_service/mls_service.go` for now, there should be no need to run multiple instances of MLS services
> 
> - I'm not sure how I should make sure that this runs in the node — where should I hook in supervisord?
> 
> 
> 
> @bas-vk @texuf @sergekh2 

I'm surprised that it's a service. I expected some kind of static library. Is there state?

texuf:
> > Some thoughts:
> > 
> > - Am I overengineering the justfile? Can't all the nodes just share the same instance of mls_service when running locally? That way we can just `killall mls_service` and be done with it. Also simplifies the the call site (mls_service.go)
> > 
> > - I'm fine hardcoding the socket address `"unix:/tmp/mls_service"` in `core/node/mls_service/mls_service.go` for now, there should be no need to run multiple instances of MLS services
> > 
> > - I'm not sure how I should make sure that this runs in the node — where should I hook in supervisord?
> > 
> > 
> > 
> > @bas-vk @texuf @sergekh2 
> 
> I'm surprised that it's a service. I expected some kind of static library. Is there state?

Locally, in my opinion, each node should behave as close as possible to how it's going to behave remotely. So each node should get its own service.

jakzale:
> I'm surprised that it's a service. I expected some kind of static library. Is there state?

AFAIU there is nothing preventing us from loading mls-related code as a static library, but running it via ProtoBuf RPC was just faster to do.  We can revisit it post v1.

bas-vk:
> > I'm surprised that it's a service. I expected some kind of static library. Is there state?
> 
> AFAIU there is nothing preventing us from loading mls-related code as a static library, but running it via ProtoBuf RPC was just faster to do. We can revisit it post v1.

Correct, this is a quick & easy method that allows Erik to move forward. The cargo project is organised that we can easily switch to different methods such as ffi or cgo.

PR #1892, opened 2024-12-20
tipping: use tip event filter to parse logs for tip entitlement

salzbrenner:


texuf (changes requested):
this is awesome, and we need this code, but i think it's in the wrong place. I'll fix it in a minute

PR #1893, opened 2024-12-20
Add logging metadata to help debug race conditions in replication

clemire:
This PR separates out some additional logging and population of error metadata in returned errors that mostly only occurs in the case of error conditions, and should not affect total log volume overly. The additional metadata has been helpful in tracking down what may be happening when searching for race conditions in the code.

It also adds a service tester configuration to enable logging for specific test cases.

### DIFF
--- a/core/node/events/stream.go
+++ b/core/node/events/stream.go
@@ -736,16 +736,20 @@ func (s *streamImpl) addEventLocked(ctx context.Context, event *ParsedEvent) err
 	// TODO: for some classes of errors, it's not clear if event was added or not
 	// for those, perhaps entire Stream structure should be scrapped and reloaded
 	if err != nil {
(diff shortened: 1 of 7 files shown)


core/node/events/stream.go
```diff
@@ -734,7 +736,16 @@ func (s *streamImpl) addEventLocked(ctx context.Context, event *ParsedEvent) err
 	// TODO: for some classes of errors, it's not clear if event was added or not
 	// for those, perhaps entire Stream structure should be scrapped and reloaded
 	if err != nil {
-		return err
+		var sb strings.Builder
```
clemire, line 739: If the minipool is large, this line could perhaps have performance implications, I'm not sure how it would affect connect if we returned an error with >> metadata over the wire.

Issue #1894, opened 2024-12-20
Fix replication race condition


clemire:
### Description

Fix the race condition discovered and described in a team message below:

:::
Hey team, I'm working on tracking down some race conditions that were unearthed by the schema lock PR. These race conditions were originally exposed, I think, by how the schema lock strategy changes the timing of the pg store's various operations, and then masked again when I introduced a bug into the PR that caused us to continuously ping a postgres connection, introducing the CPU spike we saw in omega on our last release. When I saw the polling bug in code and attempted to fix it, I saw the reintroduction of race errors in test cases and I've been blocked on merging it for that reason.

The original failing test I was seeing was a heisenbug in `TestReplAdd` that disappeared when I turned on test logging, so I did some strategic logging at the ERROR level. Here's what I discovered:

When adding an event on a replicated stream, we call NewEventReceived rpc from the node2node api on the 4 remote nodes that also host the stream.
once the NewEventReceived rpc is received on the remote node, it removes the original context deadline from the rpc call and replaces it with a deadline of a minimum of 5s. Why? Because this is a write and we don't want it to time out even if the original request times out.
In tests, the quorum pool GoRemote call itself was actually timing out before all nodes received the NewEventReceived request, which was unexpected, and this was causing events to be occasionally be missing on a small subset of nodes in replication tests.
To test this theory, I added a line in quorum_pool.go like so: ctx := context.WithoutCancel(ctx), which removes the deadline from the calling context before executing the remote call, then reran the tests - and TestReplAdd repeatedly passes! 
:::

PR #1895, opened 2024-12-20
[1894][1896][BUGFIX] - Fix replication data race when adding events, occasional test case hang when connecting to unavailable node

clemire:
See #1894 for a description of the first bug, #1896 for the second. The second bug seems to be exposed when the first bug is fixed, so I'm putting them in the same PR to prevent main from getting in a bad state.


clemire:
I kicked off 20 CI runs and this fix looks good. The only failure I saw was in [this run](https://github.com/river-build/river/actions/runs/12438591661/job/34730792939), which appears to be a Blockchain-related  race condition.

clemire:
@sergekh2 for you to consider when you get back - I'm actually not sure if this is an issue in production, on mainnet / testnet?

sergekh2 (approved):

core/node/testutils/testcert/testcert.go
```diff
@@ -98,10 +98,24 @@ func GetHttp2LocalhostTLSConfig() *tls.Config {
 
 func GetHttp2LocalhostTLSClient(ctx context.Context, cfg *config.Config) (*http.Client, error) {
 	return &http.Client{
-		Transport: &http2.Transport{
+		Transport: &http.Transport{
```
bas-vk, line 101: Wat is the reason for this change?

PR #1911, opened 2024-12-24
Media stream optimization: Add internode endpoints 

rhbntt:
As part of the media stream optimization proposal, this PR includes:
1. Internode endpoint to allocate ephemeral stream. There is a flag in "es" table that marks a stream as ephemeral. 
2. Internode endpoint to save ephemeral miniblock.

### DIFF
--- a/core/node/rpc/node2node_ephemeral.go
+++ b/core/node/rpc/node2node_ephemeral.go
@@ -86,12 +86,7 @@ func (s *Service) saveEphemeralMiniblock(
 		return nil, err
 	}
 
-	mbInfo, err := NewMiniblockInfoFromProto(
-		req.Miniblock,
-		NewMiniblockInfoFromProtoOpts{
-			ExpectedBlockNumber: -1,
-		},
-	)
+	mbInfo, err := NewMiniblockInfoFromProto(req.Miniblock, NewParsedMiniblockInfoOpts())
 	if err != nil {
 		return nil, err
 	}
--- a/core/node/rpc/node2node_ephemeral_test.go
+++ b/core/node/rpc/node2node_ephemeral_test.go
@@ -68,7 +68,7 @@ func TestCreateEphemeralStream(t *testing.T) {
 			MiniblockNum:             int64(i + 1),
 			PrevMiniblockHash:        mb.Hash[:],
 			Timestamp:                nil,
-			EventHashes:              nil,
+			EventHashes:              [][]byte{envelope.Hash},
 			Snapshot:                 nil,
 			EventNumOffset:           0,
 			PrevSnapshotMiniblockNum: 0,
@@ -91,5 +91,6 @@ func TestCreateEphemeralStream(t *testing.T) {
 
 	// No events in storage since the stream still ephemeral.
 	// The first miniblock is the stream creation miniblock, the rest 10 are media chunks.
+	// TODO: Make sure AddEvent does not work for ephemeral streams.
 	tt.compareStreamDataInStorage(t, mediaStreamId, 11, 0)
 }


core/node/rpc/node2node.go
```diff
@@ -208,3 +208,46 @@ func (s *Service) saveMiniblockCandidate(
 
 	return &SaveMiniblockCandidateResponse{}, nil
 }
+
+func (s *Service) SaveEphemeralMiniblock(
```
clemire, line 212: There is a debug endpoint I just added for displaying stream state in storage, it would be great if you could add the ephemeral column to it! Perhaps in a follow-up PR

core/node/storage/migrations/000007_add_ephemeral_miniblock.up.sql
```diff
@@ -0,0 +1,22 @@
...
+    numPartitions INT;
+
+    BEGIN
+
+        SELECT num_partitions from settings where single_row_key=true into numPartitions;
+
+        FOR i IN 0.. numPartitions LOOP
+                suffix = LPAD(TO_HEX(i), 2, '0');
+
+                EXECUTE 'ALTER TABLE miniblocks_m' || suffix || ' ADD COLUMN IF NOT EXISTS ephemeral BOOLEAN DEFAULT NULL;';
```
sergekh2, line 17: Is there a value in having this bool in `miniblock` table or is it better to just add a column to the `es` table?

It seems entire stream while being created is in ephemeral state, so there is no need to have this distinction for individual mbs?

core/node/storage/pg_stream_store.go
```diff
@@ -1406,14 +1406,20 @@ func (s *PostgresStreamStore) writeMiniblocksTx(
...
-		[]string{"stream_id", "seq_num", "blockdata"},
+		[]string{"stream_id", "seq_num", "blockdata", "ephemeral"},
 		pgx.CopyFromSlice(
 			len(miniblocks),
 			func(i int) ([]any, error) {
 				if miniblocks[i].Snapshot {
 					newLastSnapshotMiniblock = miniblocks[i].Number
 				}
-				return []any{streamId, miniblocks[i].Number, miniblocks[i].Data}, nil
+
```
sergekh2, line 1416: needs a test - I suspect right now it wouldn't work (since there is no record in es table to lock)

core/node/rpc/node2node.go
```diff
@@ -208,3 +208,46 @@ func (s *Service) saveMiniblockCandidate(
...
+func (s *Service) saveEphemeralMiniblock(
+	ctx context.Context,
+	req *SaveEphemeralMiniblockRequest,
+) (*SaveEphemeralMiniblockResponse, error) {
+	streamId, err := StreamIdFromBytes(req.StreamId)
+	if err != nil {
+		return nil, err
+	}
+
+	stream, err := s.cache.GetStreamWaitForLocal(ctx, streamId)
```
sergekh2, line 242: this wouldn't work - only existing streams can be loaded into cache, ephemeral streams doesn't exist just yet (i.e. they are being created).

Code like this really needs unit tests.

core/node/rpc/node2node_ephemeral.go
```diff
@@ -0,0 +1,141 @@
...
+	req *SaveEphemeralMiniblockRequest,
+) (*SaveEphemeralMiniblockResponse, error) {
+	streamId, err := StreamIdFromBytes(req.StreamId)
+	if err != nil {
+		return nil, err
+	}
+
+	mbInfo, err := NewMiniblockInfoFromProto(
+		req.Miniblock,
+		NewMiniblockInfoFromProtoOpts{
```
clemire, line 91: I just changed this and you will probably get an error after merging.

`NewMiniblockFromProtoOpts{ ExpectedBlockNumber: -1}` is now `NewParsedMiniblockInfoOpts()`, the whole thing is

`mbInfo, err := NewMiniblockInfoFromProto(miniblock, NewParsedMiniblockInfoOpts())`

core/node/rpc/node2node_ephemeral_test.go
```diff
@@ -0,0 +1,95 @@
...
+			},
+		}))
+		tt.require.NoError(err)
+
+		mb.Num = int64(i + 1)
+		mb.Hash = common.BytesToHash(header.Hash)
+	}
+
+	// No events in storage since the stream still ephemeral.
+	// The first miniblock is the stream creation miniblock, the rest 10 are media chunks.
```
clemire, line 93: In either this PR or follow-up, can you add some unit testing that confirms the `AddEvent` rpc fails on this stream, since it's ephemeral? And any other endpoints for regular streams that should be disabled for ephemeral streams, at least until they are committed.

clemire (approved):

sergekh2 (approved):

PR #1912, opened 2024-12-24
Media stream optimization: Add internode CreateMediaStream endpoint

rhbntt:
The given endpoint should be used for creating media stream. It immediately creates a miniblock for the created media stream, and stores it in ephemeral state. It does not do a blockchain transaction to allocate a stream. TBD...

### DIFF
--- a/core/node/rpc/create_media_stream.go
+++ b/core/node/rpc/create_media_stream.go
@@ -168,16 +168,12 @@ func (s *Service) createReplicatedMediaStream(
 	remotes, isLocal := nodes.GetRemotesAndIsLocal()
 	sender := NewQuorumPool("method", "createReplicatedMediaStream", "streamId", streamId)
 
-	// Remove the last node from the list if the current node were not selected.
-	// The media stream receiver node is some sort of constant leader for the given media stream.
-	if !isLocal && len(remotes) > 0 {
-		remotes = remotes[:len(remotes)-1]
-	}
-
 	// Create ephemeral stream within the local node
-	sender.GoLocal(ctx, func(ctx context.Context) error {
-		return s.storage.CreateEphemeralStreamStorage(ctx, streamId, mbBytes)
-	})
+	if isLocal {
+		sender.GoLocal(ctx, func(ctx context.Context) error {
+			return s.storage.CreateEphemeralStreamStorage(ctx, streamId, mbBytes)
+		})
+	}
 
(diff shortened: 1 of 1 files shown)


core/node/rpc/create_media_stream.go
```diff
@@ -0,0 +1,220 @@
...
+		return nil, err
+	}
+
+	nodes := NewStreamNodesWithLock(nodesList, s.wallet.Address)
+	remotes, isLocal := nodes.GetRemotesAndIsLocal()
+	sender := NewQuorumPool("method", "createReplicatedMediaStream", "streamId", streamId)
+
+	// Remove the last node from the list if the current node were not selected.
+	// The media stream receiver node is some sort of constant leader for the given media stream.
+	if !isLocal && len(remotes) > 0 {
```
sergekh2, line 173: it's unclear why it's required. it seems it would work fine even if local node is not selected?

protocol/protocol.proto
```diff
@@ -768,6 +775,14 @@ message StreamAndCookie {
     bool sync_reset = 4;
 }
 
+message StreamAndCreationCookie {
```
sergekh2, line 778: It seems like this message is not required, why not to return CreationCookie?

core/node/rpc/service_test.go
```diff
@@ -1775,91 +1774,3 @@ func TestGetMiniblocksRangeLimit(t *testing.T) {
 		return true
 	}, 20*time.Second, 100*time.Millisecond)
 }
-
-// TestCreateMediaStream tests creating a media stream
-func TestCreateMediaStream(t *testing.T) {
```
sergekh2, line 1780: until all clients are updated (which can take awhile), old way should still work, so it seems that removing this test is premature?

core/node/rpc/create_media_stream.go
sergekh2: it's ok to forward;

Due to browser limitation apps maintain single http2 connection to one of the nodes, so forwarding is unavoidable.

If local node is not selected, lets just forward request to all remotes, it should be fine.

core/node/rpc/service_test.go
sergekh2: it's moved to another file *and modified to use new API*.

point of the previous comment is that old API should still work, so test is still required.

sergekh2 (approved):

PR #1927, opened 2024-12-30
alpha tipping updates

giuseppecrj:


contracts/scripts/interactions/InteractDiamondCut.s.sol
shuhuiluo, line 1: This is hand-wavy. Can we make the script parameterizable?

shuhuiluo (changes requested):

contracts/scripts/interactions/InteractDiamondCut.s.sol
giuseppecrj: no.

shuhuiluo (approved):

PR #1923, opened 2024-12-30
Register miniblocks only on sudden intervals

bas-vk:

PR #1934, opened 2024-12-31
river stream registry: delete genesis when stream advanced

bas-vk:
With https://github.com/river-build/river/pull/1923/commits/1c63faedffe99c6574a75645c1b9ecb79c406fae the genesis block is never removed because `stream.lastMiniblockNum` is updated before the check it's `0`. Do this check and deletion before the update.

### DIFF
--- a/contracts/src/river/registry/facets/stream/StreamRegistry.sol
+++ b/contracts/src/river/registry/facets/stream/StreamRegistry.sol
@@ -114,10 +114,8 @@ contract StreamRegistry is IStreamRegistry, RegistryModifiers {
       revert(RiverRegistryErrors.BAD_ARG);
     }
 
-    // Delete genesis miniblock if `stream` still contains the genesis block after `stream` has advanced since genesis.
-    if (stream.lastMiniblockNum == 0) {
-      delete ds.genesisMiniblockByStreamId[streamId];
-    }
+    // Delete genesis miniblock
+    delete ds.genesisMiniblockByStreamId[streamId];
 
     // Update the stream information
     stream.lastMiniblockHash = lastMiniblockHash;
--- a/contracts/src/river/registry/facets/stream/StreamRegistry.sol
+++ b/contracts/src/river/registry/facets/stream/StreamRegistry.sol
@@ -114,6 +114,11 @@ contract StreamRegistry is IStreamRegistry, RegistryModifiers {
       revert(RiverRegistryErrors.BAD_ARG);
     }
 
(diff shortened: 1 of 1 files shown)


shuhuiluo (approved):

pfives (approved):

contracts/src/river/registry/facets/stream/StreamRegistry.sol
```diff
@@ -114,6 +114,11 @@ contract StreamRegistry is IStreamRegistry, RegistryModifiers {
       revert(RiverRegistryErrors.BAD_ARG);
     }
 
+    // Delete genesis miniblock if `stream` still contains the genesis block after `stream` has advanced since genesis.
+    if (stream.lastMiniblockNum == 0) {
+      delete ds.genesisMiniblockByStreamId[streamId];
+    }
```
shuhuiluo, line 120:
```suggestion
    if (ds.genesisMiniblockByStreamId[streamId].length != 0) {
      delete ds.genesisMiniblockByStreamId[streamId];
    }
```
How about this or always delete? Need to test the gas difference though.
```solidity
    delete ds.genesisMiniblockByStreamId[streamId];
```

shuhuiluo (changes requested):

contracts/src/river/registry/facets/stream/StreamRegistry.sol
bas-vk: If one of these options save on gas we should go for that ofc. But I don't see why either the length comparison or the always delete option would be more efficient. But that is maybe my lack of solidity/evm knowledge. Could you do the simulations and make the change if one of these options is more efficient? Not sure what a good way is to test the differences.

shuhuiluo: I think simply
```solidity
    delete ds.genesisMiniblockByStreamId[streamId];
```
is the most efficient. What the bytecode generated by the compiler does is reading the length of the bytes array and erasing the storage slots one by one in a loop. If the length of the mini block is zero, it will skip.
I will do benchmarks when I work on `StreamRegistry` later. But this should do it for now. It will also remove the genesis mini block for streams whose `lastMiniblockNum` is no longer 0.

shuhuiluo, line 120:
```suggestion
    // Delete genesis miniblock
    delete ds.genesisMiniblockByStreamId[streamId];
```

shuhuiluo (changes requested):

shuhuiluo (approved):

PR #1940, opened 2024-12-31
prune stream registry functions

giuseppecrj:


contracts/src/river/registry/facets/stream/StreamRegistry.sol
```diff
@@ -285,70 +281,4 @@ contract StreamRegistry is IStreamRegistry, RegistryModifiers {
...
-    for (uint256 i = 0; i < streamCount; ++i) {
-      bytes32 streamId = streamIds[i];
-      Stream storage stream = ds.streamById[streamId];
-      if (stream.nodes.length == 0) continue;
-      streams[foundCount++] = StreamWithId({id: streamId, stream: stream});
-    }
-    return (foundCount, streams);
-  }
-
-  function getStreamsOnNode(
```
shuhuiluo, line 303: I thought `getStreamsOnNode` will be used.

giuseppecrj: nope

giuseppecrj: getStreamCountOnNode is

shuhuiluo: Isn't it supposed to make node sync faster?

giuseppecrj: I'm removing all functions we agreed on removing, if we need add new ones that are optimized we will add them back

shuhuiluo (approved):

PR #1941, opened 2024-12-31
River timeline event - add back isLocalPending and isSendFailed

texuf:
@miguel do you remember why you didn’t add these?


miguel-nascimento:
No, I don’t remember why, thanks for adding it back 🙏

Issue #1859, opened 2024-12-18
Archiver implement miniblock scrubbing


clemire:
### Description

validate that each miniblock actually hashes to the expected hash, following the linked list back in time to block 0.

Ideally we would do this in a way that easily ports to the stream node.

PR #1943, opened 2025-01-01
[RIVER-1859] Miniblock scrubber

clemire:
#1859 

Built a miniblock scrubber that checks for corrupted blocks in the stream. As part of this work I expanded the existing miniblockInfo construction to optionally validate a bunch of extra stuff. A piece of follow-up work here is validating snapshot state based on the miniblocks that have come in since the previous snapshot.

Design
======
- the miniblock scrubber will always schedule a scrub, it does not deny a scheduled scrub based on the recency of the last scrub
- The scrubber scrubs from a minimum block to the latest block, and outputs a report to a channel
- The scrubber expects a report consumer to be constantly polling, it will block the scrubber if the buffer becomes full because tasks block on emitting to the channel

Unit tests
=======
- a single-block stream
- a multi-block stream
- a corrupted stream

Extra stuff:
========
- rename existing scrubber for easier disambiguation of scrubbers
- some spelling fixes


core/cmd/stream_cmd.go
```diff
@@ -98,10 +98,11 @@ func runStreamGetEventCmd(cmd *cobra.Command, args []string) error {
 
 	for n, miniblock := range miniblocks.Msg.GetMiniblocks() {
 		// Parse header
+		expectedBlockNumber := from + int64(n)
 		info, err := events.NewMiniblockInfoFromProto(
 			miniblock,
 			events.NewMiniblockInfoFromProtoOpts{
-				ExpectedBlockNumber: from + int64(n),
+				ExpectedBlockNumber: &expectedBlockNumber,
```
texuf, line 105: why are you doing reference here now? can you explain to me?

core/node/events/stream_cache.go
```diff
@@ -219,7 +219,7 @@ func (s *streamCacheImpl) onStreamAllocated(
 		stream.nodesLocked.Reset(event.Nodes, s.params.Wallet.Address)
 		stream, created, err := s.createStreamStorage(ctx, stream, event.GenesisMiniblock)
 		if err != nil {
-			dlog.FromCtx(ctx).Error("Failed to allocate stream", "err", err, "streamId", stream.streamId)
+			dlog.FromCtx(ctx).Error("Failed to allocate stream", "err", err) // , "streamId", stream.streamId)
```
texuf, line 222: did you mean to leave in this comment? if so can you delete the commented out code?

core/cmd/stream_cmd.go
clemire: So I can explicitly check for zero values if intended. Does that make sense? The original implementation avoids this but that makes validating block 0 less strict.

clemire: If that's odd I can always revert it, but it seemed like the lowest overhead solution to having 0 as a "nonzero" for specifying that the user did set a value which should be used for validation.

core/node/events/stream_cache.go
clemire: No, that's my mistake. You caught me in the middle of iterating on this PR! Sorry about that.

sergekh2 (approved):

core/node/events/miniblock.go
```diff
@@ -171,56 +175,163 @@ func NewMiniblockInfoFromBytes(bytes []byte, expectedBlockNumber int64) (*Minibl
...
+	}
 
 	return NewMiniblockInfoFromProto(&pb, opts)
 }
 
 type NewMiniblockInfoFromProtoOpts struct {
-	ExpectedBlockNumber int64
-	DontParseEvents     bool
+	ExpectedBlockNumber               *int64
+	ExpectedLastMiniblockHash         *common.Hash
```
sergekh2, line 210: I think it should be
```suggestion
	ExpectedPrevMiniblockHash         *common.Hash
```

sergekh2 (approved):

PR #1944, opened 2025-01-01
refactor(contract): rename all RiverPoints references to TownsPoints

shuhuiluo:
- Rename all RiverPoints references to TownsPoints
- Remove ERC20 primitives and update imports


giuseppecrj:
Why not just calling it Points?

PR #1945, opened 2025-01-01
Disable miniblock validation to allow overwriting pending go code fix.

sergekh2:


jterzis (commented):
We should probably deprecate the check also in setStreamLastMiniblock for posterity

sergekh2:
> We should probably deprecate the check also in setStreamLastMiniblock for posterity

This method is not used anymore and is going to be removed

PR #1948, opened 2025-01-02
register miniblock 1 always to ensure genesis block deletion

bas-vk:


core/node/rpc/stream_test.go
```diff
@@ -131,16 +143,11 @@ func TestMiniBlockProductionFrequency(t *testing.T) {
 			}
 		}
 
-		if len(logsFound) < 3 {
+		if len(logsFound) < 10 {
```
sergekh2, line 146: but why 10 here? test intention is to produce some mbs, restart, and produce more. 

Number of mbs here should be smaller than in the second loop.

Please run this test case with -v and RIVER_TEST_PRINT=1 to see how mbs are produced step-to-step.

texuf (approved):

sergekh2 (approved):

PR #1955, opened 2025-01-03
add basic MLS snapshotting

erikolsson:
this PR adds an MLS stream state view + basic MLS snapshotting.

stacked on #1949


core/node/events/snapshot.go
```diff
@@ -607,11 +607,34 @@ func update_Snapshot_Member(
...
+	mlsPayload *MemberPayload_Mls,
+) error {
+	if iSnapshot.Members.GetMls() == nil {
+		iSnapshot.Members.Mls = &MemberPayload_Snapshot_Mls{}
+	}
+	snapshot := iSnapshot.Members.Mls
+	switch content := mlsPayload.Content.(type) {
+	case *MemberPayload_Mls_InitializeGroup_:
+		if len(snapshot.ExternalGroupSnapshot) > 0 || len(snapshot.GroupInfoMessage) > 0 {
+			return RiverError(Err_INVALID_ARGUMENT, "duplicate mls initialization")
```
texuf, line 628: if this is an expected state it shouldn't return an error, just comment and return nil

core/node/events/snapshot.go
```diff
@@ -607,11 +607,34 @@ func update_Snapshot_Member(
...
+		}
+		snapshot.ExternalGroupSnapshot = content.InitializeGroup.ExternalGroupSnapshot
+		snapshot.GroupInfoMessage = content.InitializeGroup.GroupInfoMessage
+		return nil
```
texuf, line 632: we need to copy these instead of just assigning, otherwise we get data races 
snapshot.ExternalGroupSnapshot = content.InitializeGroup.ExternalGroupSnapshot
		snapshot.GroupInfoMessage = content.InitializeGroup.GroupInfoMessage

texuf: I had this problem with the key solicitations, but i needed to clone the object, event := proto.Clone(keySolicitation).(*MemberPayload_KeySolicitation)
here because we're just moving slices, maybe we don't? worth looking into

texuf (approved):

PR #1960, opened 2025-01-03
add slot for uploading epoch secrets

erikolsson:
I find parts of the go code a bit clunky — constantly checking if fields are set. Input more than welcome!

We will need more validation here, let's sync on how to do it. We also need to account for malicious users uploading invalid keys.

Stacked on #1959

### DIFF
--- a/packages/sdk/src/streamStateView_Mls.ts
+++ b/packages/sdk/src/streamStateView_Mls.ts
@@ -54,7 +54,9 @@ export class StreamStateView_Mls extends StreamStateView_AbstractContent {
                 break
             case 'epochSecrets':
                 for (const secret of mlsEvent.content.value.secrets) {
-                    this.epochSecrets[secret.epoch.toString()] = secret.secret
+                    if (!this.epochSecrets[secret.epoch.toString()]) {
+                        this.epochSecrets[secret.epoch.toString()] = secret.secret
+                    }
                 }
                 break
             case undefined:


core/node/events/snapshot.go
```diff
@@ -144,7 +144,10 @@ func make_SnapshotMembers(iInception IsInceptionPayload, creatorAddress []byte)
 			}, &MemberPayload_Snapshot_Member{
```
texuf, line 144: can we fix this to make the MemberPayload_Snapshot once and then do the custom joined logic in the switch statement

texuf: pushed an update

core/node/events/snapshot.go
```diff
@@ -650,6 +663,11 @@ func update_Snapshot_Mls(
 		}
 		snapshot.Members[memberAddress].SignaturePublicKeys = append(snapshot.Members[memberAddress].SignaturePublicKeys, content.ExternalJoin.SignaturePublicKey)
 		return nil
+	case *MemberPayload_Mls_EpochSecrets_:
+		for _, secret := range content.EpochSecrets.Secrets {
+			snapshot.EpochSecrets[secret.Epoch] = secret.Secret
+		}
```
texuf, line 668: right now if the secret is added twice, we're accepting the second one.. is that desired? normally first one wins

core/node/events/stream_viewstate_mls.go
```diff
@@ -91,4 +92,39 @@ func (r *streamViewImpl) GetMlsExternalJoinRequest() (*mls_tools.ExternalJoinReq
...
+	}
+	updateFn := func(e *ParsedEvent, minibockNum int64, eventNum int64) (bool, error) {
+		switch payload := e.Event.Payload.(type) {
+		case *protocol.StreamEvent_MemberPayload:
+			switch content := payload.MemberPayload.Content.(type) {
+			case *protocol.MemberPayload_Mls_:
+				switch content.Mls.Content.(type) {
+				case *protocol.MemberPayload_Mls_EpochSecrets_:
+					for _, epochSecret := range content.Mls.GetEpochSecrets().GetSecrets() {
+						epochSecrets[epochSecret.Epoch] = epochSecret.Secret
```
texuf, line 114: same

packages/sdk/src/streamStateView_Mls.ts
```diff
@@ -49,6 +52,11 @@ export class StreamStateView_Mls extends StreamStateView_AbstractContent {
                     mlsEvent.content.value.signaturePublicKey,
                 )
                 break
+            case 'epochSecrets':
+                for (const secret of mlsEvent.content.value.secrets) {
+                    this.epochSecrets[secret.epoch.toString()] = secret.secret
```
texuf, line 57: same

PR #1963, opened 2025-01-03
WIP: Go MLS snapshotting (Help wanted)

erikolsson:
I could use some help here but this PR more or less describes what we need to accomplish.

Some parts of MLS are impossible for Go to snapshot on its own since it doesn't understand the commits — instead we need to round up all commits and process them in the Rust MLS Snapshot code, while keeping track of the latest valid `GroupInfoMessage`.

There is no way to _guarantee_ that an invalid commit slips through our checks and makes it into the stream, this is why we cannot simply rely on the latest `groupInfoMessage`, we need to rely on the latest `groupInfoMessage` accompanied by a valid commit. There are more issues to solve here but this is a start.

I wrote some comments inline, ideas more than welcome!

stacked on #1962

### DIFF
--- a/core/node/events/stream_view.go
+++ b/core/node/events/stream_view.go
@@ -303,39 +303,6 @@ func (r *streamViewImpl) makeMiniblockHeader(
 	if proposal.ShouldSnapshot {
 		snapshot = proto.Clone(r.snapshot).(*Snapshot)
 		mlsSnapshotRequest := r.makeMlsSnapshotRequest()
-		// TODO: this function should only run for streams with MLS support
-		mlsUpdateFn := func(e *ParsedEvent) {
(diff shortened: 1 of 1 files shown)


core/node/events/stream_view.go
```diff
@@ -300,6 +302,41 @@ func (r *streamViewImpl) makeMiniblockHeader(
...
+						}
+						mlsSnapshotRequest.Commits = append(mlsSnapshotRequest.Commits, commitInfo)
+					}
+				default:
+					break
+				}
+			default:
+				break
+			}
+		}
```
texuf, line 338: can you make this a real function?

texuf (approved):

PR #1970, opened 2025-01-06
Add MLS-specific RPC call for cheaper MLS scrollbacks

erikolsson:
MLS groups require all commits to be applied in order, which means that we need to "close the gap" when resuming sync.
For some streams (Spaces specifically), the snapshots can grow very large (easily into multiple megabytes), making this operation somewhat expensive to perform.

By exposing an MLS-only RPC call, we can make this operation super lightweight. This is my rough guess of what such a call could look like, feedback more than welcome.

The idea here is that each snapshot contains _all commits since the previous snapshot_.


texuf (commented):
The matrix protocol has a filter param that you can pass to the server to get various types of messages... did you consider that approach?

protocol/protocol.proto
```diff
@@ -831,6 +832,19 @@ message GetLastMiniblockHashResponse {
...
 
+// GetMLSSnapshotRequest is a request to get exclusively MLS-related snapshot
+// data at a given miniblock number.
+message GetMlsSnapshotRequest {
+    bytes stream_id = 1;
+    int64 miniblock_num = 2;
+}
+
+message GetMlsSnapshotResponse {
+    MemberPayload.Snapshot.Mls mls = 1;
```
sergekh2, line 843: lets not return sub-events on the protocol level: they can't be validated  since they don't carry signatures. Let's return event here. Specifically let's call api GetMiniblockHeader and return header of the requested miniblock (which contains snapshot with mls details inside).

PR #1972, opened 2025-01-06
wip add CommitLeaves MLS Event

erikolsson:
When users leave streams (voluntarily or involuntarily), we need to remove them from the group.
(I think) we have all the information we need in the `map<string, Member> members = 3;` on the MLS group in protocol.proto


texuf:
who is going to add the leave message?

texuf (approved):

PR #1986, opened 2025-01-08
[RIVER-1859] Add scrubber to archiver

clemire:
Add the miniblock scrubber to the archiver

To make this successful, I also did the following:
- Update stream corruption tracking to consider corruption due to unavailability or scrub failure
- Expose corrupted streams on the archiver via a debug endpoint gated by a config var
- Additionally, add a cli subcommand for dumping stream output to test streams that are reported as corrupt locally by the archiver running with omega config
- Also added configuration for running the archiver locally against public environments and updated the README with usage instructions.

Debug endpoint example output
Screenshot 2025-01-10 at 12 34 42 PM


core/env/archiver-local.yaml
```diff
@@ -0,0 +1,25 @@
+archive:
```
sergekh2, line 1: lets move this file to core/env/local/archiver dir

core/node/events/miniblock.go
```diff
@@ -354,6 +354,19 @@ func NewMiniblockInfoFromProto(pb *Miniblock, opts *ParsedMiniblockInfoOpts) (*M
 					Tag("headerEventHash", blockHeader.EventHashes[i])
 			}
 		}
+
+		// Validate event hashes match the hashes stored in the header.
+		for i, event := range events {
```
sergekh2, line 359: but this block of code is exactly the same as the block above.

should we turn off this check by default for now? i.e. is it going to bring down production when deployed?

core/node/storage/pg_stream_store.go
```diff
@@ -2026,18 +2031,18 @@ func (s *PostgresStreamStore) getLastMiniblockNumberTx(
 	err = tx.QueryRow(
 		ctx,
 		s.sqlForStream(
-			"SELECT MAX(seq_num) FROM {{miniblocks}} WHERE stream_id = $1",
+			"SELECT COALESCE(MAX(seq_num), -1) FROM {{miniblocks}} WHERE stream_id = $1",
```
sergekh2, line 2034: but why? this query might produce different query plan which now needs to be looked at.

core/node/events/miniblock.go
clemire: oh my bad!

I don't think it will because I have not seen any streams fail to pass the scrubber and so far, all unfetchable streams have been resolved by a call to getStream. Furthermore all streams that aren't fully fetchable now are unfetchable before we deploy this change. However the new checks will for sure cause some retrievals to fail for any corrupt streams that have mismatched length or hashes of events in the header and the body of a miniblock.

core/node/storage/pg_stream_store.go
clemire: Honestly I'm not sure, maybe this is amerge artifact too, I thought I was using a different API call, GetMaxArchivedMiniblockNumber. I'll revert / fix

core/node/storage/pg_stream_store.go
```diff
@@ -2032,12 +2037,12 @@ func (s *PostgresStreamStore) getLastMiniblockNumberTx(
 		streamID,
 	).Scan(&maxSeqNum)
 	if err != nil {
-		if err == pgx.ErrNoRows {
-			return 0, RiverError(Err_INTERNAL, "Stream exists in es table, but no miniblocks in DB")
-		}
 		return 0, err
 	}
 
+	if maxSeqNum == -1 {
```
sergekh2, line 2043: this code needs to be dropped, and deleted code above returned for this to be correct

PR #1992, opened 2025-01-09
add config flag for globally enabling/disabling mls events

erikolsson:
special request from @texuf to be able to globally toggle MLS

### DIFF
--- a/core/node/rules/can_add_event.go
+++ b/core/node/rules/can_add_event.go
@@ -588,6 +588,10 @@ func (params *aeParams) canAddMemberPayload(payload *StreamEvent_MemberPayload) 
 			check(ru.validMemberBlockchainTransaction_IsUnique).
 			check(ru.validMemberBlockchainTransaction_ReceiptMetadata)
 	case *MemberPayload_Mls_:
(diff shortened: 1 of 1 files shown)


core/config/config.go
```diff
@@ -149,6 +150,9 @@ type Config struct {
 
 	// Disable base chain contract usage.
 	DisableBaseChain bool
+	
+	// Enable MemberPayload_Mls.
```
clemire, line 154: If there are any other settings for MLS, etc, you think we might need to add to the node, it could be good to create an EncryptionConfig object and then we can pass that to the rules evaluation. Otherwise, looks good!

clemire (approved):

clemire:
- how can we override this value in tests?

In the go tests, we can add an option to initializing the service tester to disable this in the config. For end to end tests, we'd have to update the configuration we use for local/multi deploys and use this setting for sdk tests that require nodes to be running.

texuf (approved):
thank you
add it to these files for local dev
core/env/local/multi/config.yaml
core/env/local/multi_ne/config.yaml

texuf (approved):

PR #1993, opened 2025-01-09
notification service: bugfix tracking membership in new towns

bas-vk:
Currently the notification service doesn't handle sync updates that include miniblocks. The result is that once events are dropped from the minipool calculating the stream membership must be done from `snapshot + blocks + minipool` but is done from `snapshot + minipool`.

For notifications this is not an issue because these are handled immediately when the notification service receives them through a sync update. For user membership it needs to respect join/leave events in miniblocks.

This PR adds an unittest that hits the problem and introduces the `ApplyBlock` function that is called for incoming miniblocks.

### DIFF
--- a/core/node/notifications/sync/streams_tracker_worker_connectgo.go
+++ b/core/node/notifications/sync/streams_tracker_worker_connectgo.go
@@ -270,10 +270,15 @@ func (s *StreamTrackerConnectGo) Run(
 					continue
 				}
 
-				// apply update
+				for _, block := range update.GetStream().GetMiniblocks() {
+					if err := trackedStream.ApplyBlock(block, onChainConfig.Get()); err != nil {
+						log.Error("Unable to apply block", "stream", streamID, "err", err)
+					}
+				}
+
 				for _, event := range update.GetStream().GetEvents() {
-					if err := trackedStream.HandleEvent(syncCtx, event); err != nil {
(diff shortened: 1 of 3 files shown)


sergekh2 (approved):

core/node/notifications/sync/streams_tracker_worker_connectgo.go
```diff
@@ -270,10 +270,15 @@ func (s *StreamTrackerConnectGo) Run(
...
-					if err := trackedStream.HandleEvent(syncCtx, event); err != nil {
-						log.Error("Unable to handle event", "stream", streamID, "err", err)
+					if err := trackedStream.ApplyEvent(syncCtx, event); err != nil {
+						log.Error("Unable to apply event", "stream", streamID, "err", err)
+					}
+				}
+
+				for _, block := range update.GetStream().GetMiniblocks() {
+					if err := trackedStream.ApplyBlock(block, onChainConfig.Get()); err != nil {
+						log.Error("Unable to apply block", "stream", streamID, "err", err)
```
texuf, line 281: seems like you should apply the blocks before the events... is there a reason to do it in the other order?

bas-vk: My line of thinking was that events first enter the minipool and are dropped from the pool when included in a block. So the block is processed after the events are added to replicate this.

PR #1996, opened 2025-01-09
Add support for welcome messages

erikolsson:
This PR adds basic validation of welcome packages.
The idea is that key packages are cleared from Mls.pendingKeyPackages once committed into the group.
- need to decide where welcome packages should go. they're not terribly expensive but may require the recipient to perform scrollbacks which is annoying. we could send them into their inbox instead
- needs more checks in rust

stacked on #1995

### DIFF
--- a/packages/stress/package.json
+++ b/packages/stress/package.json
@@ -29,7 +29,7 @@
         "pino-pretty": "^10.2.3"
     },
     "devDependencies": {
-        "@river-build/mls-rs-wasm-node": "^0.0.7",
+        "@river-build/mls-rs-wasm-node": "^0.0.14",
         "@types/debug": "^4.1.8",
         "@types/lodash": "^4.14.186",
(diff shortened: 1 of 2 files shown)


texuf (approved):

packages/sdk/package.json
```diff
@@ -30,7 +30,7 @@
         "@river-build/dlog": "workspace:^",
         "@river-build/encryption": "workspace:^",
         "@river-build/generated": "workspace:^",
-        "@river-build/mls-rs-wasm": "^0.0.11",
+        "@river-build/mls-rs-wasm": "^0.0.13",
```
jakzale, line 33: Please make sure that you also bump `mls-rs-wasm-node` in `packages/stress/package.json`.  Not sure if this has to be as part of this PR or a different PR, but the version of those two packages should be kept in sync.

jakzale (approved):

PR #2012, opened 2025-01-10
add GetMiniblockHeader + commits since last snapshot

erikolsson:
This PR
- Adds `GetMiniblockHeader` rpc
- Adds `MemberPayload.Snapshot.Mls.CommitsSinceLastSnapshot`
the idea is that users can fetch just snapshots to close the gap.

### DIFF
--- a/core/node/rpc/get_miniblock_header.go
+++ b/core/node/rpc/get_miniblock_header.go
@@ -11,6 +11,13 @@ func (s *Service) GetMiniblockHeader(
 	ctx context.Context,
 	req *connect.Request[GetMiniblockHeaderRequest],
 ) (*connect.Response[GetMiniblockHeaderResponse], error) {
+	return executeConnectHandler(ctx, req, s, s.getMiniblockHeaderImpl, "GetMiniblockHeader")
+}
+
+func (s *Service) getMiniblockHeaderImpl(
+	ctx context.Context,
+	req *connect.Request[GetMiniblockHeaderRequest],
+) (*connect.Response[GetMiniblockHeaderResponse], error) {
 	miniblocksRequest := &GetMiniblocksRequest{
 		StreamId: req.Msg.StreamId,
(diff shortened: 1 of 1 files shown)


protocol/protocol.proto
```diff
@@ -971,12 +972,22 @@ message InfoResponse {
     string version = 3;
 }
 
+message GetMiniblockHeaderRequest {
+    bytes stream_id = 1;
+    int64 miniblock_num = 2;
+}
+
+message GetMiniblockHeaderResponse {
+    MiniblockHeader header = 1;
```
sergekh2, line 981: This should be an `Envelope`: then it will have signatures.

(on the receiving side it should be parsed as any other event to extract header)

core/node/rpc/get_miniblock_header.go
```diff
@@ -0,0 +1,26 @@
...
+func (s *Service) GetMiniblockHeader(
+	ctx context.Context,
+	req *connect.Request[GetMiniblockHeaderRequest],
+) (*connect.Response[GetMiniblockHeaderResponse], error) {
+	miniblocksRequest := &GetMiniblocksRequest{
+		StreamId: req.Msg.StreamId,
+		FromInclusive: req.Msg.MiniblockNum,
+		ToExclusive: req.Msg.MiniblockNum + 1,
+	}
+	miniblocksResponse, err := s.GetMiniblocks(ctx, connect.NewRequest(miniblocksRequest))
```
bas-vk, line 19: Several GetXXX rpc methods are executed through `executeConnectHandler`. This was added last month and I don't know if this is something temporarily or permanent.

@sergekh2 do we need to wrap this call?

sergekh2: yes, we do, I was going to fix it when I'll clean up rest of the stuff that needs cleaning up in RPC handlers.

sergekh2 (approved):

PR #2016, opened 2025-01-11
Hybrid asymmetric/symmetric encryption for streams (not enabled)

texuf:
We're getting lots of groups wanting to expand their towns. This should allow us to have much bigger towns.

just pass around a small number of aes keys instead of N asymmetric keys per stream

this sets the default to the new decryption, probably want to run tests against both for the time being


packages/encryption/src/crypto_utils.ts
```diff
@@ -0,0 +1,73 @@
+import * as crypto from 'crypto'
+
+// Function to encrypt a message
+export function encryptAESCBC(plainText: string, key: string, iv: string): string {
```
sergekh2, line 4: it's a bit confusing: this entire implementation is for node.js, 
browser native crypto APIs are quite different:
https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto
is this intentional?

protocol/payloads.proto
```diff
@@ -158,6 +158,15 @@ message SessionKeys {
     repeated string keys = 1;
 }
 
+message HybridGroupSessionKey {
+    string session_id = 1;
+    string stream_id = 2;
+    string iv = 3;
+    string key = 4;
```
sergekh2, line 165: but why not `bytes` for all fields? it essentially 2xes encoded object size

protocol/payloads.proto
```diff
@@ -158,6 +158,15 @@ message SessionKeys {
     repeated string keys = 1;
 }
 
+message HybridGroupSessionKey {
+    string session_id = 1;
+    string stream_id = 2;
+    string iv = 3;
```
sergekh2, line 164: iv shouldn't be here: it should be randomly generated for each encrypted message.

packages/encryption/src/crypto_utils.ts
texuf: GCM requires a nonce or else it's "catastrophically defective" I imagine sharing it will break things.

texuf: i was wondering if there should be salt... the iv is the salt yeah? @erikolsson that was the other thing i wanted to figure out that I couldn't think of last night.

protocol/payloads.proto
texuf: because other code wanted strings... i can try to make bytes work

protocol/payloads.proto
texuf: 🤦

PR #2017, opened 2025-01-11
Media stream optimization: addStream smart contract function

rhbntt:
As part of the media stream optimization feature, a smart contract function to create a stream should be introduced. The node will use this endpoint to immediately create a stream with sealed state.

### DIFF
--- a/contracts/test/river/registry/stream/StreamRegistry.t.sol
+++ b/contracts/test/river/registry/stream/StreamRegistry.t.sol
@@ -222,7 +222,7 @@ contract StreamRegistryTest is
     );
   }
 
-  /// @notice This test is to ensure that the node who is calling the allocateSealedStream function is registered.
+  /// @notice This test is to ensure that the node who is calling the addStream function is registered.
   function test_revertWhen_addStream_nodeNotRegistered(
     address nodeOperator,
     TestStream memory testStream,
@@ -247,7 +247,7 @@ contract StreamRegistryTest is
     );
   }
 
-  /// @notice This test is to ensure that the nodes being passed in are registered before allocating a sealed stream.
+  /// @notice This test is to ensure that the nodes being passed in are registered before allocating a stream.
   function test_revertWhen_addStream_nodesNotRegistered(
     address nodeOperator,
     address randomNode,


contracts/src/river/registry/facets/stream/StreamRegistry.sol
```diff
@@ -60,6 +60,47 @@ contract StreamRegistry is IStreamRegistry, RegistryModifiers {
     );
   }
 
+  /// @inheritdoc IStreamRegistry
+  function allocateSealedStream(
```
giuseppecrj, line 64: Don't think this is the best approach since we can end up with a bunch of param driven functions rather than allow node to pass a stream struct defining the requirements

giuseppecrj (changes requested):

contracts/src/river/registry/facets/stream/StreamRegistry.sol
giuseppecrj: I would also rewrite to `createStream(Stream)`

contracts/src/river/registry/facets/stream/StreamRegistry.sol
```diff
@@ -60,6 +60,30 @@ contract StreamRegistry is IStreamRegistry, RegistryModifiers {
     );
   }
 
+  /// @inheritdoc IStreamRegistry
+  function createStream(
+    bytes32 streamId,
+    bytes32 genesisMiniblockHash,
+    Stream memory stream
```
giuseppecrj, line 67: should lastMiniblockNum and reserved0 still be 0? we should add a check for this if so.

contracts/src/river/registry/facets/stream/StreamRegistry.sol
```diff
@@ -60,6 +60,30 @@ contract StreamRegistry is IStreamRegistry, RegistryModifiers {
...
+    uint256 nodeCount = stream.nodes.length;
+    for (uint256 i = 0; i < nodeCount; ++i) {
+      if (!ds.nodes.contains(stream.nodes[i]))
+        revert(RiverRegistryErrors.NODE_NOT_FOUND);
+    }
+
+    ds.streams.add(streamId);
+    ds.streamById[streamId] = stream;
+    ds.genesisMiniblockHashByStreamId[streamId] = genesisMiniblockHash;
+
```
giuseppecrj, line 83: are we removing? ` ds.genesisMiniblockByStreamId[streamId] = genesisMiniblock;`

giuseppecrj: great! can you add it to your function?

contracts/src/river/registry/facets/stream/StreamRegistry.sol
giuseppecrj: ?

contracts/src/river/registry/facets/stream/StreamRegistry.sol
giuseppecrj: this was answered during river triage, since we're migrating to not requiring bytes `genesisMiniblock` this function is the correct way to go. Thanks!

giuseppecrj:
Once this is merged, we need to create governance proposals on gamma and omega.

giuseppecrj (approved):

contracts/test/river/registry/stream/StreamRegistry.t.sol
```diff
@@ -135,6 +136,148 @@ contract StreamRegistryTest is
...
+    vm.prank(node.node);
+    vm.expectRevert(bytes(RiverRegistryErrors.ALREADY_EXISTS));
+    streamRegistry.addStream(
+      testStream.streamId,
+      testStream.genesisMiniblockHash,
+      streamToCreate
+    );
+  }
+
+  /// @notice This test is to ensure that the node who is calling the allocateSealedStream function is registered.
```
clemire, line 225: comment nit: `allocateSealedStream` is `addStream`, right?

contracts/test/river/registry/stream/StreamRegistry.t.sol
```diff
@@ -135,6 +136,148 @@ contract StreamRegistryTest is
...
+    vm.prank(node.node);
+    vm.expectRevert(bytes(RiverRegistryErrors.NODE_NOT_FOUND));
+    streamRegistry.addStream(
+      testStream.streamId,
+      testStream.genesisMiniblockHash,
+      streamToCreate
+    );
+  }
+
+  /// @notice This test is to ensure that the nodes being passed in are registered before allocating a sealed stream.
```
clemire, line 250: Does it matter here if the stream is sealed, btw?

clemire (approved):
Approval with nits/clarification

sergekh2 (approved):

PR #2022, opened 2025-01-12
Group Encryption Refactor 1: Patch up crypto backend abstraction

texuf:
make the device private, force everything through the same api


packages/encryption/src/tests/decryptionExtensions.test.ts
```diff
@@ -406,6 +398,12 @@ class MockGroupEncryptionClient
         return Promise.resolve({})
     }
 
+    public getMiniblockInfo(
+        _streamId: string,
+    ): Promise<{ miniblockNum: bigint; miniblockHash: Uint8Array }> {
+        return Promise.resolve({ miniblockNum: 0n, miniblockHash: new Uint8Array() })
+    }
+
```
texuf, line 406: the new getMiniblockInfo isn't needed until we add the new hybrid encryption

PR #2040, opened 2025-01-14
export entitlement checking to base registry

giuseppecrj:
This PR migrates entitlement checking to the base registry and finalizes the check in the caller, being the space.

entitlement-checker-v2


contracts/src/spaces/facets/gated/EntitlementGatedBase.sol
```diff
@@ -164,6 +155,83 @@ abstract contract EntitlementGatedBase is IEntitlementGatedBase {
...
+        }
+      }
+    }
+
+    // if the entitlement checker has not been set, set it
+    if (address(ds.entitlementChecker) == address(0)) {
+      _setFallbackEntitlementChecker();
+    }
+
+    if (!transaction.hasBenSet) {
```
clemire, line 186: not: do you need to check !transaction.hasBeenSet considering line 172?

contracts/src/spaces/facets/gated/EntitlementGatedBase.sol
```diff
@@ -164,6 +155,83 @@ abstract contract EntitlementGatedBase is IEntitlementGatedBase {
...
+
+    bool allRoleIdsCompleted = _checkAllRoleIdsCompleted(transactionId);
+
+    if (allRoleIdsCompleted) {
+      _removeTransaction(transactionId);
+    }
+
+    if (result == NodeVoteStatus.PASSED || allRoleIdsCompleted) {
+      _onEntitlementCheckResultPosted(transactionId, result);
+      emit EntitlementCheckResultPosted(transactionId, result);
```
clemire, line 227: In viem, at least, I think I need to look up the actual transaction that emitted this log to get the address of the node that sent it. I'm not sure about ethers.

Would it be much more expensive to add the sender to the emitted event here as a part of this change?

contracts/src/spaces/facets/gated/EntitlementGatedBase.sol
giuseppecrj: you mean remove the check and just set directly?

contracts/src/spaces/facets/gated/EntitlementGatedBase.sol
giuseppecrj: the poster would be the base registry since that's where the voting resolution would take place now.

giuseppecrj: we can add an event there to just log what the node posted

clemire (approved):
xchain service-wise, this change looks good. I'll defer to @shuhuiluo for final contract review

contracts/src/base/registry/facets/checker/EntitlementChecker.sol
```diff
@@ -146,20 +148,59 @@ contract EntitlementChecker is IEntitlementChecker, Facet {
...
       nodes
     );
   }
 
+  function requestEntitlementCheckV2(
+    address walletAddress,
+    bytes32 transactionId,
+    uint256 requestId
+  ) external payable {
+    address space = msg.sender;
```
shuhuiluo, line 170: Should we verify the space address?

contracts/src/xchain/XChainLib.sol
```diff
@@ -0,0 +1,43 @@
...
+
+// contracts
+
+library XChainLib {
+  // keccak256(abi.encode(uint256(keccak256("xchain.entitlement.transactions.storage")) - 1)) & ~bytes32(uint256(0xff))
+  bytes32 internal constant STORAGE_SLOT =
+    0xf501c51c066c21fd640901535874a71171bb35113f6dc2832fce1b1f9da0cc00;
+
+  struct Check {
+    bytes32 txId;
```
shuhuiluo, line 18:
```suggestion
```
Not needed.

contracts/src/spaces/facets/gated/EntitlementGatedBase.sol
```diff
@@ -164,6 +157,63 @@ abstract contract EntitlementGatedBase is IEntitlementGatedBase {
...
+      CustomRevert.revertWith(EntitlementGated_InvalidEntitlement.selector);
+
+    EntitlementGatedStorage.Layout storage ds = EntitlementGatedStorage
+      .layout();
+    Transaction storage transaction = ds.transactions[transactionId];
+
+    // if the entitlement checker has not been set, set it
+    if (address(ds.entitlementChecker) == address(0)) {
+      _setFallbackEntitlementChecker();
+    }
```
shuhuiluo, line 181: Vulnerable to front running.

shuhuiluo (changes requested):

contracts/src/base/registry/facets/checker/EntitlementChecker.sol
giuseppecrj: Don't think we need to do this for now.

shuhuiluo (changes requested):

contracts/src/base/registry/facets/checker/EntitlementChecker.sol
```diff
@@ -162,15 +163,61 @@ contract EntitlementChecker is IEntitlementChecker, Facet {
...
+        XChainLib.Request storage request = layout.requests[transactionId];
+
+        // Skip if already completed or not yet eligible for refund
+        // TODO: make this configurable
+        if (request.completed || block.number - request.blockNumber <= 900)
+          continue;
+
+        totalRefund += request.value;
+        request.completed = true;
+        layout.requestsBySender[msg.sender].remove(transactionId);
```
shuhuiluo, line 189: nit: cache the storage pointer.

shuhuiluo (approved):

PR #2044, opened 2025-01-14
WIP close commit gap

erikolsson:
WIP

### DIFF
--- a/packages/sdk/vitest.config.ts
+++ b/packages/sdk/vitest.config.ts
@@ -1,9 +1,8 @@
 import { defineConfig, mergeConfig } from 'vitest/config'
 import { rootConfig } from '../../vitest.config.mjs'
-import { sdkRootConfig } from './vitest.sdk.rootConfig'
 
 export default mergeConfig(
-    sdkRootConfig,
+    rootConfig,
     defineConfig({
         test: {
             environment: 'happy-dom',


packages/sdk/src/mls/utils/mlsutils.ts
```diff
@@ -1,12 +1,40 @@
...
+    if (payload?.case !== 'memberPayload') {
+        return undefined
+    }
+    if (payload?.value?.content.case !== 'mls') {
+        return undefined
+    }
+
+    const mlsPayload = payload.value.content.value
+    switch (mlsPayload.content.case) {
+        case 'externalJoin':
```
jakzale, line 25: Not sure if this is a valid suggestion, because I haven't tried it yet, but
- if you check if group info and commit are present first, then
- does TypeScript narrow down `content.case` to all cases that have both of them present, so that
- you can switch on it, and then
- check if you handled all the cases with `logNever`?

packages/sdk/src/client.ts
```diff
@@ -2512,6 +2517,44 @@ export class Client
...
+        let commits: Uint8Array[] = []
+        if (!streamView || !streamView.isInitialized) {
+            streamView = await this.getStream(streamId)
+        }
+        commits = mlsCommitsFromStreamView(streamView)
+        check(isDefined(streamView), `stream not found: ${streamId}`)
+        let miniblockNum = streamView.miniblockInfo?.min
+        check(isDefined(miniblockNum), `miniblockNum not found: ${streamId}`)
+
+        function checkDone(commits: Uint8Array[]) {
```
jakzale, line 2531: Minor change: you do not need to pass commits as an argument here.

PR #2046, opened 2025-01-14
Refactor encryption to use AES-GCM, updating related types and methods.

sergekh2:


packages/encryption/src/olmLib.ts
```diff
@@ -14,8 +14,8 @@ export enum EncryptionAlgorithmId {
 export enum GroupEncryptionAlgorithmId {
     // group olm encryption based on signal protocol with ratcheting
     GroupEncryption = 'r.group-encryption.v1.aes-sha2',
-    // asymmetric encryption (olm) to share symmetric encryption (AES) keys
-    HybridGroupEncryption = 'r.hybrid-group-encryption.v1.aes-256-cbc',
+    // asymmetric encryption (olm) to share symmetric encryption (AES-GSM-256) keys
+    HybridGroupEncryption = 'gaes',
```
texuf, line 18: I'd prefer a versioned string here... but for no reason other than thats what ive seen other people do

texuf (commented):
Thank you. Ill get this all in on thursday

packages/encryption/src/olmLib.ts
sergekh2: This string is not for humans, but for code to run switch statement on, but it's persisted as part of each event. It seems that optimal choice in such case to keep it short, but descriptive. I'll update it to `grpaes`

PR #2054, opened 2025-01-15
Media stream optimization: Implement a logic to handle the last media chunk

rhbntt:
When the last media chunk is uploaded, the following steps should be done:
1. Verify miniblocks chain - must be correct
2. Make sure all replicas has a correct chain
3. Populate minipools table since this table is involved in a bunch of queries, for instance, GetMiniblocks
4. Register stream onchain
5. Remove ephemeral label

### DIFF
--- a/core/node/storage/pg_stream_store.go
+++ b/core/node/storage/pg_stream_store.go
@@ -436,7 +436,7 @@ func (s *PostgresStreamStore) CreateStreamStorage(
 		"CreateStreamStorage",
 		pgx.ReadWrite,
 		func(ctx context.Context, tx pgx.Tx) error {
-			return s.createStreamStorageTx(ctx, tx, streamId, genesisMiniblock, false)
+			return s.createStreamStorageTx(ctx, tx, streamId, genesisMiniblock)
 		},
 		nil,
 		"streamId", streamId,
@@ -453,7 +453,7 @@ func (s *PostgresStreamStore) CreateEphemeralStreamStorage(
 		"CreateEphemeralStreamStorage",
 		pgx.ReadWrite,
 		func(ctx context.Context, tx pgx.Tx) error {
-			return s.createStreamStorageTx(ctx, tx, streamId, genesisMiniblock, true)
+			return s.createEphemeralStreamStorageTx(ctx, tx, streamId, genesisMiniblock)
 		},
 		nil,
 		"streamId", streamId,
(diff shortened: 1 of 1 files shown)


core/node/storage/pg_stream_store.go
```diff
@@ -2146,6 +2152,102 @@ func (s *PostgresStreamStore) getLastMiniblockNumberTx(
...
+	tx pgx.Tx,
+	streamId StreamId,
+) (common.Hash, common.Hash, error) {
+	_, err := s.lockEphemeralStream(ctx, tx, streamId, true)
+	if err != nil {
+		return common.Hash{}, common.Hash{}, err
+	}
+
+	// Read all miniblocks for the ephemeral stream.
+	miniblocks := make([]*Miniblock, 0)
```
sergekh2, line 2192: objective of these changes is to enable large memory-efficient media streams, this operation does the reverse: it loads entire stream into memory. 

It's know what is the last mb number from api call (because there is `bool last`).

So for normalize call it's possible to pass in max_mb_num as argument, and then here read all mb num without loading byte content like this:

```sql
SELECT seq_num 
FROM {{miniblocks_m}}
WHERE stream_id = $1
ORDER BY seq_num;
```

then check that 0..max_mb_num all present.

As an extra check it's also possible to read mb 0 and check parameter there (i.e. check implemented below), but I'm not sure it it's required.

core/node/storage/pg_stream_store.go
```diff
@@ -2146,6 +2152,102 @@ func (s *PostgresStreamStore) getLastMiniblockNumberTx(
...
+	// The number of miniblocks must be <chunks-number>+1 where 1 is the genesis miniblock.
+	if inception.GetChunkCount() != int32(len(miniblocks[1:])) {
+		return common.Hash{}, common.Hash{}, RiverError(Err_INTERNAL, "The ephemenral stream can not be normalized due to missing miniblocks")
+	}
+
+	// Update generation in the minipools table
+	_, err = tx.Exec(
+		ctx,
+		s.sqlForStream(
+			"UPDATE {{minipools}} SET generation = $1 WHERE stream_id = $2",
```
sergekh2, line 2226: This query is not going to work as intended: since there are no rows at this point for given stream_id, it's just going to be a no op.

For should be done here is this:
```sql
INSERT INTO {{minipools}} (stream_id, generation, slot_num) VALUES ($1, $2, -1)
```

protocol/protocol.proto
```diff
@@ -765,6 +765,7 @@ message CreationCookie {
   repeated bytes nodes = 2;
   int64 miniblock_num = 3;
   bytes prev_miniblock_hash = 4;
+  bool last = 5;
```
sergekh2, line 768: API structure would be better if `last` is a field on AddMediaEventRequest:
CreationCookie is data-structure that is returned to client and then passed by client back. It's better to pass it back unmodified and have `last` on the request itself, it makes API more "obvious".

core/node/rpc/replicated_add.go
```diff
@@ -86,12 +84,40 @@ func (r *replicatedStream) AddMediaEvent(ctx context.Context, event *ParsedEvent
...
+		}
+
+		// Normalize stream locally
+		genesisMiniblockHash, lastMiniblockHash, err := r.service.storage.NormalizeEphemeralStream(ctx, r.streamId)
+		if err != nil {
+			return err
+		}
+
+		// Register the given stream onchain with sealed flag
+		if err = r.service.streamRegistry.AddStream(
```
sergekh2, line 108: it should only register stream in the contract if quorum of replicas successfully normalized stream (basic principle of distributed storage system: how to guarantee data availability if some replicas fail forever?)

sergekh2 (approved):

PR #2055, opened 2025-01-15
fix(web3): bump `createSpace` gas limit

miguel-nascimento:
Previously, we're getting: `Transaction failed after retries: cannot estimate gas` in the playground app while trying to create a space on omega network due to the default gas limit. 

This PR bumps the gas limit to `50_000`


miguel-nascimento:
Using a fixed gas amount in playground works, but in other places we need the gas estimation. Calling manually `SpaceRegistrar.read.estimateGas[prepaidSpaceFn...]` doesn't fix it 😵‍💫

miguel-nascimento:
there should be a better way to fix this one

PR #2059, opened 2025-01-16
Media stream optimization: Update SDK to use media endpoints 

rhbntt:
Update JS SDK to use media stream specific endpoints

### DIFF
--- a/packages/sdk/src/types.ts
+++ b/packages/sdk/src/types.ts
@@ -162,12 +162,6 @@ export interface ParsedStreamAndCookie {
     events: ParsedEvent[]
 }
 
-export interface ParsedCreationCookie {
-    nextSyncCookie: SyncCookie
-    miniblocks: ParsedMiniblock[]
-    events: ParsedEvent[]
-}
-
 export interface ParsedStreamResponse {
     snapshot: Snapshot
     streamAndCookie: ParsedStreamAndCookie


packages/sdk/src/client.ts
```diff
@@ -782,28 +782,14 @@ export class Client
...
-            unpackedResponse.streamAndCookie.events,
-            unpackedResponse.snapshot,
-            unpackedResponse.streamAndCookie.miniblocks,
-            [],
-            unpackedResponse.prevSnapshotMiniblockNum,
-            undefined,
-            [],
-            undefined,
-        )
+        check(response?.nextCreationCookie !== undefined, 'nextCreationCookie must be defined')
```
texuf, line 790: A "check(...)" is basically an assert. My normal rule is to not assert on data/objects that come from external sources (in this case the response from a remote network call). 

In this case... we're expecting this call to either throw and error or have proper data right? So I guess this is fine. Can you update the error message to be more explicit? Something like `nextCreationCookie was expected but was not returned in response?`

packages/sdk/src/client.ts
```diff
@@ -2285,6 +2270,24 @@ export class Client
...
+        const streamIdStr = streamIdAsString(creationCookie.streamId)
+        check(isDefined(streamIdStr) && streamIdStr !== '', 'streamId must be defined')
+        const event = await makeEvent(this.signerContext, payload, creationCookie.prevMiniblockHash)
+
+        const resp = await this.rpcClient.addMediaEvent({
+            event,
+            creationCookie,
+        })
+
+        check(isDefined(resp.creationCookie), 'creationCookie not found in response')
```
texuf, line 2286: the original makeEventWithHashAndAddToStream has a retry loop... we don't need to retry any more? I guess that makes sense... can you add a doc string with a comment explaining how this function is different than the original?

texuf (approved):

packages/sdk/src/types.ts
```diff
@@ -162,6 +162,12 @@ export interface ParsedStreamAndCookie {
     events: ParsedEvent[]
 }
 
+export interface ParsedCreationCookie {
```
sergekh2, line 165: This interface seems to be not used anymore

sergekh2 (approved):

PR #2071, opened 2025-01-17
Ignore Rust target directory

jakzale:
Running `yarn prettier:check` will give false positives if `mls-tools` was built.


texuf (approved):
Ty - saw this today

PR #2091, opened 2025-01-20
WIP node: add support to recover from missing a stream update event

bas-vk:
It is possible that a node misses the log/receipt of a `setStreamLastMiniblock` transaction due to a restart, RPC provider problem, or ... Currently the node tries to register a new miniblock candidate and succeeds due to relaxed checks in the stream registry contract. With these checks enabled it will go into an endless loop of trying to register a miniblock candidate that fails over and over again.

This change enables the checks in the contract again and adds the ability to the river node to detect and recover from a situation where a mini-block candidate fails to register because it was already registered.

TODO:
- [ ] discuss (test) approach
- [ ] determine if we want to enable 1 of 1 writes again, or keep on 1 of N (currently 50) or query the current config setting and use that in the check.
- [ ] determine if the node needs to loop over all pending miniblock candidates on boot and check if these are registered in the time it was down. The current approach will remediate this situation if a stream has a non-empty minipool which (I think?) is the case with a pending miniblock candidate in its storage.


core/node/events/parsed_event.go
```diff
@@ -33,7 +32,7 @@ func (e *ParsedEvent) GetEnvelopeBytes() ([]byte, error) {
 		Func("GetEnvelopeBytes")
 }
 
-func ParseEvent(envelope *Envelope) (*ParsedEvent, error) {
+func ParseEventWithCachedStreamEvent(envelope *Envelope, streamEvent *StreamEvent) (*ParsedEvent, error) {
```
sergekh2, line 35: changes in this file seem to be not related to the rest of PR?

sergekh2 (approved):

sergekh2:
rebased and submitted in #2428

PR #2096, opened 2025-01-20
Return event refs to added and derived events in AddEventResponse and CreateStreamResponse

texuf:
This is a bookkeeping update that makes it a little more explicit what happened when you call add event/create stream and allows for easier testing of chained events.


clemire (changes requested):
This is very much excellent, can you add some unit test coverage in go? (And maybe TS as well?)

texuf:
> This is very much excellent, can you add some unit test coverage in go? (And maybe TS as well?)

I updated some existing go tests, thanks for the push

sergekh2 (approved):

protocol/protocol.proto
```diff
@@ -882,6 +884,7 @@ message AddEventResponse {
         repeated string funcs = 3;
     }
     Error error = 1; // only set if AddEventRequest.optional is true
+    repeated EventRef new_events = 2;
```
sergekh2, line 887: please add comment here as well

PR #2103, opened 2025-01-21
replace river. in typeName

erikolsson:
I propose we change this, since `Message.getType()` isn't available on all platforms


texuf (commented):
I was afraid of that. Ill make a define

PR #2104, opened 2025-01-21
Media stream optimization: Added checks to AddMediaEvent endpoint

rhbntt:
- Only media stream creator address can add media events
- Chunk index must be within the range 
- Chunk data but not exceed the configured max limit

### DIFF
--- a/core/node/rpc/add_media_event.go
+++ b/core/node/rpc/add_media_event.go
@@ -32,21 +32,21 @@ func (s *Service) localAddMediaEvent(
 		return nil, AsRiverError(err).Func("localAddMediaEvent")
 	}
 
-	genesisStreamEvent, err := s.getGenesisStreamEvent(ctx, streamId)
+	genesisEvent, err := s.getGenesisMediaEvent(ctx, streamId)
 	if err != nil {
 		return nil, AsRiverError(err).Func("localAddMediaEvent")
 	}
 
-	genesisInception := genesisStreamEvent.GetMediaPayload().GetInception()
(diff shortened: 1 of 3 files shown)


core/node/storage/pg_stream_store.go
```diff
@@ -1142,6 +1142,52 @@ func (s *PostgresStreamStore) readMiniblocksByStreamTx(
 	return err
 }
 
+// ReadGenesisMiniblock returns the genesis miniblock for the given stream ID.
+func (s *PostgresStreamStore) ReadGenesisMiniblock(ctx context.Context, streamId StreamId) ([]byte, error) {
```
sergekh2, line 1146: It seems there is not need to add this API: `ReadMiniblocks(ctx, streamId, 0, 1)` gives same result?

core/node/rpc/add_media_event.go
```diff
@@ -29,6 +32,40 @@ func (s *Service) localAddMediaEvent(
 		return nil, AsRiverError(err).Func("localAddMediaEvent")
 	}
 
+	genesisStreamEvent, err := s.getGenesisStreamEvent(ctx, streamId)
+	if err != nil {
+		return nil, AsRiverError(err).Func("localAddMediaEvent")
+	}
+
+	genesisInception := genesisStreamEvent.GetMediaPayload().GetInception()
```
sergekh2, line 40: perhaps as extra check here it should check that this is actually genesis event for media stream, i.e. GetMediaPayload() != nil && GetInception() != nil

sergekh2 (approved):

PR #2105, opened 2025-01-21
Small fixes to the new returned event refs

texuf:
1) follow convention when creating new pointers to slices
2) check for nil on the returned Msg payload (i already checked for error)
3) don’t pre allocate the derived events slice, i don’t actually know how big it will be


bas-vk (approved):
We can try it but I don't think this is the problem.

PR #2112, opened 2025-01-21
Better errors and better retry timings on transactions in tests

texuf:
https://github.com/river-build/river/actions/runs/12894046742/job/35951816490?pr=2097

looks like our chain doesn’t run as fast as it used to? We where having lots of issues with this in the past and it cleared up after we started retrying transacitons. This pr impvoves the error message and adds more time and randomness between retries.

### DIFF
--- a/packages/web3/src/v3/SpaceDapp.ts
+++ b/packages/web3/src/v3/SpaceDapp.ts
@@ -1735,7 +1735,7 @@ async function wrapTransaction(
     txFn: () => Promise<ContractTransaction>,
     txnOpts?: TransactionOpts,
 ): Promise<ContractTransaction> {
-    const retryLimit = txnOpts?.retryCount ?? isTestEnv() ? 3 : 0
+    const retryLimit = txnOpts?.retryCount ?? isTestEnv() ? 5 : 0
 
     const runTx = async () => {
         let retryCount = 0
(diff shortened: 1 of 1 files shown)


texuf:
welp, that wasn't it

clemire:
> https://github.com/river-build/river/actions/runs/12894046742/job/35951816490?pr=2097
> 
> looks like our chain doesn’t run as fast as it used to? We where having lots of issues with this in the past and it cleared up after we started retrying transacitons. This pr impvoves the error message and adds more time and randomness between retries.

Is there a way for us to increase our resource allocation in CI? Because a lot of these tests are stable except for peak hours, which are usually around 3-7pm PST, give or take. That kind of indicates to me that we're getting more resource shares except during times of high contention.

clemire:
> > https://github.com/river-build/river/actions/runs/12894046742/job/35951816490?pr=2097
> > looks like our chain doesn’t run as fast as it used to? We where having lots of issues with this in the past and it cleared up after we started retrying transacitons. This pr impvoves the error message and adds more time and randomness between retries.
> 
> Is there a way for us to increase our resource allocation in CI? Because a lot of these tests are stable except for peak hours, which are usually around 3-7pm PST, give or take. That kind of indicates to me that we're getting more resource shares except during times of high contention.
> 
> I see also below tests are still failing, so I'm wondering if there's a solution in that direction instead.

Honestly I wonder if we should consider moving to another workflow runner like CircleCI.

PR #2113, opened 2025-01-21
allow setting nil encryption algorithm

erikolsson:
this PR allows nil values for MemberPayload.EncryptionAlgorithm in snapshots.

the way we unset the algorithm is that we simply set a nil value:
Client.ts
https://github.com/river-build/river/blob/main/packages/sdk/src/client.ts#L899
Types.ts
https://github.com/river-build/river/blob/main/packages/sdk/src/types.ts#L802-L804
Test:
https://github.com/river-build/river/blob/main/packages/sdk/src/tests/multi_ne/memberMetadata.test.ts#L703-L709

### DIFF
--- a/core/node/events/snapshot.go
+++ b/core/node/events/snapshot.go
@@ -594,9 +594,6 @@ func update_Snapshot_Member(
 		snapshot.Pins = snapPins
 		return nil
 	case *MemberPayload_EncryptionAlgorithm_:
-		if content.EncryptionAlgorithm == nil {
-			return RiverError(Err_INVALID_ARGUMENT, "member payload encryption algorithm not set")
-		}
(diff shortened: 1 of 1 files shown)


core/node/events/snapshot.go
```diff
@@ -594,9 +594,6 @@ func update_Snapshot_Member(
 		snapshot.Pins = snapPins
 		return nil
 	case *MemberPayload_EncryptionAlgorithm_:
-		if content.EncryptionAlgorithm == nil {
-			return RiverError(Err_INVALID_ARGUMENT, "member payload encryption algorithm not set")
-		}
```
texuf, line 599: this check should be inconsequential... the EncryptionAlgorithm payload should always be there via the type system. The .Algorithm property will be nil or defined.

PR #2115, opened 2025-01-22
Replication: implement leader replica proactive reconciliation

sergekh2:
Candidate production protocol is extended to immediately add events missing from local replica,
this smooths miniblock production in case of missing events.

Code is refactored to move `mbJob` into separate file and to re-implement candidate and proposal production functions in the cleaner manner.


core/node/events/miniblock_job.go
```diff
@@ -0,0 +1,291 @@
...
+func (j *mbJob) combineProposals(proposals []*mbProposal) (*mbProposal, error) {
+	// Sanity check: all proposals must have the same miniblock number and prev hash.
+	for _, p := range proposals {
+		if p.newMiniblockNum != proposals[0].newMiniblockNum || p.prevMiniblockHash != proposals[0].prevMiniblockHash {
+			return nil, RiverError(Err_INTERNAL, "mbJob.combineProposals: different miniblock numbers or prev hashes")
+		}
+	}
+
+	// Sanity check: there should be quorum of proposals.
+	quorumNum := TotalQuorumNum(len(j.remoteNodes) + 1)
```
bas-vk, line 183: If the check for enough proposal to reach quorum is done first im the function we end up with 3 iterations over proposals in a row that can be combined into a single loop.

core/node/events/miniblock_job.go
```diff
@@ -0,0 +1,291 @@
...
+
+func (j *mbJob) makeReplicatedProposal(ctx context.Context) (*mbProposal, *streamViewImpl, error) {
+	proposals, view, err := j.processRemoteProposals(ctx)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	localProposal := view.proposeNextMiniblock(ctx, j.params.ChainConfig.Get(), j.forceSnapshot)
+
+	proposals = append(proposals, localProposal)
```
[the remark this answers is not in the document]
sergekh2: proposal for wrong blocks are not sent, and then rejected if received, so this check is extra defensive check for consistency to catch code bugs, i.e. it always should be on the same block at this point in processing, so order of check doesn't matter.

core/node/events/miniblock_job.go
sergekh2: These are sanity checks to make sure invariant is upheld in other parts of the code (this is why error code is Err_INTERNAL).

Readability-wise I like 3 loops more: separates concerns.

PR #2130, opened 2025-01-23
Media stream optimization: Create ephemeral stream if not found

rhbntt:
Replicas should accept ephemeral miniblocks even if there is no ephemeral stream records. If no es record found, create one.

### DIFF
--- a/core/node/storage/pg_stream_store.go
+++ b/core/node/storage/pg_stream_store.go
@@ -2200,13 +2200,7 @@ func (s *PostgresStreamStore) normalizeEphemeralStreamTx(
 	streamId StreamId,
 ) (common.Hash, error) {
 	if _, err := s.lockEphemeralStream(ctx, tx, streamId, true); err != nil {
-		// Ignore a case when an ephemeral stream does not exist. A new record will be inserted.
-		// There might be a situation when a request to create an ephemeral stream failed,
-		// but the rest of ephemeral miniblocks were processed.
-		// In this case, the stream does not exist in the table so a new record must be inserted.
-		if !IsRiverErrorCode(err, Err_NOT_FOUND) {
-			return common.Hash{}, err
-		}
+		return common.Hash{}, err
 	}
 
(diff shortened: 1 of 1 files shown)


core/node/rpc/media_stream_test.go
```diff
@@ -92,7 +92,7 @@ func TestCreateMediaStream(t *testing.T) {
 			mp := events.Make_MediaPayload_Chunk(mediaChunks[i], int32(i))
 			envelope, err := events.MakeEnvelopeWithPayload(alice.wallet, mp, mb)
 			tt.require.NoError(err)
-			fmt.Println(i, i == chunks-1)
```
sergekh2, line 95: btw, there is `testfmt` package to print from tests for debugging. Printing is enabled by env var.

core/node/storage/pg_stream_store.go
```diff
@@ -2199,28 +2199,37 @@ func (s *PostgresStreamStore) normalizeEphemeralStreamTx(
...
 ) (common.Hash, error) {
-	_, err := s.lockEphemeralStream(ctx, tx, streamId, true)
-	if err != nil {
-		return common.Hash{}, err
+	if _, err := s.lockEphemeralStream(ctx, tx, streamId, true); err != nil {
+		// Ignore a case when an ephemeral stream does not exist. A new record will be inserted.
+		// There might be a situation when a request to create an ephemeral stream failed,
+		// but the rest of ephemeral miniblocks were processed.
+		// In this case, the stream does not exist in the table so a new record must be inserted.
+		if !IsRiverErrorCode(err, Err_NOT_FOUND) {
```
sergekh2, line 2207: but in write miniblock function it always inserts a record in es table if it's missing. As such situation described here should never happen? I.e. if there is at least one mb, there will be a record in the es table. Without at least one mb it's impossible to normalize streams anyway.

begmaroman: Totally makes sense, fixed

sergekh2 (approved):

PR #2131, opened 2025-01-23
fix multinode tests on arc runners

mechanical-turk:


.github/workflows/ci.yml
```diff
@@ -203,7 +203,7 @@ jobs:
                   access_token: ${{ github.token }}
 
             - name: Install Binaries
-              run: sudo apt update && sudo apt-get install -y postgresql-client zstd make gcc
+              run: sudo apt update && sudo apt-get install -y postgresql-client zstd make gcc netcat
```
clemire, line 206: netcat kept for ease of debugging?

.github/workflows/ci.yml
```diff
@@ -302,8 +302,10 @@ jobs:
               working-directory: core
               run: just RUN_ENV=multi_ne config-and-start
 
-            - name: Run SDK Tests (without entitlements)
-              run: yarn turbo run test:ci:multi:ne --filter=@river-build/sdk
+            - name: Run River Tests (without entitlements)
+              working-directory: packages/sdk
+              run: yarn run test:ci:multi:ne
+              timeout-minutes: 5
```
clemire, line 308: nice, I'm not sure if this is tight but I like the idea

infra/values.yaml
```diff
@@ -15,11 +15,11 @@ template:
         command: ["/home/runner/run.sh"]
         resources:
           requests:
-            memory: "30Gi"
-            cpu: "10000m"
+            memory: "15.5Gi"
```
clemire, line 18: 💵

clemire (approved):

texuf (approved):

.github/workflows/ci.yml
mechanical-turk: `wait-for` depends on it. native runners come preinstalled with it. you have to explicitly install it on self-hosted runners

.github/workflows/ci.yml
mechanical-turk: it generally wraps up in 2 minutes, we should be fine

PR #2133, opened 2025-01-23
Media stream optimization: Seal ephemeral stream on smart contract event

rhbntt:
Listen for StreamCreated event from the stream registry contract and seal the given stream

### DIFF
--- a/core/node/events/stream_ephemeral.go
+++ b/core/node/events/stream_ephemeral.go
@@ -124,15 +124,22 @@ func (s *streamCacheImpl) normalizeEphemeralStream(
 
 			// Start processing miniblocks from the stream.
 			// If the processing breaks in the middle, the rest of missing miniblocks will be fetched from the next sticky peer.
+			var toNextPeer bool
 			for resp.Receive() {
 				mbInfo, err := NewMiniblockInfoFromProto(resp.Msg().GetMiniblock(), NewParsedMiniblockInfoOpts())
 				if err != nil {
-					return err
+					logging.FromCtx(ctx).Errorw("Failed to parse miniblock info", "err", err, "streamId", stream.streamId)
+					_ = resp.Close()
+					toNextPeer = true
+					break
 				}
 
 				mbBytes, err := mbInfo.ToBytes()
 				if err != nil {
-					return err
+					logging.FromCtx(ctx).Errorw("Failed to serialize miniblock", "err", err, "streamId", stream.streamId)
+					_ = resp.Close()
+					toNextPeer = true
+					break
 				}
 
 				if err = s.params.Storage.WriteEphemeralMiniblock(ctx, stream.streamId, &storage.WriteMiniblockData{
@@ -141,7 +148,10 @@ func (s *streamCacheImpl) normalizeEphemeralStream(
 					Snapshot: mbInfo.IsSnapshot(),
 					Data:     mbBytes,
 				}); err != nil {
-					return err
+					logging.FromCtx(ctx).Errorw("Failed to write miniblock to storage", "err", err, "streamId", stream.streamId)
+					_ = resp.Close()
+					toNextPeer = true
+					break
 				}
 
 				// Delete the processed miniblock from the missingMbs slice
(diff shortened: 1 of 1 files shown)


core/node/events/stream_ephemeral.go
```diff
@@ -0,0 +1,170 @@
...
+	ctx context.Context,
+	event *river.StreamCreated,
+	blockNum crypto.BlockNumber,
+) {
+	if !slices.Contains(event.Stream.Nodes, s.params.Wallet.Address) {
+		return
+	}
+
+	if event.Stream.Flags&uint64(registries.StreamFlagSealed) == 0 {
+		// Stream is node sealed, no need to process it yet.
```
sergekh2, line 29:
```suggestion
		// Stream is not sealed, no need to process it yet.
```

core/node/events/stream_ephemeral.go
```diff
@@ -0,0 +1,170 @@
...
+	stream := &streamImpl{
+		params:              s.params,
+		streamId:            event.GetStreamId(),
+		lastAppliedBlockNum: blockNum,
+		lastAccessedTime:    time.Now(),
+		local:               &localStreamState{},
+	}
+	stream.nodesLocked.Reset(event.Stream.Nodes, s.params.Wallet.Address)
+
+	if err := s.normalizeEphemeralStream(ctx, stream, event); err != nil {
```
sergekh2, line 42: this potentially performs bunch of RPCs, i.e. can take quite awhile.
RPC part shouldn't be run in the loop of onBlockWithLogs, i.e. goroutine should be used.

core/node/storage/pg_stream_store.go
```diff
@@ -1147,6 +1147,117 @@ func (s *PostgresStreamStore) readMiniblocksByStreamTx(
...
+	args := make([]interface{}, len(mbs)+1)
+	args[0] = streamId
+	for i := 1; i <= len(mbs); i++ {
+		// Adding +1 here since the first placeholder is the stream ID
+		placeholders[i] = fmt.Sprintf("$%d", i+1)
+		args[i] = mbs[i-1]
+	}
+
+	query := fmt.Sprintf(
+		"SELECT blockdata, seq_num FROM {{miniblocks}} WHERE stream_id = $1 AND seq_num IN (%s) ORDER BY seq_num",
```
sergekh2, line 1193: it seems it's possible to provide ids without custom string building with something like this:

```go
ids := []int{1, 2, 3}
rows, err := db.Query(ctx, "SELECT * FROM table WHERE id IN (SELECT unnest($1::int[]))", ids)
```

core/node/events/stream_ephemeral.go
```diff
@@ -0,0 +1,182 @@
...
+// normalizeEphemeralStream normalizes the ephemeral stream.
+// Loads the missing miniblocks from the sticky peers and writes them to the storage.
+// Seals the stream if it is ephemeral and all miniblocks are loaded.
+func (s *streamCacheImpl) normalizeEphemeralStream(
+	ctx context.Context,
+	stream *streamImpl,
+	lastMiniblockNum int64,
+	isSealed bool,
+) error {
+	if !isSealed {
```
bas-vk, line 59: Is it possible to early return if the stream isn't a media stream?

core/node/events/stream_ephemeral.go
```diff
@@ -0,0 +1,182 @@
...
+					if v != mbNum {
+						missingMbs[i] = v
+						i++
+					}
+				}
+				missingMbs = missingMbs[:i]
+
+				// No missing miniblocks left, just return.
+				if len(missingMbs) == 0 {
+					return resp.Close()
```
bas-vk, line 160: Maybe just call close and always return nil? All miniblocks are written and if resp.Close() returns an error it doesn't matter?

core/node/events/stream_ephemeral.go
```diff
@@ -0,0 +1,182 @@
...
+				currentStickyPeer = stream.AdvanceStickyPeer(currentStickyPeer)
+				continue
+			}
+
+			// Start processing miniblocks from the stream.
+			// If the processing breaks in the middle, the rest of missing miniblocks will be fetched from the next sticky peer.
+			for resp.Receive() {
+				mbInfo, err := NewMiniblockInfoFromProto(resp.Msg().GetMiniblock(), NewParsedMiniblockInfoOpts())
+				if err != nil {
+					return err
```
bas-vk, line 130: In several places an error is returned. Should the node try the next remote if the current sticky peer is returning an error or garbage?

And in these code paths `resp.Close` isn't called.

sergekh2 (approved):

PR #2136, opened 2025-01-23
Update Turbo package to version 2.3.4, disable update notifier & telemetry collection

sergekh2:


package.json
```diff
@@ -33,6 +31,7 @@
         "prettier:fix": "prettier --write .",
         "syncpack:check": "syncpack list-mismatches",
         "syncpack:fix": "syncpack fix-mismatches && syncpack format",
+        "turbo": "TURBO_NO_UPDATE_NOTIFIER=1 TURBO_TELEMETRY_DISABLED=1 turbo",
```
mechanical-turk, line 34: Sure but what's the expectation? If we care about disabling telemetry on ci we can put these as global env vars. 

If we care about disabling telemetry on local, are you expecting people to run "yarn run turbo x"? Unless we can set this somewhere like turbo.json, this wont do much imo

sergekh2: these setting are not available in turbo.json. Expectation is the same as now: run `yarn turbo` and this will add these setting to turbo.

sergekh2: fair enough that other commands in this script should be updates as well...

texuf:
can we close this?

texuf:
can we close this?

PR #2139, opened 2025-01-23
[RIVER-614][RIVER-2138] Zap error logging, hex encoding of binary

clemire:
Polish zap logging:
- River error logging is now JSON-compatible
   - using error.Log methods will inline the tags as fields on the logger
   - logging the error as a field will emit an unquoted, datadog-parsable JSON object with log message and tags
- fixed length byte arrays and slices are all updated to emit hex-encoded strings
- strings will format as shortened hex if the raw string being logged is a hex string 

Implementation: I copied in zap's json encoder and re-implemented a few specific methods. Unfortunately we could not just extend it with method overrides, as the json encoder has an internal implementation class that makes copies of itself when parsing.


core/node/logging/hex_string.go
clemire, line 1: Old code from the dlog package copied back in for detecting and shortening hex strings

core/node/logging/json_encoder.go
```diff
@@ -0,0 +1,663 @@
...
+	// add remaining
+	appendTo(buf, s[last:])
+}
+
+// Note: this is another place we wrote custom changes.
+// the addFields function looks for protos, Addresses, and byte arrays.
+// It hex-encodes addreses and byte arrays, and marshals the proto into a json-friendly
+// type for logging.
+// All other fields are handled the regular way via field.AddTo(enc).
+var byteType = reflect.TypeOf(byte(0))    // a cached reflect.Type for 'byte'
```
clemire, line 594: Custom changes here

core/node/logging/json_encoder.go
```diff
@@ -0,0 +1,663 @@
...
+	enc.addKey(key)
+	return enc.AppendArray(arr)
+}
+
+func (enc *jsonEncoder) AddObject(key string, obj zapcore.ObjectMarshaler) error {
+	enc.addKey(key)
+	return enc.AppendObject(obj)
+}
+
+// Note: we have updated this method to use hex encoding. The original binary encoding was b64.
```
clemire, line 114: Some custom changes here

core/node/logging/json_encoder.go
clemire, line 1: This file copied in from zap's implementation with a few key changes

clemire: Note to @sergekh2 : please verify licensing text looks good

core/node/logging/log_test.go
```diff
@@ -103,8 +107,19 @@ func TestByteType(t *testing.T) {
 	assert.Contains(buf.String(), "0102030405060708090a")
 }
 
+func TestByteSliceType(t *testing.T) {
```
clemire, line 110: byte slices are handled differently by the encoder than byte arrays, so an extra test was added.

core/node/logging/pool.go
```diff
@@ -0,0 +1,41 @@
+// Copyright (c) 2023 Uber Technologies, Inc. All rights reserved.
```
clemire, line 1: @sergekh2 license text here as well as above, please check

core/node/logging/hex_string.go
```diff
@@ -0,0 +1,66 @@
...
+
+const (
+	shortenHexBytes        = 20
+	shortenHexChars        = shortenHexBytes * 2
+	shortenHexCharsPartLen = shortenHexChars/2 - 2
+)
+
+// formatString will optionally shorten strings if they are determined to be parsable as
+// hex and are past a threshold length.
+func formatString(s string) string {
```
bas-vk, line 52: Do we need this function? I can see the need for this when a byte slice is passed. But when a string is passed should we not just log it instead of trying to solve a caller passing in a too long string issue in a lib?

If this is necessary it can be done without string allocations.

```
func formatString(s string) (string, string, bool) {
   if len(s) > someLimit {
      return s[:cutOff1], s[cutoff2:], true
   }
   return s, "", false
}

And in the caller code:
first, last, truncated := formatString(val)
if truncated {
   enc.AppendTruncatedString(first, last)
} else {
   enc.AppendString(val)
}
```
Add to the json encoder add `AppendTruncatedString(first, last string)` that writes the first string followed by `..` followed by last. This approach doesn't allocate extra strings slices.

clemire: In our dlog implementation, we did detect hex strings and we shortened for both strings and byte slices / arrays. Here, when a byte array or slice is passed, I encode it as hex and add it as a string, so we can intercept hex strings and byte sequences at one spot.

I like your idea of not copying, I will add! TY

clemire: I'm actually fine to take the shortening out honestly if it's preferred by the team...

core/node/base/error.go
```diff
@@ -392,6 +392,20 @@ func (e *RiverErrorImpl) LogLevel(l *zap.SugaredLogger, level zapcore.Level) *Ri
 	return e.LogWithLevel(l, level)
 }
 
+func (e *RiverErrorImpl) MarshalLogObject(enc zapcore.ObjectEncoder) error {
+	var err error
```
sergekh2, line 396: it seems err is never set?

core/node/base/error.go
```diff
@@ -392,6 +392,20 @@ func (e *RiverErrorImpl) LogLevel(l *zap.SugaredLogger, level zapcore.Level) *Ri
...
+	e.ForEachTag(func(name string, value any) bool {
+		zap.Any(name, value).AddTo(enc)
+		return true
+	})
+	if err != nil {
+		return err
+	}
+	
+	enc.AddString("message", e.GetMessage())
```
sergekh2, line 405: lets also add e.Code.String() as "kind" and lets set "message" and "kind" before setting tags.

core/node/logging/hex_string.go
```diff
@@ -0,0 +1,67 @@
...
+	for i := start; i < len(s); i++ {
+		if reverseHexTable[s[i]] > 0x0f {
+			return false, prefix
+		}
+	}
+	return true, prefix
+}
+
+const (
+	shortenHexBytes        = 20
```
sergekh2, line 45: lets set it to 32 so hashes are reported fully.

core/node/logging/json_encoder.go
sergekh2: to easier see what is changed it's possible to push original files in different PR, then rebase this one on top, then it will be clear what changed.

core/node/logging/json_encoder.go
```diff
@@ -0,0 +1,671 @@
...
+func (enc *jsonEncoder) AddBinary(key string, val []byte) {
+	enc.AddString(key, hex.EncodeToString(val))
+}
+
+// Note: we have updated this method to use hex encoding. The original code added the byte
+// string directly.
+func (enc *jsonEncoder) AddByteString(key string, val []byte) {
```
sergekh2, line 121: It's a bit confusing: it seems "ByteString" is a string that is passed as byte array (not 100% sure). In this case should same logic as AddString string be used?

core/node/logging/json_encoder.go
```diff
@@ -0,0 +1,671 @@
...
+	enc.addKey(key)
+	enc.AppendTime(val)
+}
+
+func (enc *jsonEncoder) AddUint64(key string, val uint64) {
+	enc.addKey(key)
+	enc.AppendUint64(val)
+}
+
+func (enc *jsonEncoder) AppendArray(arr zapcore.ArrayMarshaler) error {
```
sergekh2, line 227: any need to modify this one for byte arrays?

core/node/logging/json_encoder.go
clemire: It is used, because we call `AddString` on the hex encoded string.

core/node/base/error.go
clemire: That's right, I missed that in a refactor. Thank you!

core/node/logging/hex_string.go
clemire: NP!

core/node/logging/json_encoder.go
clemire: Original files slightly modified namespaces, etc just to get things to compile: https://github.com/river-build/river/pull/2149

core/node/logging/json_encoder.go
```diff
@@ -106,13 +110,15 @@ func (enc *jsonEncoder) AddObject(key string, obj zapcore.ObjectMarshaler) error
...
-	enc.AddString(key, base64.StdEncoding.EncodeToString(val))
+	enc.AddString(key, hex.EncodeToString(val))
 }
 
+// Note: we have updated this method to use hex encoding. The original code added the byte
+// string directly.
 func (enc *jsonEncoder) AddByteString(key string, val []byte) {
-	enc.addKey(key)
-	enc.AppendByteString(val)
+	enc.AddString(key, hex.EncodeToString(val))
```
sergekh2, line 121: If I understand code correctly, `AddByteString` is used to add slice that contains text. I.e. it should be added as text, not as hex. I.e. there is no need to modify it.

core/node/logging/json_encoder.go
```diff
@@ -577,8 +594,78 @@ func safeAppendStringLike[S []byte | string](
...
 	for i := range fields {
-		fields[i].AddTo(enc)
+		field := fields[i]
+
+		// Use a switch statement here on the field type to avoid
+		// double computation of reflected types.
+		switch (field.Type) { //nolint
+		// common.Addresses and protos come in a a stringer type.
+		case zapcore.StringerType:
+			if address, ok := field.Interface.(common.Address); ok {
```
sergekh2, line 613: I wonder why there is a need to handle it manually: common.Address.String() already returns hex?

core/node/logging/json_encoder.go
```diff
@@ -577,8 +594,78 @@ func safeAppendStringLike[S []byte | string](
...
+			} else if pb, ok := field.Interface.(protoreflect.ProtoMessage); ok {
+				b, err := json.Marshal(pb)
+				// If for some reason json marshalling of the proto fails, just encode
+				// it the default way. It won't be pretty but at least it's in the logs.
+				if err != nil {
+					field.AddTo(enc)
+					continue
+				}
+				var m map[string]interface{}
+				if err := json.Unmarshal(b, &m); err != nil {
```
sergekh2, line 624: this seems a bit round-about, marshaling and then immediately unmarshalling is opposite of effiecient.

(and also, json marshaling is going to do base64 for byte buffers?)

core/node/logging/json_encoder.go
```diff
@@ -577,8 +594,78 @@ func safeAppendStringLike[S []byte | string](
...
+		// double computation of reflected types.
+		switch (field.Type) { //nolint
+		// common.Addresses and protos come in a a stringer type.
+		case zapcore.StringerType:
+			if address, ok := field.Interface.(common.Address); ok {
+				enc.AddString(field.Key, hex.EncodeToString(address[:]))
+			} else if pb, ok := field.Interface.(protoreflect.ProtoMessage); ok {
+				b, err := json.Marshal(pb)
+				// If for some reason json marshalling of the proto fails, just encode
+				// it the default way. It won't be pretty but at least it's in the logs.
```
sergekh2, line 618: I wonder why it's not pretty? because Proto.String is used?

core/node/logging/testdata/envelope_json.txt
```diff
@@ -1 +1 @@
-{"level":"INFO","timestamp":"[TIMESTAMP]","msg":"Logging envelope","envelope":{"hash":\"2346ad27d7568ba9896f1b7da6b5991251debdf2\"}}
+{"level":"INFO","timestamp":"[TIMESTAMP]","msg":"Logging envelope","envelope":{"hash":"MjM0NmFkMjdkNzU2OGJhOTg5NmYxYjdkYTZiNTk5MTI1MWRlYmRmMg=="}}
```
sergekh2, line 1: but this is opposite of what we want

core/node/logging/json_encoder.go
clemire: Yep! I put in this custom handling to avoid that.

core/node/logging/testdata/envelope_json.txt
clemire: This is actually what it used to be with dlog, and we were ok with it. I checked in a change to this file accidentally, but I also disabled this test because it wasn't working right.

But if this is an issue I can definitely work on it.

core/node/logging/json_encoder.go
clemire: The common.Address.String() returned hex is mixed case, but hex.EncodeToString is all lowercase. This was just to preserve behavior captured in unit tests, but I think it's also not needed, I can take it out.

core/node/logging/json_encoder.go
clemire: I don't think that's correct. Text does not go through this entrypoint into the encoder when added via the sugared logger. Furthermore, I think this change was necessary to support hex encoding of byte arrays.

core/node/logging/json_encoder.go
clemire: No, no need. I think we have a unit test that shows that they are working.

core/node/logging/json_encoder.go
sergekh2: But why original implementation has both `AddBinary` and `AddByteString` then?

So original call graph: AddByteString->AppendByteString->safeAddByteString->safeAppendStringLike, compare to 
AddString->AppendString->safeAddString->safeAppendStringLike.

Which I read as AddByteString is a convenience method to marshal string that happen to be in byte slice. 

Please double check that it's required for byte arrays to work correctly.

core/node/logging/testdata/envelope_json.txt
sergekh2: it was this way, but it was practically unusable in json;
but different formatter was used for text, so it text it was still usable (i.e. for local debugging). 

Anyway, maybe lets drop protos from this PR and we can take a stub at them separately?

core/node/logging/json_encoder.go
sergekh2: General philosophy for zap is to be performant and not allocate if can be avoided, and this is a bit reverse of that. 

Do protos go through 'AddObject'? I.e. is there a need to special-case them at all?

Anyway, I suggest to drop protos from this PR and then we can see what handling they need separately.

core/node/logging/json_encoder.go
clemire: You're right, I must have made a mistake when testing / tracing when I implemented this. I took it out!

core/node/logging/json_encoder.go
clemire: (and also, json marshaling is going to do base64 for byte buffers?)

The go json marshaller does appear to use b64 actually, you can see some evidence of this in a test case I added if you unskip it.

clemire: Removed proto code!

sergekh2 (approved):

PR #2140, opened 2025-01-23
Reduce the largish log volume for a small set of corrupt streams we've deemed not worthy of repair

clemire:
There are about 30 streams that were accidentally created when some nodes were running under their operator addresses, and these streams will consistently come up as unable to archive since the node the stream was allocated, but these errors are uninformative. They come up repeatedly because we retry stuff to self-heal the corrupted streams list, so let's just ignore this error type altogether.


mechanical-turk (approved):

bas-vk (approved):
If you want to be complete, the notification service also hits several of those streams and prints a warning in the logging. Search for `Ignore stream, no valid node found` and drop the log statement there or reduce it to debug logging.

PR #2144, opened 2025-01-24
skip the sdk timeline scrollback test

texuf:
It can be flaky if it runs too fast, all the events end up in the same bock.
Force snapshots so that when alice calls get stream she is only returned the last few events.


clemire (approved):

texuf:
welp, that didn't fix it.

texuf:
skipping for now

PR #2146, opened 2025-01-24
feat(contract) executor facet for third-party space transactions

giuseppecrj:
Executor Flow


nkrishang:
Hi @giuseppecrj 👋🏽  I've read through the changed files and the following is what I've understood. Please correct any misunderstanding.

---

The job of the main function `execute`:
```solidity
function execute(
    address target,
    bytes calldata data
  ) external payable returns (uint32 nonce)
```

is to perform a regular call on the provided `target`, with the provided `data`.
```solidity
Address.functionCallWithValue(target, data, msg.value);
```

Each target maps 4 byte `selectors` to a Group with a `groupId`. At a high level, each group stores whether a given account has "access" to the group. To execute a call a particular function (selector) on a target, the caller (account) must have access to the group associated with that (target, selector) pair.

This is the _core_ flow. There are things like scheduling calls, ordering them based on nonces, guardians and time delays related to access/permissions built around this core flow.

---

Based on the above understanding, the considerations listed below are the main "missing things" that come to mind. Let me know if the contract already accounts for these via some clever usage.

1. **Batch of calls**: 
    Currently, each execution is a pair of (target, selector (implicitly), data). There may be actions that require a batch of ordered calls to execute within the same transaction.
    Technically, the contract does allow such ordered calls to be scheduled one after another, but I think the right abstraction would be to treat a batch of calls as a single unit.

2. **Validation hooks**:  
    Even though call requests are scheduled by an authorized actor, it'll be useful to associate function selectors with:
    - "before" validation hooks to ensure calls are executed within a desired context, and 
    - "after" validation hooks to ensure the executed calls created the desired / expected state changes in the Space or external smart contracts.
    
    The actual abstraction for these hooks can be kept flexible (e.g. associate selector with a validation target address and call selector) or fixed (e.g. well defined _before_ and _after_ validation hook structs to associate with a selector, and a well defined hook function interface that validation smart contracts must implement).

giuseppecrj:
@nkrishang 

Thanks for looking at the code.

1. We'll add batched calls
2. Yep, hooks are in the works
3. If a group has a delay of 0 they're able to make executions without having to schedule. This feature allows owners or guardians the ability to opt-out of malicious calls from bots or apps. On your end a space owner will just sign a transaction allowing the app calls with no delay.

shuhuiluo:
@coderabbitai summary

PR #2151, opened 2025-01-25
Update Xchain to use entitlement coordinator, establish tests for using V1 and V2 check events

clemire:
Branched from G's entitlement coordinator PR here, this PR updates the xchain service to respond to both v1 and v2 checks. The mock entitlement checker is also updated to include xchain methods so the final event will be emitted and the client simulator library can detect it and validate that the check reached quorum.


contracts/test/mocks/MockEntitlementGated.sol
```diff
@@ -70,6 +70,28 @@ contract MockEntitlementGated is EntitlementGated {
     return transactionId;
   }
 
+  function requestLegacyEntitlementCheckV3(
```
clemire, line 73: This entrypoint to the mock entitlement gated contract allows us to emit v2 check events with v1 rule datas.

core/config/config.go
```diff
@@ -169,13 +169,6 @@ type Config struct {
 
 	ChainConfigs map[uint64]*ChainConfig `mapstructure:"-"` // This is a derived field from Chains.
 
-	// extra xChain configuration
-	EntitlementContract     ContractConfig `mapstructure:"entitlement_contract"`
-	TestEntitlementContract ContractConfig `mapstructure:"test_contract"`
```
clemire, line 174: This test config is something we've been meaning to remove from the node config for a long time, I just went ahead and took care of it here because it's a small change.

sergekh2 (approved):

PR #2194, opened 2025-01-29
StreamRegistry: support allocating stream without genesis block

bas-vk:
Currently the node passes the protobuf serialised genesis block when allocating a stream to streams registry. As part of migrating to replicated streams the node will stop doing this and nodes can request the genesis block from one of the replicas if it needs it. This saves on calldata and storage costs.

This PR adds 2 functions:
- new overload of `AllocateStream` that doesn't accept the genesis block bytes
- `getStreamWithGenesisHash` as a future replacement for `getStreamWithGenesis`

Follow-up changes after this is deployed:
- migrate the node to move away from registering the genesis block with the 2 newly introduces functions. This will be a 2 step process, first ensure all nodes are able to deal with "missing" genesis block and migrate nodes to strop writing genesis block.
- drop the 2 old functions that accept/return the genesis block from the stream facet and mark the storage slot for genesis block data as used/freed/...?


sergekh2 (approved):

shuhuiluo (approved):

shuhuiluo:
Can we close it?

PR #2206, opened 2025-01-30
don't notify people who only reacted to the parent event

### LINKED ISSUE #2197: Getting unexpected notification on Mobile that someone replied to me
Getting a notification on mobile that someone replied to me on a message that I reacted to in the chat.
Expected behaviour is to not receive such notification.

erikolsson:
Fixes #2197


bas-vk (approved):

texuf (changes requested):
i think the tags payload is right and the text is wrong

texuf (approved):

PR #2207, opened 2025-01-30
Media stream optimization: Delete "dead" ephemeral streams

rhbntt:
A media uploading process might fail in the middle. In this case, the node storage would contain "uncompleted" ephemeral stream which has no value. It should be deleted if there were no updates for N time.

### DIFF
--- a/core/node/events/stream_ephemeral.go
+++ b/core/node/events/stream_ephemeral.go
@@ -64,7 +64,7 @@ func (s *StreamCache) normalizeEphemeralStream(
 
 	// Sealed stream has 0 in the last miniblock parameter so the given value should be fetched from the genesis
 	// TODO: Either store genesis miniblock data on chain or fetch it from replicas. Discuss.
-	if lastMiniblockNum == 0 {
+	if lastMiniblockNum == 0 && false {
 		_, _, blockdata, _, err := s.params.Registry.GetStreamWithGenesis(ctx, stream.StreamId())
 		if err != nil {
 			return err


core/node/rpc/dead_ephemeral_stream_monitor.go
```diff
@@ -0,0 +1,102 @@
...
+	// Start the dead stream monitor.
+	go m.monitor(ctx)
+
+	return nil
+}
+
+// monitor is the main loop of the dead ephemeral stream clean up procedure.
+func (m *deadEphemeralStreamMonitor) monitor(ctx context.Context) {
+	const (
+		cleanupInterval = time.Minute * 5
```
bas-vk, line 39: Should `cleanupInterval` not be something like 1 minute?

core/node/rpc/dead_ephemeral_stream_monitor.go
```diff
@@ -0,0 +1,102 @@
...
+			m.Unlock()
+		}
+	}
+
+}
+
+// onUpdated is called when a stream is updated, e.g. new ephemeral miniblock was added.
+func (m *deadEphemeralStreamMonitor) onUpdated(streamId StreamId) {
+	m.Lock()
+	m.ephemeralStreams[streamId] = time.Now()
```
bas-vk, line 74: Would it make sense to store the first time an ephemeral stream was updated and check that it is sealed within lets say 5 minutes instead of detecting that a stream is dead after it hasn't got an update in 5 minutes?

core/node/rpc/ephemeral_stream_monitor.go
```diff
@@ -0,0 +1,105 @@
...
+// ephemeralStreamMonitor is a monitor that keeps track of ephemeral streams and cleans up dead ones.
+type ephemeralStreamMonitor struct {
+	// ephemeralStreams is a map of ephemeral stream IDs to the time they were last updated.
+	ephemeralStreams *xsync.MapOf[StreamId, time.Time]
+
+	storage storage.StreamStorage
+}
+
+// initEphemeralStreamMonitor creates and starts a dead ephemeral stream monitor.
+func initEphemeralStreamMonitor(
```
sergekh2, line 24:
```suggestion
func newEphemeralStreamMonitor(
```

core/node/rpc/ephemeral_stream_monitor.go
```diff
@@ -0,0 +1,105 @@
...
+		select {
+		case <-ctx.Done():
+			if err := ctx.Err(); !errors.Is(err, context.Canceled) {
+				logging.FromCtx(ctx).Error("dead ephemeral stream monitor stopped", "err", err)
+			}
+			return
+		case <-ticker.C:
+			m.ephemeralStreams.Range(func(streamId StreamId, lastUpdated time.Time) bool {
+				if time.Since(lastUpdated) > ttl {
+					if err := m.storage.DeleteEphemeralStream(ctx, streamId); err != nil {
```
sergekh2, line 64: but if stream already is not ephemeral, this will produce error here as well.
I suggest calling `m.ephemeralStreams.Delete(streamId)` unconditionally (and not retyring) - in the worst case stream will be deleted on the next restart

core/node/storage/pg_ephemeral_store_monitor.go
```diff
@@ -0,0 +1,165 @@
...
+			logging.FromCtx(ctx).Info("dead ephemeral stream monitor stopped")
+			return
+		case <-ctx.Done():
+			if err := ctx.Err(); !errors.Is(err, context.Canceled) {
+				logging.FromCtx(ctx).Error("dead ephemeral stream monitor stopped", "err", err)
+			}
+			m.close()
+			return
+		case <-ticker.C:
+			m.streams.Range(func(streamId StreamId, createdAt time.Time) bool {
```
sergekh2, line 94: As code readability goes, it's better to put longer lambdas into separate methods

sergekh2 (approved):

sergekh2 (approved):

PR #2234, opened 2025-02-03
[2252] Auth refactor - share auth functionality between notification and bots registry

clemire:
Also, add authentication to the bot registry service, with a test.


texuf (approved):

core/node/authentication/service.go
```diff
@@ -89,15 +91,68 @@ func (c authenticationChallenge) Verify(
...
+	sessionTokenSigningAlgo       string
+	pendingAuthenticationRequests sync.Map
+	challengePrefix               string
+}
+
+func (s *AuthServiceMixin) ShortServiceName() string {
+	return strings.ToLower(s.challengePrefix[:2])
+}
+
+func (s *AuthServiceMixin) Init(challengePrefix string, config *config.AuthenticationConfig) error {
```
bas-vk, line 109: Lets call this `InitAuthentication`.

bas-vk:
Code looks good but would like to hear what the future architecture looks like to validate that some design chooses are correct.

core/node/bot_registry/service.go
```diff
@@ -54,8 +66,33 @@ func (s *Service) RegisterWebhook(
 			Tag("bot_owner_id", req.Msg.BotOwnerId)
 	}
 
-	// TODO: authorization
-	// auth signer should match owner or bot address
+	rawUserId := ctx.Value(authentication.UserIDCtxKey{})
```
bas-vk, line 69: We can introduce `authentication.UserIDFromContext(context.Context)` and have that function return the userID and a bool indicating that is was found instead of getting it here and casting it to an address everywhere the user id is needed. `UserIDCtxKey` can be made private after that.

bas-vk (approved):
Good job.

core/node/bot_registry/service.go
clemire: done!

PR #2250, opened 2025-02-05
[BUGFIX] Double auth_impl entitlements timeout

clemire:
Shillr is gated by a role with a rule data that has 43 token's OR'ed together, and calls to AddEvent for membership events to add members to channels were sometimes timing out on the node. Investigation showed that these timeouts often occur 5-6s into an AddEvent call, and that the chain_auth limits the duration of the segment of code that makes contracts queries in order to evaluate an entitlement to a maximum of 5s by default. Furthermore, I wrote a CLI tool to test evaluating channel entitlements locally and I see that this Shillr channel's role entitlements can sometimes can take >5s to evaluate.

### DIFF
--- a/core/node/auth/auth_impl.go
+++ b/core/node/auth/auth_impl.go
@@ -169,7 +169,7 @@ func newArgsForLinkedWallets(principal common.Address) *ChainAuthArgs {
 }
 
 const (
-	DEFAULT_REQUEST_TIMEOUT_MS = 5000
+	DEFAULT_REQUEST_TIMEOUT_MS = 10000
 	DEFAULT_MAX_WALLETS        = 10
 )
 


sergekh2 (approved):

jterzis:
We still have the potential for evals to timeout at 10 seconds if the rule evaluation list for the entitlement is large enough ? Any reason why we don't set the timeout to the default max set by rpc nodes (geth), which I believe is 30s.

clemire:
> We still have the potential for evals to timeout at 10 seconds if the rule evaluation list for the entitlement is large enough ? Any reason why we don't set the timeout to the default max set by rpc nodes (geth), which I believe is 30s.

We could go up to 20s, but `AddEvent` maximum timeout is 20s so I think probably not any reason to go above that.

PR #2257, opened 2025-02-05
node: quorum pool return immediately once quorum is, or can't be reached

bas-vk:
Return immediately in `QuorumPool#Wait` when quorum is reached.


sergekh2 (approved):

core/node/rpc/replicated_add.go
```diff
@@ -208,30 +237,45 @@ func (s *Service) replicatedAddMediaEventImpl(ctx context.Context, event *Parsed
...
+		quorumCheckMu.Lock()
+		streamSuccessCount++
+		quorumCheckMu.Unlock()
+
 		return nil
 	})
 
-	if err = sender.Wait(); err != nil {
+	if err = quorum.Wait(); err != nil {
+		fmt.Println("wait error", err)
```
sergekh2, line 266: should be a log statement?

core/node/rpc/replicated_add.go
```diff
@@ -208,30 +237,45 @@ func (s *Service) replicatedAddMediaEventImpl(ctx context.Context, event *Parsed
...
 			),
-		); err != nil {
+		)
+
+		if err != nil {
 			return err
 		}
 
+		quorumCheckMu.Lock()
+		streamSuccessCount++
```
sergekh2, line 259: if non-nil hash is returned, let's assign it here as well: then the process will complete even if local has failed. (i.e. at least one good hash is required among all replicas)

sergekh2 (approved):

core/node/events/quorum_pool.go
```diff
@@ -3,122 +3,171 @@ package events
...
-				success++
-				if success >= remoteQuorum {
-					return nil
-				}
-			} else {
-				errs = append(errs, err)
-				if len(errs) > q.remotes-remoteQuorum {
-					return RiverErrorWithBases(Err_QUORUM_FAILED, "quorum failed", errs, "remotes", q.remotes, "remoteQuorum", remoteQuorum, "failed", len(errs), "succeeded", success)
-				}
+		if q.externalQuorumCheck == nil && len(q.errors) >= quorumNum { // not able to achieve quorum anymore
```
sergekh2, line 133:
```suggestion
		if q.externalQuorumCheck == nil && len(q.errors) > (q.totalTasks - quorumNum) { // not able to achieve quorum anymore
```

PR #2258, opened 2025-02-05
Fix jest.env.ts path

rbhnt:


texuf (changes requested):
I think this goes up one level too far. Can you just delete the top level directory?

texuf (approved):

clemire (approved):

packages/stress/src/utils/stressClient.ts
```diff
@@ -5,12 +5,7 @@ import {
     SyncAgent,
     spaceIdFromChannelId,
 } from '@river-build/sdk'
-import {
-    ExtendedInboundGroupSessionData,
-    GroupSessionRecord,
-    HybridGroupSessionRecord,
-    type ExportedDevice,
-} from '@river-build/encryption'
```
texuf, line 13: sorry about that, was late, should have waited.

PR #2274, opened 2025-02-06
Media stream optimization: Introduce feature flag to sdk

rhbntt:


packages/sdk/src/tests/multi_ne/media.test.ts
```diff
@@ -35,6 +35,318 @@ describe('mediaTests', () => {
         return bobsClient.createMediaStream(channelId, spaceId, undefined, chunkCount)
     }
 
+    async function bobSendMediaPayloads(
```
texuf, line 38: these are cut and paste right?

instead of making `createMediaStreamNew` can you just add a param to `createMediaStream(..., opts: { streamVersion: 'v1' | 'v2' }`

then use the `.each` test method to run these tests on each type? https://github.com/towns-protocol/towns/blob/ea932e94b30f061b68f92119e82ec83d4804c616/packages/sdk/src/tests/multi_ne/userInboxMessage.test.ts#L29

PR #2278, opened 2025-02-07
Stop saving the snapshots in the stream view, we weren't using it for anything and it takes loads of memory in the browser

texuf:
loads of memory
not using it for anything


texuf:
Actually that statement is a lie - the snapshots are already in the timeline, i still need to strip them

ttillberg (approved):
Like, how much?

texuf:
when i do it right, by half 
image

PR #2280, opened 2025-02-07
Refactor notification stream syncing code for tandem use by bot service

clemire:
The actual streams tracker will be sufficiently different for each service that it makes sense to go ahead and implement them separately: the notification service cares about all DM, GDM, and channel streams, and tracks all of these, but the bot registry service only cares about streams associated with a registered bot, and the channels in the spaces that bot belongs to. Furthermore, channel streams are sometimes picked up on allocation by the notification service, but they are almost always going to be added retroactively for the bot registry service at the point at which we detect a bot has been added to a channel.

However, the logic for keeping an up-to-date internal copy of the view of each stream, and for syncing that stream, are very similar. This PR refactors those common elements into a shared library and plugs them back into the notification service.


core/node/events/tracked_stream_view.go
```diff
@@ -0,0 +1,114 @@
...
+// relevant streams. It is essentially a wrapper around StreamView, to apply events, and to execute callbacks.
+// OnViewLoaded is executed upon TrackedStreamView creation when the constructed view is fully reified.
+// onNewEvent is called whenever a new event is added to the view, ensuring that the onNewEvent callback is
+// never called twice for the same event.
+func NewTrackedStreamView(
+	ctx context.Context,
+	streamID shared.StreamId,
+	cfg crypto.OnChainConfiguration,
+	stream *StreamAndCookie,
+	onViewLoaded func(view *StreamView) error,
```
bas-vk, line 38: Instead of passing the `onViewLoaded` callback can't this function return the view and let the caller initialise something?

bas-vk:
One thing to note is that currently the notification service misses events during downtime. There is still a wish to make the sync cookies persistent so the notification service can resume processing events after it resumed working. This is probably also required for other services.

bas-vk (approved):

core/node/events/tracked_stream_view.go
clemire: On the view itself? I was thinking of the view as a read-only representation of the state of the stream, to me that does not make sense. Can you clarify?

clemire:
> One thing to note is that currently the notification service misses events during downtime. There is still a wish to make the sync cookies persistent so the notification service can resume processing events after it resumed working. This is probably also required for other services.

Oof, we probably don't want to miss events for the bot registry service.

core/node/events/tracked_stream_view.go
clemire: Ok I read this again and it makes sense. I'll refactor.

PR #2285, opened 2025-02-07
notification: send notifications in blocks received on sync reset for…

bas-vk:
… new allocated streams

### DIFF
--- a/core/node/notifications/sync/streams_tracker_worker_connectgo.go
+++ b/core/node/notifications/sync/streams_tracker_worker_connectgo.go
@@ -253,6 +253,7 @@ func (s *StreamTrackerConnectGo) Run(
 				if reset {
 					trackedStream, err = events.NewNotificationsStreamTrackerFromStreamAndCookie(
 						syncCtx, streamID, onChainConfig, update.GetStream(), listener, userPreferences)
+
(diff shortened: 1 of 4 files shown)


core/node/notifications/sync/streams_tracker_worker_connectgo.go
```diff
@@ -276,11 +277,19 @@ func (s *StreamTrackerConnectGo) Run(
...
-						log.Errorw("Unable to apply block", "stream", streamID, "err", err)
+					if !reset {
+						if err := trackedStream.ApplyBlock(block, onChainConfig.Get()); err != nil {
+							log.Debugw("Unable to apply block", "stream", streamID, "err", err)
+						}
+					}
+					if justAllocated {
+						// send notifications for all events in all blocks
+						for _, event := range block.GetEvents() {
+							if parsedEvent, err := events.ParseEvent(event); err == nil {
```
clemire, line 289: @bas-vk, is there a bug here where memberships from the current block are applied before the events are sent instead of after?

clemire: Ah Austin is clarifying, nevermind

PR #2271, opened 2025-02-06
remove mls-rs-wasm

erikolsson:

PR #2301, opened 2025-02-08
Remove MLS from ts packages

sergekh2:


texuf:
@erikolsson has a pr for this here https://github.com/towns-protocol/towns/pull/2271

sergekh2:
> @erikolsson has a pr for this here #2271

I guess it's a bit of dup then 🤷 
I have dependent pr for go, so I'll merge this one if it passes

PR #2313, opened 2025-02-10
Fix triggering stream reconciliation if lapse is detected during mb production

sergekh2:
Error code check didn't work since returned error was ConnectError.
Also schedule reconciliation on the correct pool

### DIFF
--- a/core/node/base/error.go
+++ b/core/node/base/error.go
@@ -234,10 +234,7 @@ func IsRiverError(err error) bool {
 }
 
 func IsRiverErrorCode(err error, code protocol.Err) bool {
-	if e, ok := err.(*RiverErrorImpl); ok {
-		return e.Code == code
-	}
-	return false
+	return AsRiverError(err).Code == code
 }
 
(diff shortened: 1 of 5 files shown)


core/node/base/error.go
```diff
@@ -234,10 +234,7 @@ func IsRiverError(err error) bool {
 }
 
 func IsRiverErrorCode(err error, code protocol.Err) bool {
-	if e, ok := err.(*RiverErrorImpl); ok {
-		return e.Code == code
-	}
-	return false
```
texuf, line 240: we've had a few of these bad error code checks in the last week yeah?

texuf: or have i just seen this pr a few times?

texuf (approved):

core/node/base/error.go
sergekh2: this changes logic a bit: now it also converts ConnectError back to RiverError so code comparison actually works correctly. Previous version only worked if error was already converted.

bas-vk (approved):

PR #2334, opened 2025-02-11
Omit alchemy secret from repo

clemire:
I already deleted the app and rotated my usage of alchemy URLs, but I'm deleting this from the repo as well for hygienic reasons.

### DIFF
--- a/packages/xchain-monitor/docker-compose.yml
+++ b/packages/xchain-monitor/docker-compose.yml
@@ -5,6 +5,6 @@ services:
             dockerfile: ./packages/xchain-monitor/Dockerfile
         environment:
             RIVER_ENV: omega
-            INITIAL_BLOCK_NUM: 25287260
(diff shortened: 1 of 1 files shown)


packages/xchain-monitor/docker-compose.yml
```diff
@@ -5,6 +5,6 @@ services:
             dockerfile: ./packages/xchain-monitor/Dockerfile
         environment:
             RIVER_ENV: omega
-            INITIAL_BLOCK_NUM: 25287260
+            INITIAL_BLOCK_NUM: 26051573
```
texuf, line 8: Why did initial block num change?

clemire: I think this setting is a bit arbitrary, but good point, no need to change it.

texuf (approved):

PR #2341, opened 2025-02-11
create space v2

giuseppecrj:


contracts/src/factory/facets/create/CreateSpace.sol
```diff
@@ -23,18 +29,31 @@ contract CreateSpaceFacet is
     _addInterface(type(ICreateSpace).interfaceId);
   }
 
+  /// @inheritdoc ICreateSpace
+  function createSpaceV2(
```
shuhuiluo, line 33: Consider `createSpaceGeneric` with an action enum and `bytes calldata` to avoid further fragmentation.

shuhuiluo (changes requested):

contracts/src/factory/facets/create/CreateSpace.sol
giuseppecrj: let's do another time, this is blocking development

shuhuiluo (approved):

PR #2350, opened 2025-02-12
Block clients from uploading key solicitations with empty session ids

texuf:
Sadly some clients are requesting empty session ids over and over (tracked general down to caroline’s client)

Need to track down the actual cause, this will prevent the streams from blowing up.


texuf:
@erikolsson @andreyz I am seeing loads of repeated solicitations in streams with just [""] as the session id. I don't know where it's coming from yet, maybe mobile, maybe web, so i expect this change to potentially break something on there... but we need to stop polluting the streams.

PR #2359, opened 2025-02-12
pause gas metering for helper contract deployment to prevent oog errors

jterzis:
I added try..catch blocks and caught the error we were seeing in [CI](https://github.com/HereNotThere/harmony/actions/runs/13295919423/job/37127735707) with more context. The script through out of gas evm error when deploying the spaceOwner contract helper. Pausing gas metering around the helper contract deployments in the constructor allowed me to run the InteractAlphaSparse script locally without any error.


shuhuiluo (approved):
I wonder why the deployer helpers don't revert in unit tests.

PR #2374, opened 2025-02-14
Bot registry-to-bot-service authentication

clemire:
This PR does the following:
- Split out rpc calls and logic for registration and web hook registration 
- Generate shared secrets for each bot upon registration and store in pg encrypted by a data encryption key that can be set with config / env variable
- Write / Read shared secrets to/from pg
- Implement jwt token generation for web hook calls using the shared secret generated by the `Register` rpc call
- Implement test bot server stub as a test utility that validates proper authorization from the caller
- Update tests to call the web hook with the auth token added to the header when webhooks are registered


core/config/config.go
```diff
@@ -438,6 +438,16 @@ type BotRegistryConfig struct {
...
 	// Authentication holds configuration for the Client API authentication service.
 	Authentication AuthenticationConfig
+
+	// SharedSecretDataEncryptionKey stores the 256-bit key used to encrypt shared secrets in database
+	// storage via AES256. This key is stored as a string in hex format, with an expected length of 64
+	// characters, plus an optional '0x' prefix.
+	SharedSecretDataEncryptionKey string `json:"-" yaml:"-"` // Omit sensitive field from logging
+
+	// AllowLoopbackWebhooks allows webhooks that resolve to a loopback address via DNS. This setting
+	// was added for local/unit testing only and is highly discouraged to enable for production environments,
```
texuf, line 448:
```suggestion
	// was added for local/unit testing only and should not be used in production environments,
```

protocol/bots.proto
```diff
@@ -10,23 +10,47 @@ option go_package = "github.com/towns-protocol/towns/core/node/protocol";
 // This session token can be obtained from the AuthenticationService. If the session token is missing or invalid an
 // Err_UNAUTHENTICATED (code=16) is returned.
 service BotRegistryService {
+    rpc Register(RegisterRequest) returns (RegisterResponse);
     rpc RegisterWebhook(RegisterWebhookRequest) returns (RegisterWebhookResponse);
+    // rpc GetRegisteredBotInfo(GetRegisteredBotInfoRequest) returns (GetRegisteredBotInfoResponse);
```
texuf, line 15: I know this is commented out, but is there a way to get unregistered bot info? If not this should just be GetBotInfo, or even just GetInfo, since this is the bot registry service

texuf (approved):

PR #2376, opened 2025-02-14
Media stream optimization: Per chunk encryption setting

rhbntt:
Add a stream setting that marks a media stream as per chunk encrypted. It means that each individual chunk is encrypted separately before uploading rather than encrypting the full blob and splitting into chunks.


packages/encryption/src/cryptoAesGcm.ts
```diff
@@ -14,12 +14,20 @@ export async function importAesGsmKeyBytes(key: Uint8Array): Promise<CryptoKey>
 export async function encryptAesGcm(
     key: CryptoKey,
     data: Uint8Array,
+): Promise<{ ciphertext: Uint8Array; iv: Uint8Array }> {
+    const iv = crypto.getRandomValues(new Uint8Array(12))
+    return encryptAesGcmWithIv(key, data, iv)
+}
+
+export async function encryptAesGcmWithIv(
```
sergekh2, line 22: but what is the use case?
IV should never be reused, so why this API?

protocol/protocol.proto
```diff
@@ -669,6 +669,11 @@ message EventRef {
 message StreamSettings {
     // Test setting for testing with manual miniblock creation through Info debug request. 
     bool disable_miniblock_creation = 1;
+
+    // Setting for enabling per chunk encryption for media payloads.
+    // Meaning, the given media stream has each chunk individually encrypted
+    // rather than encrypting the full data and splitting into chunks.
+    bool per_chunk_encryption = 2;
```
sergekh2, line 676: Lets add it to MediaPayload.Inception message: it's specific to media streams only.

texuf (approved):

PR #2377, opened 2025-02-14
Fixed test_revertWhen_addStream_nodesNotRegistered flaky test

rhbntt:


giuseppecrj (approved):

clemire (approved):
I think this line should be there to ensure that randomNode is not registered, and at the very least doesn't hurt.

shuhuiluo (approved):

PR #2395, opened 2025-02-18
Space Review to nodes and back

texuf:
upload space review to the user stream, verify it, forward it to the space stream
tip it
verify we can see tips


packages/web3/src/v3/IReviewShim.ts
```diff
@@ -0,0 +1,185 @@
...
+    public async addReview(
+        rating: number,
+        comment: string,
+        signer: ethers.Signer,
+    ): Promise<ContractTransaction> {
+        // Encode the struct [string, uint8] into bytes
+        const encodedData = ethers.utils.defaultAbiCoder.encode(
+            ['string', 'uint8'],
+            [comment, rating],
+        )
```
shuhuiluo, line 61: It should be encoded as `ReviewStorage.Content` which is `tuple(string,uint8)`. Please try
```typescript
        const encodedData = ethers.utils.defaultAbiCoder.encode(
            ['tuple(string,uint8)'],
            [[comment, rating]],
        )
```
or use viem because the typechain struct doesn't have convenience methods.

shuhuiluo (dismissed):

shuhuiluo (approved):

PR #2404, opened 2025-02-19
Space Review Web3 shims and components

texuf:
with half of the transaction test to verify it’s working and the events are comming back in the receipts


packages/web3/src/v3/IReviewShim.ts
```diff
@@ -0,0 +1,130 @@
...
+    ReviewStorage,
+} from '@river-build/generated/dev/typings/IReview'
+
+import { ContractTransaction, ethers } from 'ethers'
+import { BaseContractShim } from './BaseContractShim'
+
+import DevAbi from '@river-build/generated/dev/abis/IReview.abi.json' assert { type: 'json' }
+import { Address } from 'abitype'
+
+// solidity doesn't export enums, so we need to define them here, boooooo
```
shuhuiluo, line 13: That's not correct.

shuhuiluo (approved):

PR #2406, opened 2025-02-19
add support for token transfers

erikolsson:
this PR adds:
- protocol slot for `TokenTransfer` transactions
- validation of EVM txs
- validation of Solana txs


protocol/protocol.proto
```diff
@@ -627,11 +627,21 @@ message BlockchainTransaction {
         Event event = 1; // event emitted by the tipping facet
         bytes toUserAddress = 2; // user that received funds
     }
+
+    message Transfer {
+        bytes address = 1;
+        string amount = 2; // uint64 isn't big enough
+        bytes sender = 3;
+        bytes message_id = 4;
+        bytes channel_id = 5;
```
texuf, line 636: should we add a flag here for buy/sell mapping to the to/from the sender?

core/node/rules/can_add_event.go
```diff
@@ -815,6 +858,24 @@ func (ru *aeReceivedBlockchainTransactionRules) parentEventForReceivedBlockchain
...
+		}
+		// forward the transfer to the stream as a member event, preserving the original sender as the from address
+		return &DerivedEvent{
+			Payload: events.Make_MemberPayload_BlockchainTransaction(
+				ru.receivedTransaction.FromUserAddress,
+				transaction,
+			),
+			StreamId: streamId,
+			Tags:     ru.params.parsedEvent.Event.Tags, // forward tags
+		}, nil
```
texuf, line 878: This can just error out for now. "Received Transaction" is meant for multi party transactions like tips.

packages/sdk/src/tests/multi_ne/trading.test.ts
```diff
@@ -0,0 +1,182 @@
...
+            address: bin_fromHexString(tokenAddress),
+            sender: bin_fromHexString(aliceClient.userId),
+            messageId: bin_fromHexString(threadParentId),
+            channelId: bin_fromHexString(channelId),
+        }
+
+        await expect(
+            aliceClient.addTransaction_Transfer(31337, receipt, transferEvent),
+        ).rejects.toThrow('duplicate transaction')
+    })
```
texuf, line 181: you can add a test - "transaction shows up in channel stream" I've got one in tips. It has a goofy waitfor, but it's just checking the stream events.

protocol/protocol.proto
texuf: something like:
enum TransferType {
    fromSender,
    toSender,
}
and validate here
```
if transfer.From.Cmp(senderAddress) != 0 && transfer.To.Cmp(senderAddress) != 0 {
				continue
			}
```

protocol/protocol.proto
```diff
@@ -1023,6 +1033,7 @@ enum MessageInteractionType {
     MESSAGE_INTERACTION_TYPE_EDIT = 4;
     MESSAGE_INTERACTION_TYPE_REDACTION = 5;
     MESSAGE_INTERACTION_TYPE_TIP = 6;
+    MESSAGE_INTERACTION_TYPE_TRADE = 7;
```
texuf, line 1037: this gets hooked up in tags.ts

protocol/protocol.proto
texuf: idk if enum is better, it's late here lol. bool is great!

protocol/protocol.proto
```diff
@@ -627,11 +627,22 @@ message BlockchainTransaction {
...
+        bytes channel_id = 5;
+        bool is_buy = 6;
+    }
+
     // required fields
     BlockchainTransactionReceipt receipt = 1;
     // optional metadata to be verified by the node
     oneof content {
         Tip tip = 101;
+        TokenTransfer token_transfer = 102;
```
texuf, line 645: heads up i'm adding a new space review type, the merge is going to suck, but i'm numbering mine 103 so you don't have to change this.

texuf: code is here: https://github.com/towns-protocol/towns/pull/2395

texuf (approved):

PR #2438, opened 2025-02-24
Node-2-node auth

rhbntt:
This PR implements mTLS auth for node-to-node communication within the towns-protocol project. The changes include the addition of middleware for requiring client certificates, certificate verification logic, and e2e tests to ensure the functionality of mTLS.

The certificate contains a sender node Ethereum address and a signature of cert's public key by the node's Ethereum private key so the server can verify the signature inside the certificate and check the signer address in the nodes registry.

### DIFF
--- a/core/node/nodes/node_record.go
+++ b/core/node/nodes/node_record.go
@@ -20,26 +20,6 @@ type NodeRecord struct {
 	nodeToNodeClient    NodeToNodeClient
 }
 
-func NewNodeRecord(
-	address common.Address,
-	operator common.Address,
-	url string,
-	status uint8,
-	local bool,
-	streamServiceClient StreamServiceClient,
-	nodeToNodeClient NodeToNodeClient,
-) *NodeRecord {
-	return &NodeRecord{
-		address:             address,
-		operator:            operator,
-		url:                 url,
-		status:              status,
-		local:               local,
-		streamServiceClient: streamServiceClient,
-		nodeToNodeClient:    nodeToNodeClient,
-	}
-}
-
 func (n *NodeRecord) Address() common.Address {
 	return n.address
 }


core/node/rpc/server_node2node_auth.go
```diff
@@ -0,0 +1,224 @@
...
+	privateKeyA, err := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
+	if err != nil {
+		return nil, err
+	}
+
+	publicKeyDER, err := x509.MarshalPKIXPublicKey(&privateKeyA.PublicKey)
+	if err != nil {
+		return nil, err
+	}
+	hash := crypto.Keccak256(publicKeyDER)
```
sergekh2, line 185: when input is hashed for signing, it's a good practice to wrap it uniquely for the purpose so it can't be used for something else.

I refactor hashing function in https://github.com/towns-protocol/towns/pull/2442, create new TownsHashForCert with unique prefix.

core/node/rpc/server_node2node_auth.go
```diff
@@ -0,0 +1,224 @@
...
+
+	signature, err := crypto.Sign(hash, pk)
```
sergekh2, line 187: Pass Wallet class and then wallet.SignHash

core/go.mod
```diff
@@ -8,14 +8,17 @@ require (
 	github.com/SherClockHolmes/webpush-go v1.3.0
 	github.com/cenkalti/backoff/v4 v4.3.0
 	github.com/deckarep/golang-set/v2 v2.6.0
-	github.com/ethereum/go-ethereum v1.14.11
+	github.com/ethereum/go-ethereum v1.15.2
```
texuf, line 11: did you mean to upgrade all these packages in this pr?

texuf (commented):
Is this going to work while nodes are on different versions?

core/node/rpc/node2nodeauth/cert.go
```diff
@@ -0,0 +1,208 @@
...
+// Returns nil if the certificate is valid or not found.
+func VerifyPeerCertificate(logger *zap.SugaredLogger, nodeRegistry nodes.NodeRegistry) func([][]byte, [][]*x509.Certificate) error {
+	return func(rawCerts [][]byte, _ [][]*x509.Certificate) error {
+		for _, rawCert := range rawCerts {
+			peerCert, err := x509.ParseCertificate(rawCert)
+			if err != nil {
+				return err
+			}
+
+			if peerCert.Subject.CommonName == node2NodeCertName {
```
bas-vk, line 46: Would it make sense to structure the subject common name in the certifiate to something like `<node-addr>.<operator-addr>.towns` instead of a hard coded one that is the same for all?

core/node/rpc/node2nodeauth/middleware.go
```diff
@@ -0,0 +1,26 @@
+package node2nodeauth
+
+import (
+	"net/http"
+)
+
+// RequireCertMiddleware is a middleware that requires the node-2-node client certificate.
+// This works together with verifyNode2NodePeerCertificate which verifies the certificate.
+func RequireCertMiddleware(next http.Handler) http.Handler { //nolint:unused
```
bas-vk, line 9: This function is only used in tests, is that intentional?

And is `verifyNode2NodePeerCertificate` `VerifyPeerCertificate` because I can find `verifyNode2NodePeerCertificate`?

core/node/rpc/node2nodeauth/cert.go
```diff
@@ -0,0 +1,208 @@
...
+		cert *tls.Certificate
+		exp  time.Time
+	)
+
+	return func(info *tls.CertificateRequestInfo) (*tls.Certificate, error) {
+		lock.Lock()
+		defer lock.Unlock()
+
+		if cert == nil || time.Now().Add(certRenewGap).After(exp) {
+			newCert, err := createCert(logger, wallet)
```
bas-vk, line 146: `createCert` creates the certificate with:

```
template := &x509.Certificate{
		...
		NotBefore:       time.Now(),
		NotAfter:        time.Now().Add(time.Hour),
		...
	}
```

That means the certificate expires in 1 hour?

Later in this function: `exp = time.Now().Add(certTTL)` with `certTTL = time.Hour * 24`

And `exp` is used to determine if a new certificate must be generated. Does this mean that when the first time a certificate is created it expires after 1 hour and this function only generates a new one after 24 hours (or a restart?)

core/node/rpc/node2nodeauth/cert.go
```diff
@@ -0,0 +1,208 @@
...
+		Signature: signature,
+	})
+	if err != nil {
+		return nil, AsRiverError(err, Err_INTERNAL).Message("Failed to marshal extra extension").LogError(logger)
+	}
+
+	template := &x509.Certificate{
+		SerialNumber:    big.NewInt(1),
+		Subject:         pkix.Name{CommonName: node2NodeCertName},
+		NotBefore:       time.Now(),
```
bas-vk, line 192:
```
template := &x509.Certificate{
		...
		NotBefore:       time.Now(),
}
```

Due to clock drift it is possible that the other nodes clock is slightly behind the clock of the node generating its client cert. So it is possible that when the certificate is used immediately after it is generated the other node might consider it not (yet) valid.

Maybe we should subtract 30s from `NotBefore` to compensate for small clock differences?

core/node/rpc/node2nodeauth/cert.go
```diff
@@ -0,0 +1,208 @@
...
+	. "github.com/towns-protocol/towns/core/node/protocol"
+)
+
+const (
+	node2NodeCertName = "node2node"
+)
+
+var (
+	// node2NodeCertExtOID is the OID of the custom node-2-node client certificate extension.
+	node2NodeCertExtOID = asn1.ObjectIdentifier{1, 3, 6, 1, 4, 1, 50000, 1, 1}
```
bas-vk, line 33: How did you pick this oid?

core/node/rpc/node2nodeauth/cert.go
```diff
@@ -0,0 +1,224 @@
...
+	"github.com/towns-protocol/towns/core/node/crypto"
+	"github.com/towns-protocol/towns/core/node/http_client"
+	"github.com/towns-protocol/towns/core/node/logging"
+	"github.com/towns-protocol/towns/core/node/nodes"
+	. "github.com/towns-protocol/towns/core/node/protocol"
+)
+
+const (
+	certTTL            = time.Hour
+	certDurationBuffer = time.Second * 30
```
sergekh2, line 30: 30 sec seems too aggressive (compared to an hour): clock jitter, networking delays, etc can probably periodically hit that. Minute or two?

core/node/rpc/node2nodeauth/cert.go
```diff
@@ -0,0 +1,224 @@
...
+	certNameFmt = "%s.%s.towns"
+	certIssuer  = "towns"
+)
+
+var (
+	// certExtOID is the OID of the custom node-2-node client certificate extension.
+	certExtOID = asn1.ObjectIdentifier{1, 3, 6, 1, 4, 1, 50000, 1, 1}
+)
+
+// VerifyPeerCertificate goes through the peer certificates and verifies the node-2-node client certificate.
```
sergekh2, line 42: comment is incorrect: it returns a function, which bla-bla

core/node/rpc/node2nodeauth/cert.go
```diff
@@ -0,0 +1,224 @@
...
+	. "github.com/towns-protocol/towns/core/node/protocol"
+)
+
+const (
+	certTTL            = time.Hour
+	certDurationBuffer = time.Second * 30
+	// certNameFmt is the format for the node-2-node client certificate name.
+	// The format is: <chain_id>.<node-address>.towns
+	certNameFmt = "%s.%s.towns"
+	certIssuer  = "towns"
```
sergekh2, line 34:
```suggestion
	certIssuer  = "towns.com"
```

core/node/rpc/node2nodeauth/cert.go
```diff
@@ -0,0 +1,224 @@
...
+	return func(rawCerts [][]byte, certs [][]*x509.Certificate) error {
+		for _, raw := range rawCerts {
+			cert, err := x509.ParseCertificate(raw)
+			if err != nil {
+				return err
+			}
+
+			// The given function is applicable for both internode and stream services.
+			// The given certificate is provided for the internode service only.
+			if len(cert.Subject.Organization) == 1 && cert.Subject.Organization[0] == certIssuer {
```
sergekh2, line 54: this is a bit confusing: it essentially skips verification of any certs it doesn't recognize. Is this intended behavior?

core/node/rpc/node2nodeauth/cert.go
```diff
@@ -0,0 +1,224 @@
...
+		return RiverError(Err_UNAUTHENTICATED, "Node-2-node cert signature does not match address").LogError(logger)
+	}
+
+	node, err := nodeRegistry.GetNode(recoveredAddr)
+	if err != nil {
+		return err
+	}
+
+	// Check that that sender node is operational
+	if node.Status() != river.NodeStatus_Operational {
```
sergekh2, line 132: node can communicate, but not necessary be fully operational: i.e. if it's exiting, and data still replicate out, or it's read-only mode, etc.

core/node/rpc/node2nodeauth/middleware.go
```diff
@@ -0,0 +1,19 @@
...
+import (
+	"net/http"
+)
+
+// RequireCertMiddleware is a middleware that requires the node-2-node client certificate.
+// This works together with VerifyPeerCertificate which verifies the certificate.
+// This middleware will be attached to the node-2-node RPC server after mTLS auth is deployed to all envs.
+func RequireCertMiddleware(next http.Handler) http.Handler { //nolint:unused
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if len(r.TLS.PeerCertificates) != 1 {
```
sergekh2, line 12: so here it required exactly 1 cert, but in validate func it succeeds if cert is not recognized. (maybe test)

core/node/rpc/node2nodeauth/middleware.go
sergekh2: please add on-chain option to require node2node certs, then we can enable it in CI, etc.

core/node/rpc/node2nodeauth/cert.go
bas-vk: Same issue for `NotAfter` if the remote nodes clock is just ahead of the local node.

core/node/rpc/node2nodeauth/middleware.go
sergekh2: I don't think this check is enough, it seems correct check is to succeed if required cert is present among all PeerCerts.

core/node/rpc/node2nodeauth/cert.go
```diff
@@ -0,0 +1,212 @@
...
+	if err != nil {
+		return AsRiverError(err, Err_UNAUTHENTICATED).Message("Failed to extract public key from the cert signature").LogError(logger)
+	}
+	recoveredAddr := ethcrypto.PubkeyToAddress(*sigPublicKey)
+
+	if recoveredAddr.Cmp(common.HexToAddress(certExt.Address)) != 0 {
+		return RiverError(Err_UNAUTHENTICATED, "Node-2-node cert signature does not match address").LogError(logger)
+	}
+
+	return nil
```
sergekh2, line 124: it appears that also it should check if cert matches active nodes in the node registry here

clemire (approved):
LGTM!

core/node/rpc/node2nodeauth/middleware_test.go
```diff
@@ -0,0 +1,52 @@
...
+import (
+	"crypto/tls"
+	"crypto/x509"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+)
+
+func TestRequireCertMiddleware(t *testing.T) {
+	tests := []struct {
```
sergekh2, line 12: For just a single test it seems it's not necessary to have struct and loop. Extra code comes at the cost, it's better if it's used or not there.

PR #2460, opened 2025-02-26
Process messages in a separate goroutine when syncing stream

rhbntt:
Currently, sending a message to a slow client blocks processing commands within the stream sync operation.
This PR separates sending messages to the client from a process of handling commands and stream updates.


core/node/rpc/sync/operation.go
```diff
@@ -123,82 +119,126 @@ func (syncOp *StreamSyncOperation) Run(
...
-			}
+	// Create a channel to send messages to the client
+	type clientMsg struct {
+		msg *SyncStreamsResponse
+		err chan error
+	}
+	clientMsgs := make(chan clientMsg, 1)
+
+	// Start a separate goroutine to process stream update messages
+	go func() {
```
sergekh2, line 145: lets refactor to have goroutines as separate functions or methods. As a general rule, if goroutine is longer than couple lines, lets have it as a standalone function.

core/node/rpc/sync/operation.go
```diff
@@ -123,82 +119,126 @@ func (syncOp *StreamSyncOperation) Run(
...
-				})
-
-				cmd.Reply(nil)
-				return nil
-			}
+		case <-stop:
+			return nil
+		case msg := <-clientMsgs:
+			msg.msg.SyncId = syncOp.SyncID
+			msg.err <- res.Send(msg.msg)
```
sergekh2, line 241: sending and error back seems like an unnecessary complication: if Send returns error, this sync is done, why not to exit and teardown everything?

additionally: it seems redundant with the goroutine in line 145, why not to read from messages right here?

bas-vk: I think it is fine in this case. The client should get the signal that the sync session was stopped if the send failed and the node terminates the session. If it never receives the signal that connection is closed it will also not get the corresponding pong (checked through the nonce) over the stream and stop and restart the sync.

PR #2500, opened 2025-03-04
chore: remove generated contract artifacts from git

shuhuiluo:
Adjust build scripts to include contract artifact generation and synchronize dependencies between packages. Updated workflows to use stable Foundry versions and added steps for building contract artifacts before publishing. Adjusted `.gitignore` for dev artifacts and updated package management details.


texuf (approved):
we can try it. I know checking in generated artifacts is a fools errand, but I really love knowing exactly what version of code i'm working off of, and as someone who's constantly switching branches and working on multiple things it's one less thing to worry about. I think this will need a watch to play nice with the rest of our dev process, but I can add that later.

giuseppecrj (approved):

PR #2502, opened 2025-03-04
repltest: ensure sync works in replicated scenario testing

bas-vk:
With this change each time when test clients are instantiated a stream sync session is started. When `testClients#createChannelAndJoin` is called all clients will subscribe for updates on the created stream though modify sync. When the test calls `testClients#compare(StreamId)` not only the stream are compared between clients but also ensured that all clients received the same updates through the sync session.

### DIFF
--- a/core/node/rpc/tester_test.go
+++ b/core/node/rpc/tester_test.go
@@ -3,7 +3,6 @@ package rpc
 import (
 	"context"
 	"fmt"
-	"github.com/google/go-cmp/cmp"
 	"github.com/puzpuzpuz/xsync/v3"
 	"go.uber.org/atomic"
 	"hash/fnv"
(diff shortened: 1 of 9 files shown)


sergekh2 (approved):

core/node/rpc/tester_test.go
```diff
@@ -560,6 +588,9 @@ func (st *serviceTester) newTestClients(numClients int) testClients {
 	clients.parallelForAll(func(tc *testClient) {
 		tc.createUserStream()
 	})
+	clients.parallelForAll(func(tc *testClient) {
```
sergekh2, line 591: perhaps it should be an option so it's possible to test in different modes?

sergekh2 (approved):

sergekh2 (approved):

PR #2520, opened 2025-03-05
node: on add event sync an out of sync stream before trying to add an event

bas-vk:


core/node/rpc/add_event.go
```diff
@@ -31,8 +31,39 @@ func (s *Service) localAddEvent(
...
+			}
+
+			if !scheduled {
+				s.cache.SubmitSyncStreamTask(ctx, localStream)
+				scheduled = true
+				timeout = time.After(5000 * time.Millisecond)
+			}
+
+			select {
+			case <-time.After(10 * time.Millisecond):
```
sergekh2, line 53: 10 ms is too aggressive. Lets use BackoffTracker https://github.com/towns-protocol/towns/blob/3f6e9b93964c1bf7cd5ea502cb3b84768fb4a729/core/node/base/backoff_tracker.go#L10 (create a derived context with 5 sec timeout to also have a timeout)

core/node/rpc/add_event.go
```diff
@@ -31,8 +31,39 @@ func (s *Service) localAddEvent(
 
 	log.Debugw("localAddEvent", "parsedEvent", parsedEvent)
 
-	newEvents, err := s.addParsedEvent(ctx, streamId, parsedEvent, localStream, streamView)
+	if parsedEvent.MiniblockRef.Num >= 0 {
```
sergekh2, line 34: since both codeblocks are the same, please put this code in a function.

sergekh2 (approved):

core/node/events/stream_cache.go
```diff
@@ -67,9 +69,11 @@ type StreamCache struct {
 	streamCacheUnloadedGauge prometheus.Gauge
 	streamCacheRemoteGauge   prometheus.Gauge
 
-	stoppedMu            sync.RWMutex
-	onlineSyncWorkerPool *workerpool.WorkerPool
-	stopped              bool
+	stoppedMu                         sync.RWMutex
```
sergekh2, line 72: (vars can be grouped with respective mutexes here)

core/node/rpc/add_event.go
```diff
@@ -31,8 +31,39 @@ func (s *Service) localAddEvent(
...
+		// before trying to add the event to the stream schedule a sync request if outdated.
+		var timeout <-chan time.Time
+		scheduled := false
+		for {
+			// TODO: this has the potential to cause a lot of syn stream tasks.
```
sergekh2, line 39: it's a concern, but I think a simple mitigation is to not schedule for first couple retries (hoping that this stream is being already updated by regular mb code)

core/node/rpc/add_event.go
```diff
@@ -31,8 +31,39 @@ func (s *Service) localAddEvent(
...
+			//  Consider not scheduling a sync task when the node is almost up to date and trust on the client retry mechanism
+			streamUpToDate := streamView.GetStats().LastMiniblockNum >= parsedEvent.MiniblockRef.Num
+			if streamUpToDate {
+				break
+			}
+
+			if !scheduled {
+				s.cache.SubmitSyncStreamTask(ctx, localStream)
```
sergekh2, line 47: server context from s should be used here: RPC context can be canceled before task completes. (in this sense maybe Submit API needs to be redesigned not to take context and just always use default server context internally)

texuf (commented):
neat

sergekh2 (approved):

PR #2521, opened 2025-03-05
Use ModifySync in the remote syncer

rhbntt:


core/node/rpc/service_test.go
```diff
@@ -932,6 +932,11 @@ OuterLoop:
 		channelHash,
 	)
 	require.Nilf(err, "error creating message event: %v", err)
+
+	// Wait 5 seconds for the remote syncer to send ModifySync request to alice's node with the remoted stream.
+	// Check remote syncer implementation.
+	time.Sleep(time.Second * 4)
```
clemire, line 938: Are we able to replace this wait with a `require.EventuallyWithT` statement that moves forward as soon as the necessary condition is satisfied?

core/node/rpc/sync/client/remote.go
```diff
@@ -240,51 +249,102 @@ func (s *remoteSyncer) connectionAlive(latestMsgReceived *atomic.Value) {
...
 }
 
-func (s *remoteSyncer) Address() common.Address {
-	return s.remoteAddr
+// startStreamModifier starts a goroutine that periodically sends modify sync requests to the remote.
+func (s *remoteSyncer) startStreamModifier() {
+	var (
+		log = logging.FromCtx(s.syncStreamCtx)
+		// check every modifySyncTicker if it's time to send a modify sync req to remote
+		modifySyncTicker = time.NewTicker(3 * time.Second)
```
clemire, line 257: Unactionable comment: I'm unsure if we want to eventually make this interval configurable, food for thought.

core/node/rpc/sync/client/remote.go
```diff
@@ -297,7 +357,95 @@ func (s *remoteSyncer) RemoveStream(ctx context.Context, streamID StreamId) (boo
...
+	defer s.pendingModifySyncLock.Unlock()
+
+	if s.pendingModifySync == nil {
+		s.pendingModifySync = &ModifySyncRequest{}
+	}
+
+	// If the given stream exists in the pending list to add, remove the last one from there
+	var removed bool
+	for i := len(s.pendingModifySync.AddStreams) - 1; i >= 0; i-- {
+		addStreamID, err := StreamIdFromBytes(s.pendingModifySync.AddStreams[i].GetStreamId())
```
clemire, line 425: nit: it seems like the code to find if a slice contains an element is repeated a few times in this implementation, and I wonder if we could use something like `https://pkg.go.dev/slices#ContainsFunc` and/or other methods from the `slices` package to make this more concise.

clemire (changes requested):

clemire:
Great progress! <3

core/node/rpc/sync/client/remote.go
```diff
@@ -240,51 +250,102 @@ func (s *remoteSyncer) connectionAlive(latestMsgReceived *atomic.Value) {
...
+		streamId, err := StreamIdFromBytes(cookie.GetStreamId())
+		if err != nil {
+			logger.Errorw("Failed to parse stream ID", "err", err)
+			continue
+		}
 
-	return err
-}
+		var isFailed bool
+		for _, failed := range resp.Msg.GetAdds() {
```
clemire, line 318: I missed this a while ago, but it would be more helpful if the Adds and Removals in the response were renamed FailedAdds and FailedRemovals, because as I was reading this, I thought they were successful adds and fails, and i was really confused by this piece of code.

texuf: I just had the same reading comprehension failure. Was very confused by the code.

bas-vk:
With the clients moving to modify sync is it possible to group the add and remove stream commands from the clients modify sync request per node and forward all the add/rm stream commands for a remote syncer at once instead of one by one in the rpc layer?

If that is possible there is no need to batch commands in 2s intervals in the remote syncer while reducing latency, complexity and a background go-routine.

bas-vk:
> > With the clients moving to modify sync is it possible to group the add and remove stream commands from the clients modify sync request per node and forward all the add/rm stream commands for a remote syncer at once instead of one by one in the rpc layer?
> > If that is possible there is no need to batch commands in 2s intervals in the remote syncer while reducing latency, complexity and a background go-routine.
> 
> A client calls ModifySync RPC endpoint to add/remove streams. The list of streams from the request may contain streams stored on different nodes so the receiver node should go over all streams and forward it to the right node. The receiver node still has to handle the request to client and prepare a request(s) to remote(s). Could you elaborate more on this idea (I'm not sure I fully get it)? A proper solution would be to have a single remote syncer per remote but it requires much more research and refactoring.

Here: https://github.com/towns-protocol/towns/blob/main/core/node/rpc/service_sync_streams.go#L88 the modify sync call iterates over all adds and removals from the request and sends them one by one to the handler for processing.

These end up in calls to `remoteSyncer#AddStream` and `remoteSyncer#RemoveStream` that append/remove the stream to `remoteSyncer.pendingModifySync` that every 2s is used to make the actual call on the remote node.

So the commands are first split and later combined again in the remote syncer. Why not do the grouping per remote node of the adds and removals in `Service#ModifySync` and sends all commands for a node to the remote at once?

sergekh2 (approved):

core/node/rpc/sync/client/syncer_set.go
```diff
@@ -304,6 +301,136 @@ func (ss *SyncerSet) RemoveStream(ctx context.Context, streamID StreamId) error
 	return nil
 }
 
+// Modify splits the given request into add and remove operations and forwards them to the responsible syncers.
+func (ss *SyncerSet) Modify(ctx context.Context, req ModifyRequest) error {
+	if len(req.ToAdd) > 0 && ss.stopped {
```
bas-vk, line 306: `ss.stopped` should be checked while `ss.muSyncers` lock is held.

clemire (approved):

PR #2535, opened 2025-03-06
Bot message queueing with key solicitations

clemire:
A queue/cache is implemented on top of postgres storage primitives for the app registry service. The forwarding test case has been updated to demonstrate that the app registry service requests a key solicitation from the bot when it is missing key material for sending a message.

Other updates:
- postgres has been updated to store groups of session ids together with a ciphertext since this is how they are sent to the stream, and the app registry service does not have key material to decrypt these and make the ciphertexts more granular.
- DB queries were also updated to reflect that multiple instances of key material with the same session id may be published to a particular user's inbox stream, and each is saved to the db. However, when enqueueing messages, only one set of (device, key material) information is returned for each device that has key material for the message.
- There was a bug in the code for waiting for user metadata stream when registering the webhook that would cause this call to always take 10 seconds, even if the encryption device was already available in the stream. This has been fixed


core/node/app_registry/encrypted_message_store.go
```diff
@@ -0,0 +1,195 @@
...
+	deviceKey string,
+	fallbackKey string,
+) error {
+	if err := q.store.RegisterWebhook(ctx, app, webhook, deviceKey, fallbackKey); err != nil {
+		return err
+	}
+
+	q.appIdCache.Store(app, struct{}{})
+	return nil
+}
```
texuf, line 78: is this file named properly? kinda funny that the encrypted_message_store is Publishing Session Keys

core/node/app_registry/event_processor.go
```diff
@@ -0,0 +1,74 @@
...
+				event,
+				"streamId",
+				channelId,
+				"spaceId",
+				spaceId,
+				"error",
+				err,
+			)
+		}
+	}
```
texuf, line 73: i think we want to send all messages, and just do the queueing for messages with encrypted payloads

core/node/app_registry/sync/streams_tracker.go
```diff
@@ -11,15 +12,37 @@ import (
...
 )
 
+type EncryptedMessageStore interface {
+	PublishSessionKeys(
+		ctx context.Context,
+		streamId shared.StreamId,
+		deviceKey string,
+		sessionIds []string,
+		ciphertexts string,
+	) (err error)
```
texuf, line 25: just a nit on naming - why did you call this one "Publish" SessionKeys, and not EnqueueSessionKeys or "AddSessionKeys"

core/node/app_registry/testutil.go
```diff
@@ -10,14 +10,24 @@ import (
...
+	"github.com/towns-protocol/towns/core/node/protocol/protocolconnect"
+	"github.com/towns-protocol/towns/core/node/shared"
 	"github.com/towns-protocol/towns/core/node/testutils/testcert"
 	"github.com/towns-protocol/towns/core/node/utils"
 )
 
+type AppServiceRequestEnvelope struct {
+	Command string          `json:"command"`
+	Data    json.RawMessage `json:",omitempty"`
+}
```
texuf, line 30: I assumed we would define this in the protobuf. Do we plan to move it there?

texuf (approved):

core/node/app_registry/testutil.go
clemire: Did we expect bot servers to implement a grpc endpoint? I figured JSON REST over https was the lowest friction expectation.

core/node/app_registry/encrypted_message_store.go
clemire: "publish" probably came to mind because adding the session keys can cause messages to be "published" to subscribers, in the language of a queue or a pubsub.

No I think it's not named right, since the struct is a CachedEncryptedMessageQueue. Does publish make sense for cached_encrypted_message_queue.go?

core/node/app_registry/event_processor.go
clemire: `EnqueueMessages` could perhaps also be renamed... it immediately sends sendable messages and only enqueues what we don't have keys for

core/node/app_registry/sync/streams_tracker.go
clemire: answer given above

clemire: Happy to rename either way though, I'm not married to this or strongly opinionated about it.

core/node/storage/app_registry_migrations/000001_create_initial_schema.up.sql
```diff
@@ -11,14 +11,26 @@ ALTER TABLE app_registry ADD CONSTRAINT unique_device_key UNIQUE (device_key);
...
+    CHECK (array_length(session_ids, 1) > 0), -- session ids array contains at least 1 element
+    CHECK (array_has_no_duplicates(session_ids)), -- all session ids are unique within an array
     CONSTRAINT fk_device_key FOREIGN KEY (device_key) REFERENCES app_registry(device_key)
 );
 
+CREATE EXTENSION IF NOT EXISTS btree_gin;
+
+-- Index on individual elements of the session_ids column so that we can search the db
+-- by session id.
+CREATE INDEX idx_app_session_keys_device_key_session_id ON app_session_keys USING GIN (device_key, session_ids);
```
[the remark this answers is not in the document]
clemire: ?

core/node/app_registry/cached_encrypted_message_queue.go
```diff
@@ -0,0 +1,195 @@
...
+		DeviceKey:             deviceKey,
+		SessionIds:            sessionIds,
+		CipherTexts:           ciphertexts,
+		WebhookUrl:            sendableMessages.WebhookUrl,
+		StreamEvents:          sendableMessages.StreamEvents,
+	}
+	return q.appDispatcher.SubmitMessages(ctx, messages)
+}
+
+// EnqueueMessages will immediately send a message for each device that has session keys, and will
```
clemire, line 103: note to self- update this

core/node/app_registry/testutil.go
clemire: I will follow up on this in a future PR

core/node/app_registry/cached_encrypted_message_queue.go
clemire, line 103:
```suggestion
// DispatchOrEnqueueMessages will immediately send a message for each device that has session keys, and will
```

PR #2542, opened 2025-03-07
allow missing preTokenBalance[mint]

erikolsson:
oversight on my end — when a user opens a token account (buys a token for the first time), the `preTokenBalance` isn't set — this PR makes sure that we allow receipts both with and without `preTokenBalance[address: mint]`.

### DIFF
--- a/core/node/rules/can_add_event.go
+++ b/core/node/rules/can_add_event.go
@@ -927,9 +927,10 @@ func (ru *aeBlockchainTransactionRules) validBlockchainTransaction_CheckReceiptM
 		var amountBefore = big.NewInt(0)
 		if idx != len(meta.GetPreTokenBalances()) {
 			var ok bool
-			amountBefore, ok = new(big.Int).SetString(meta.GetPreTokenBalances()[idx].Amount.Amount, 0)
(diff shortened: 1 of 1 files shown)


core/node/rules/can_add_event.go
```diff
@@ -921,12 +921,16 @@ func (ru *aeBlockchainTransactionRules) validBlockchainTransaction_CheckReceiptM
...
-			return false, RiverError(Err_INVALID_ARGUMENT, "invalid pre token balance amount")
+
+		// preTokenBalances isn't set when a user opens a token account (buys a token for the 1st time),
+		// so we need to check if it's not empty otherwise, we use 0 as the amount before
+		var amountBefore = big.NewInt(0)
+		if idx != len(meta.GetPreTokenBalances()) {
+			var ok bool
+			amountBefore, ok = new(big.Int).SetString(meta.GetPreTokenBalances()[idx].Amount.Amount, 0)
+			if !ok {
+				return false, RiverError(Err_INVALID_ARGUMENT, "invalid pre token balance amount")
```
clemire, line 932: Nit: can you add a tag here with "amount" so we can see what broke it if this ever happens?

texuf (approved):

PR #2555, opened 2025-03-09
setup wagmi for contract typings

giuseppecrj:


shuhuiluo (approved):

contracts/wagmi.config.ts
```diff
@@ -0,0 +1,56 @@
...
+                '**/IERC721A.sol/*.json',
+                '**/INodeOperator.sol/*.json',
+                '**/ISpaceDelegation.sol/*.json',
+                '**/IEntitlementChecker.sol/*.json',
+                '**/IERC5267.sol/*.json',
+                '**/ICreateSpace.sol/*.json',
+                '**/IDropFacet.sol/*.json',
+                '**/ITownsPoints.sol/*.json',
+                '**/ITipping.sol/*.json',
+                '**/IReview.sol/*.json',
```
texuf, line 49: Can this just be `**/*.sol/*.json` so you don't have to manually add things here? If not why?

jterzis: You could use an `exclude` array, but there are also interfaces we choose not to build types for that we'd need to exclude like IRewardsDistribution so it's probably not worth it over your solution. 

```
export default defineConfig({
    out: 'typings/index.ts',
    plugins: [
        foundry({
            project: '../',
            artifacts: 'contracts/out',
            include: ['**/I*.sol/*.json'],
            exclude: [
                '**/*.t.sol',
                '**/IERC721Metadata.sol/*.json',
                '**/IERC20.sol/*.json',
                '**/IERC165.sol/*.json',
                '**/IERC721.sol/*.json',
                '**/IMetadata.sol/*.json',
                '**/Initializable.sol/*.json',
                '**/IRewardsDistribution.sol/*.json',
            ],
            forge: {
                build: false,
            },
        }),
    ],
})
```

jterzis (approved):
Looks good - assuming this will eventually replace the need to `build-contract-types`.

scripts/anvil-fork.sh
```diff
@@ -0,0 +1,129 @@
...
+done
+
+# If no fork block number is specified, get the latest block number
+if [ -z "$FORK_BLOCK_NUMBER" ]; then
+  # Only show debug messages if not in output-block-number mode
+  if [ "$OUTPUT_BLOCK_NUMBER" = "false" ]; then
+    echo "Getting latest block number from $FORK_URL..."
+  fi
+
+  LATEST_BLOCK=$(curl -s -X POST -H "Content-Type: application/json" --data '{"jsonrpc":"2.0","method":"eth_blockNumber","params":[],"id":1}' "$FORK_URL" | grep -o '"result":"0x[^"]*' | cut -d'"' -f4)
```
shuhuiluo, line 84: What's this? Just do [`cast block-number`](https://book.getfoundry.sh/reference/cast/cast-block-number).

scripts/setup-subgraph.sh
```diff
@@ -0,0 +1,129 @@
...
+
+# Register the cleanup function to be called on exit
+trap cleanup EXIT
+
+# Wait for Anvil to start
+echo "Waiting for Anvil to start..."
+for i in {1..30}; do
+  if curl -s -X POST -H "Content-Type: application/json" \
+     --data '{"jsonrpc":"2.0","method":"eth_blockNumber","params":[],"id":1}' \
+     http://localhost:$PORT > /dev/null; then
```
shuhuiluo, line 106: `cast block-number`.

PR #2560, opened 2025-03-10
Dynamic messages buffer in sync stream

rhbntt:


core/node/rpc/sync/dynmsgbuf/dynmsgbuf.go
```diff
@@ -0,0 +1,101 @@
...
+	// ErrClosed is returned when the buffer is closed.
+	ErrClosed = errors.New("buffer is closed")
+)
+
+// DynamicBuffer is a thread-safe, dynamically resizing message buffer.
+type DynamicBuffer[T any] struct {
+	mu         sync.Mutex
+	buffer     []T
+	signalChan chan struct{} // Signals when new items are added
+	closed     int64
```
sergekh2, line 27: lets use buffer == nil as signal for buffer being closed. It's rare operation, having extra interlocked var doesn't seem justified

core/node/rpc/sync/dynmsgbuf/dynmsgbuf.go
```diff
@@ -0,0 +1,101 @@
...
+	db.mu.Lock()
+	defer db.mu.Unlock()
+
+	if len(db.buffer) >= maxBufferSize {
+		return ErrBufferFull
+	}
+
+	db.buffer = append(db.buffer, item)
+
+	// Non-blocking signal (only if empty, avoids duplicate wake-ups)
```
sergekh2, line 53: it's better to unlock before channel write here: then buffer read can acquire mu and proceed immediately after channel read.

core/node/rpc/sync/dynmsgbuf/dynmsgbuf.go
```diff
@@ -0,0 +1,101 @@
...
+		batchSize = len(db.buffer)
+	}
+
+	// Copy batch & remove from buffer
+	batch := make([]T, batchSize)
+	copy(batch, db.buffer[:batchSize])
+	db.buffer = db.buffer[batchSize:]
+
+	return batch
+}
```
sergekh2, line 82: Doing this append-only read from front approach is gonna trash memory needlessly.
Apart from using circular buffers it's possible to make it efficient by using two buffers like this:

```suggestion
func (db *DynamicBuffer[T]) GetBatch(prev []T) []T {
    // Reset if capacity is unused
    if prev == nil || (cap(prev) > minBufferSize * 2 && len(prev) < cap(prev) / 2){
        prev = make([]T, 0, minBufferSize)
    } else {
        // Deref pointers so they can be gc'ed faster
        var zero T
        for i := range prev {
            prev[i] = zero
        }
        prev = prev[:0]
    }
    
	db.mu.Lock()
	defer db.mu.Unlock()
    ret := db.buffer
    db.buffer = prev
	return ret
}
```

core/node/rpc/sync/dynmsgbuf/dynmsgbuf.go
```diff
@@ -0,0 +1,101 @@
...
+	mu         sync.Mutex
+	buffer     []T
+	signalChan chan struct{} // Signals when new items are added
+	closed     int64
+}
+
+// NewDynamicBuffer initializes a new dynamic buffer.
+func NewDynamicBuffer[T any]() *DynamicBuffer[T] {
+	return &DynamicBuffer[T]{
+		buffer:     make([]T, 0),
```
sergekh2, line 33: lets also have minBufferSize 16

sergekh2 (approved):

bas-vk (approved):

PR #2574, opened 2025-03-11
Add optional syncCookie param to GetStream

texuf:
allows for partial get stream requests, simplifies client code, allows clients to cache streams that they aren’t members of in the same format, reducing duplicate requests


core/node/events/stream_view.go
```diff
@@ -892,3 +893,69 @@ func (r *StreamView) AllEvents() iter.Seq[*ParsedEvent] {
...
+		return &StreamAndCookie{
+			Events:         envelopes,
+			NextSyncCookie: r.SyncCookie(localNodeAddress),
+		}, nil
+
+	} else {
+
+		miniblockIndex, err := r.indexOfMiniblockWithNum(cookie.MinipoolGen)
+		if err != nil {
+			// The user's sync cookie is out of date. Send a sync reset and return an up-to-date StreamAndCookie.
```
clemire, line 930: I think this comment doesn't make sense here

texuf: this code is just moved, not changed, but can you help me make it make more sense? in this case the minipoolGen passed in the sync cookie is no longer in memory in the node, so a delta of events cannot be passed back to the client, so sync reset toggled to true and an up to date stream and cookie are returned

clemire: Ah yes, you're right. There is a SyncReset on the return type, my bad

core/node/events/stream_view.go
```diff
@@ -892,3 +893,69 @@ func (r *StreamView) AllEvents() iter.Seq[*ParsedEvent] {
...
+
+func (r *StreamView) GetStreamSince(
+	ctx context.Context,
+	localNodeAddress common.Address,
+	cookie *SyncCookie,
+) (*StreamAndCookie, error) {
+	log := logging.FromCtx(ctx)
+	slot := cookie.MinipoolSlot
+	if slot < 0 {
+		return nil, RiverError(Err_BAD_SYNC_COOKIE, "bad slot", "cookie.MinipoolSlot", slot).Func("Stream.Sub")
```
clemire, line 905: Nit: some of the logs and error messages here and below are referring to "Stream.Sub", can we update them?

core/node/rpc/forwarder.go
```diff
@@ -197,8 +198,34 @@ func (s *Service) getStreamImpl(
...
+			} else {
+				return nil, RiverError(Err_BAD_SYNC_COOKIE, "Stream not found").
+					Func("service.getStreamImpl").
+					Tag("streamId", req.Msg.StreamId)
+			}
+		} else {
+			stub, err := s.nodeRegistry.GetStreamServiceClientForAddress(nodeAddress)
+			if err != nil {
+				return nil, AsRiverError(err).
+					Func("peerNodeRequestWithRetries").
```
clemire, line 216: `Func` value is not right here.

...I assume you deliberately did not use peerNodeRequestWithRetries here?

packages/sdk/src/tests/multi_ne/streamRpcClientGetSince.test.ts
```diff
@@ -0,0 +1,140 @@
...
+    test('make block with 2 events', async () => {
+        await client.info({
+            debug: ['make_miniblock', streamIdAsString(bobsSettingsStreamId), 'true'],
+        })
+        // eventually the block should get made and we should have miniblocks instead of events in the pool
+        const streamSince = await client.getStream({
+            streamId: bobsSettingsStreamId,
+            syncCookie: cookie,
+        })
+        expect(streamSince.stream?.events.length).toBe(3) // all events since last cookie
```
clemire, line 104: I'm a little confused by this test... shouldn't there be one miniblock and no events?

clemire (approved):

packages/sdk/src/tests/multi_ne/streamRpcClientGetSince.test.ts
texuf: the streamAndCookie just returns events. I was confused as well.

core/node/rpc/forwarder.go
```diff
@@ -197,8 +198,33 @@ func (s *Service) getStreamImpl(
...
+			stub, err := s.nodeRegistry.GetStreamServiceClientForAddress(nodeAddress)
+			if err == nil {
+
+				ret, err := stub.GetStream(ctx, req)
+				if err != nil {
+					return nil, err
+				}
+				return connect.NewResponse(ret.Msg), nil
+			}
+			// in the case were we couldn't get a stub for this node, fall through and try to get the stream from scratch
```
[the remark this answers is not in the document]
texuf: There is a legitimate case where this could happen in the future, once we allow nodes to exit the network. But for now its an error. I'll add one.

sergekh2 (approved):

bas-vk (approved):

PR #2578, opened 2025-03-11
refactor(contract): unify stream event handling with `StreamUpdated`

shuhuiluo:
Introduce a single `StreamUpdated` event to replace specific stream-related events, simplifying event handling. Added an enum to categorize stream event types and standardized event data encoding. Deprecated old event emissions for future removal.


bas-vk:
`StreamLastMiniblockUpdateFailed` events still use the current tailored eventing model which I think is fine. But the event type is also defined as `StreamEventType.LastMiniblockUpdateFailed`. Can we drop it if it isn't used? Want to prevent linters reporting issues with missing arms in switch statements.

Will the replication factor in the stream land in a different PR or will it be included in this one?

shuhuiluo:
> `StreamLastMiniblockUpdateFailed` events still use the current tailored eventing model which I think is fine. But the event type is also defined as `StreamEventType.LastMiniblockUpdateFailed`. Can we drop it if it isn't used? Want to prevent linters reporting issues with missing arms in switch statements.
> 
> Will the replication factor in the stream land in a different PR or will it be included in this one?

`StreamEventType.LastMiniblockUpdateFailed` will be deleted. The replication factor will be another one.

contracts/src/river/registry/facets/stream/StreamRegistry.sol
```diff
@@ -78,6 +90,10 @@ contract StreamRegistry is IStreamRegistry, RegistryModifiers {
 
     if (miniblockCount == 0) revert(RiverRegistryErrors.BAD_ARG);
 
+    bytes32[] memory streamIds = new bytes32[](miniblockCount);
+    Stream[] memory streams = new Stream[](miniblockCount);
```
sergekh2, line 94: There is StreamWithId datastructure, would it be better to use it here?

Alternatively, since this is the most common operation, we can leave only relevant data here (streamId, lastMiniblockHash, lastMiniblockNum, isSealed) for sake of efficiency.

Also, do we want to add failures to this event as well?

contracts/src/river/registry/facets/stream/StreamRegistry.sol
```diff
@@ -175,6 +219,12 @@ contract StreamRegistry is IStreamRegistry, RegistryModifiers {
       stream.flags |= StreamFlags.SEALED;
     }
 
+    emit StreamUpdated(
+      StreamEventType.LastMiniblockUpdated,
```
sergekh2, line 223: setStreamLastMiniblock is not used anymore, only Batch version is called. This function and event value can be removed.

shuhuiluo: Good to know.

giuseppecrj (approved):

PR #2594, opened 2025-03-13
compress APNS payload if needed

erikolsson:
Tx receipts are massive! But they zip very well — from my testing, the payload size for a specific tx went down from ~7k to 2k. We can set this to a separate field, `zippedEvent` for backwards compatibility.


texuf (approved):
Neat

PR #2598, opened 2025-03-13
node: add support for logging components on custom log levels

bas-vk:
Add support for multiple loggers per component with their own log level.

Update in configuration:
```
log:
    level: info
    file: logs/dev.log
    fileLevel: info,miniblock=debug,rpc=warn
    consoleLevel: info,miniblock=info,rpc=info
```

Log level config now starts with a default log level and is followed by list of key value pairs, comma separated with the key being a component and the value being a log level. In the above config the everything written to the miniblock component with debug level or higher will be written to file. Anything written with warm or higher to the rpc component will also be written while the default level is info. See `TestParseLogLevels` for more examples and `logging.Component` for a list of supported components.

For this uber zap cores that write to stdout/file are wrapped into a custom core that keeps track the log level per component or falls back to the default level if not specified. Because zap structure only uses a single core and our process can write to stdout and a file at the same time `logging.MultiCores` is introduced that wraps 1 or more cores.

The process now uses `logging.Log` underneath that provides per component a logger. To prevent rewriting all log statement this structures has the `Debug`, `Info`, ... calls and forwards the call to the `Default` logger.

Once this PR is merged a follow-up action is to introduce more components and update the code to use these component loggers.


core/node/logging/component.go
```diff
@@ -0,0 +1,252 @@
...
+		Miniblock *zap.SugaredLogger
+		// Miniblock represents the logger for the Rpc component.
+		Rpc *zap.SugaredLogger
+	}
+
+	// LogLevels represents the set of log levels for each component.
+	LogLevels map[Component]zapcore.Level
+
+	// CoreWithLevels is a zapcore.Core that supports custom log levels per component.
+	CoreWithLevels struct {
```
sergekh2, line 30: seems to be not used anymore

core/node/logging/component.go
```diff
@@ -0,0 +1,252 @@
...
+		zapcore.Core
+		Levels LogLevels
+	}
+
+	// MultiCores combines multiple cores into one.
+	MultiCores []zapcore.Core
```
sergekh2, line 36: same

core/cmd/root_cmd.go
```diff
@@ -88,7 +89,14 @@ func initConfigAndLog() {
 		fmt.Println("Failed to initialize config, error=", err)
 		os.Exit(1)
 	}
-	InitLogFromConfig(&cmdConfig.Log)
+
+	logging.Init(
```
sergekh2, line 93: please also update "log_level" option doc in this file

sergekh2 (approved):

PR #2600, opened 2025-03-13
Prevent passing duplicates in modify sync

rhbntt:


core/node/rpc/sync/client/syncer_set.go
```diff
@@ -311,6 +311,7 @@ func (ss *SyncerSet) Modify(ctx context.Context, req ModifyRequest) error {
 	}
 
 	modifySyncs := make(map[common.Address]*ModifySyncRequest)
+	duplicates := make(map[StreamId]struct{})
```
texuf, line 314: Why are you building up the duplicates map but only using it in 1 of your 3 comparisons?

Seems like you could either use it in all three places, or just use containsFunc for the last case

Or just do one loop over adds and removes and check the map and be done with it...

Sorry for picking on style, I'm spoiled by typescript

texuf (approved):

clemire (approved):

PR #2605, opened 2025-03-14
chore(contract): update Foundry settings and remove unused Multicall …

shuhuiluo:
…mock

Deleted `MockMulticall` contract and its associated test files as they are no longer needed. Improved Foundry configuration by updating Solidity compiler settings, EVM version, optimizer runs, and enabling additional output options. Refactored a test function name for clarity.

Moved deployment initializations inside the `__interact` functions to reduce unnecessary global contract instantiations. Added `vm.pauseGasMetering` to improve gas measurement control during deployment operations. Removed redundant remapping in `remappings.txt`.


foundry.toml
```diff
@@ -1,40 +1,43 @@
...
+evm_version = "cancun" # The EVM version to use during tests.
+via_ir = false # Whether to use the Solidity intermediate representation.
+optimizer = true # Whether to enable the Solidity optimizer.
+optimizer_runs = 4294967295 # The amount of optimizer runs to perform.
 bytecode_hash = "none" # Determines the hash method for the metadata hash that is appended to the bytecode
 cbor_metadata = false # Remove all metadata hashes from your contract's bytecode.
-evm_version = "paris" # The EVM version to use during tests.
-fuzz = { runs = 256 } # The amount of fuzz runs to perform for each fuzz test case.
-gas_reports = ["*"] # The contracts to print gas reports for.
 libs = ["node_modules", "lib"]
```
giuseppecrj, line 11: Remove lib

foundry.toml
```diff
@@ -44,7 +47,6 @@ int_types = "long"
 multiline_func_header = "all"
 number_underscore = "thousands"
 quote_style = "double"
-wrap_comments = true
-
+wrap_comments = false
```
giuseppecrj, line 50: Y?

giuseppecrj (approved):

foundry.toml
shuhuiluo: I did. https://github.com/towns-protocol/towns/pull/1119

foundry.toml
shuhuiluo: Want to use `forge fmt` instead of prettier.

foundry.toml
giuseppecrj: Might be running force install someplace?

foundry.toml
giuseppecrj: I've tried, prettier is better especially for argument wrapping

foundry.toml
shuhuiluo: `forge install` brings it back.

foundry.toml
shuhuiluo: What do you mean?
https://book.getfoundry.sh/reference/config/formatter#multiline_func_header

giuseppecrj: https://github.com/foundry-rs/foundry/discussions/3889

giuseppecrj: Doesn't work last time I tried using it. If this still is the case it's a nonstarter to switch

shuhuiluo: Consider changing the line length to 100. And I don't know why the tab width isn't 4.

> This is a balance between the old-school 80 character limit and the newer 120 character limit in the Solidity style guide. This makes it easier to read code on small or split screens.

https://github.com/Vectorized/solady/issues/19

foundry.toml
shuhuiluo: Actually we don't need `forge install` since we don't use git submodules.

PR #2615, opened 2025-03-17
Make node address optional in sync cookies

rhbntt:


texuf (commented):
Does the getstream code in forwarder.go still work now that the node address is optional?

core/node/rpc/forwarder.go
```diff
@@ -198,34 +198,79 @@ func (s *Service) getStreamImpl(
...
 			stub, err := s.nodeRegistry.GetStreamServiceClientForAddress(nodeAddress)
-			if err == nil {
-				ret, err := stub.GetStream(ctx, req)
-				if err != nil {
-					return nil, err
+			if err != nil {
+				return nil, err
+			}
+
+			ret, err := stub.GetStream(ctx, req)
```
bas-vk, line 242: Maybe we should consider passing in a context with a timeout. If one of the nodes is misbehaving and not returning a response within reasonable time we should try another node or in case of non replicated scenario leave it up to the client to retry it.

core/node/rpc/sync/client/syncer_set.go
```diff
@@ -202,57 +202,84 @@ func (ss *SyncerSet) AddStream(
...
-	if nodeAddress == ss.localNodeAddress {
-		var span trace.Span
-		if ss.otelTracer != nil {
-			_, span = ss.otelTracer.Start(ctx, "NewLocalSyncer",
-				trace.WithAttributes(attribute.String("stream", streamID.String())))
-		}
+		i := 0
+		remotes, useLocal := stream.GetRemotesAndIsLocal()
+		nodeAddress := stream.GetStickyPeer()
+		getNodeAddress = func() (common.Address, *SyncCookie, bool) {
```
bas-vk, line 219: There is similar functionality in the getStreamImp. Can you extract this functionality into reusable logic?

protocol/protocol.proto
```diff
@@ -781,7 +781,7 @@ message WrappedEncryptedData {
 }
 
 message SyncCookie  {
-    bytes node_address = 1;
+    optional bytes node_address = 1;
     bytes stream_id = 2;
     int64 minipool_gen = 3;
     int64 minipool_slot = 4;
```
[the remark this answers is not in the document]
texuf: in the case of a sync reset we reset and re-initialize the entire view. In the case of duplicate events, we just update the ref and move on. So all good on the client side.
image

sergekh2 (approved):

PR #2618, opened 2025-03-17
Pass first chunk in media stream creation request

rhbntt:


texuf (approved):

core/node/rpc/create_media_stream.go
```diff
@@ -48,6 +48,12 @@ func (s *Service) createMediaStream(ctx context.Context, req *CreateMediaStreamR
 		return nil, err
 	}
 
+	var firstChunkEvent *ParsedEvent
+	if len(parsedEvents) > 1 {
```
sergekh2, line 52: please modify to return error if there are more than 2 events (perhaps should be done in CanCreateStream for consistency).

Also, second event should have correct payload type.

sergekh2 (approved):

PR #2622, opened 2025-03-17
update to use latest crypto lib

giuseppecrj:


contracts/package.json
```diff
@@ -21,14 +21,14 @@
...
-        "crypto-lib": "github:towns-protocol/crypto-lib",
+        "crypto-lib": "github:towns-protocol/crypto-lib#bcc1e3d0655372a023e3d9d667787ba081e35a30",
         "solady": "^0.1.2"
     },
     "devDependencies": {
         "@prb/test": "^0.6.4",
         "@wagmi/cli": "^2.2.0",
         "account-abstraction": "github:eth-infinitism/account-abstraction",
-        "forge-std": "github:foundry-rs/forge-std#v1.9.5",
+        "forge-std": "github:foundry-rs/forge-std#v1.9.6",
```
shuhuiluo, line 31: Last time I tried forge-std v1.9.6 things broke.

contracts/package.json
```diff
@@ -21,14 +21,14 @@
         "@openzeppelin/contracts-upgradeable": "^5.2.0",
         "@prb/math": "^4.1.0",
         "@river-build/diamond": "github:river-build/diamond#552c970ed60cf452d28e7713b00a95c7bd10a5d9",
-        "crypto-lib": "github:towns-protocol/crypto-lib",
+        "crypto-lib": "github:towns-protocol/crypto-lib#bcc1e3d0655372a023e3d9d667787ba081e35a30",
```
texuf, line 24: we're not versioning this repo?

texuf (approved):

contracts/package.json
shuhuiluo: It's not on npm.

contracts/package.json
giuseppecrj: seems to be a thing with using super, reverted for now.

PR #2624, opened 2025-03-18
V2 delegate.cash implementation for the node

clemire:
On our second attempt at supporting delegate.xyz, we are having the node and client compute, for all linked wallets, the set of mainnet (and sepolia, for our testnet nodes) wallets that delegated to any linked wallet on the V1 DelegationRegistry deployed by delegate.xyz.

Testing here is done via a newly added unit test that rpc calls to Sepolia to find any delegator wallets. I have also added a new `linked-wallets` node command that can be run against any environment as long as chain rpc urls are provided for mainnet and, if needed, sepolia.


giuseppecrj (approved):

core/config/blockchain_info.go
```diff
@@ -12,7 +12,12 @@ type BlockchainInfo struct {
 	Name    string
 	// IsEtherBased is true for chains that use Ether as the currency for fees.
 	IsEtherBased bool
-	Blocktime    time.Duration
+	// IsEthereum applies to ethereum mainnet and all testnets. This setting is
+	// mainly used for testing features on our test networks, as towns mainnet
+	// should not have any test networks configured and only mainnet ethereum
+	// should qualify.
+	IsEthereum bool
```
texuf, line 19: I want to suggest that you call this IsMainnet instead of IsEthereum, so you can just stick with ethChainIds and mainnetChainIds, but it's also testnet... so idk. I guess this is fine.

core/xchain/entitlement/evaluator.go
```diff
@@ -7,14 +7,21 @@ import (
...
 
 type Evaluator struct {
 	clients        BlockchainClientPool
 	evalHistrogram *prometheus.HistogramVec
-	ethChainIds    []uint64
+	// etherBasedChainIds includes any chain that uses ethereum as a native currency.
+	etherBasedChainIds []uint64
+	// ethereumChainIds refers to the list of actual ethereum mainnet and testnets
+	// this network is configured to support.
+	ethereumChainIds []uint64
```
texuf, line 23: can you add a comment indicating that ethereumChainIds is a subset of etherBasedChainIds (or at least should be)

texuf (approved):

PR #2625, opened 2025-03-18
Delegate.cash client updates

clemire:
The client is now responsible for retrieving mainnet delegators and combining them with the list of linked wallets for the total list of linked wallets with delegations. This change requires a harmony update to ensure parsing and construction of XChainConfigs in harmony still works.

### DIFF
--- a/packages/web3/src/DelegateRegistry.ts
+++ b/packages/web3/src/DelegateRegistry.ts
@@ -72,16 +72,26 @@ export async function computeDelegatorsForProvider(
     wallets: string[],
 ): Promise<string[]> {
     const contract = newDelegateContract(provider)
-    const delegators: string[] = []
-    for (const wallet of wallets) {
-        const delegations = (await contract.callStatic.getDelegationsByDelegate(
-            wallet,
-        )) as DelegationInfo[]
-        for (const delegation of delegations) {
-            if (delegation.type_ == delegationTypeAll) {
-                delegators.push(delegation.vault)
-            }
-        }
-    }
-    return delegators
+    const delegatorWallets = (
+        await Promise.all(
+            wallets.map(async (wallet) => {
+                return (
+                    (
+                        (await contract.callStatic.getDelegationsByDelegate(
+                            wallet,
+                        )) as DelegationInfo[]
+                    )
+                        // Keep only delegations that cede the entire wallet
+                        .filter((info) => info.type_ == delegationTypeAll)
+                        // The 'vault' is the delegator wallet that cedes to one of wallets
+                        // passed in via the parameters
+                        .map((info) => info.vault)
+                )
+            }),
+        )
+    ).reduce((left, right) => [...left, ...right])
+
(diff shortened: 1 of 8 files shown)


packages/web3/src/DelegateRegistry.ts
```diff
@@ -0,0 +1,77 @@
...
+    const contract = newDelegateContract(provider)
+    const delegators: string[] = []
+    for (const wallet of wallets) {
+        const delegations = await contract.callStatic.getDelegationsByDelegate(wallet)
+        for (const delegation of delegations) {
+            if (delegation[0] == delegationTypeAll) {
+                delegators.push(delegation[1])
+            }
+        }
+    }
```
salzbrenner, line 75: would it make sense to do concurrent w/ Promise.all?

packages/web3/src/entitlement.ts
```diff
@@ -1263,6 +1267,20 @@ async function findEtherChainProviders(xchainConfig: XchainConfig) {
     return etherChainProviders
 }
 
+export async function findEthereumProviders(xchainConfig: XchainConfig) {
+    const ethereumProviders = []
+    for (const chainId of xchainConfig.ethereumChains) {
+        if (!(Number(chainId) in xchainConfig.supportedRpcUrls)) {
```
salzbrenner, line 1273: are these type conversions necessary? isn't chainId already number?

packages/web3/src/v3/SpaceDapp.ts
```diff
@@ -1687,7 +1703,7 @@ export class SpaceDapp implements ISpaceDapp {
         if (!space) {
             throw new Error(`Space with spaceId "${spaceId}" is not found.`)
         }
-        const linkedWallets = await this.getLinkedWalletsWithDelegations(owner)
+        const linkedWallets = await this.getLinkedWallets(owner)
```
salzbrenner, line 1706: why doesn't this one use getLinkedWalletsWithDelegations?

clemire: it needs an xchainconfig and I didn't want to change the signature for this call.

packages/web3/src/entitlement.ts
clemire: Honestly this is copy pasta from the above function, perhaps neither of them require it? I will test.

packages/web3/src/v3/SpaceDapp.ts
clemire: @nachoiacovino or @texuf do you have a sense of whether or not this would cause a problem?

texuf: other than tests, i only see one use, you can pull the xchain config from the context here

image

texuf: actually it's not currently in the context, but it can easily be added

texuf (approved):

PR #2645, opened 2025-03-20
refactor(contract): update drop facet base to be a lib

giuseppecrj:
This updates DropFacet to use libraries instead of base abstract contracts, functionality stays the same only typings changed.


contracts/src/airdrop/drop/DropFacetLib.sol
```diff
@@ -0,0 +1,235 @@
...
+    (uint48 newStartId, uint48 existingConditionCount) = (
+      self.conditionStartId,
+      self.conditionCount
+    );
+
+    if (uint256(newStartId) + conditions.length > type(uint48).max) {
+      CustomRevert.revertWith(
+        IDropFacetBase.DropFacet__CannotSetClaimConditions.selector
+      );
+    }
```
shuhuiluo, line 139: 1. This PR was merged without review.
2. The naming `newStartId` is misleading. It should be `existingStartId`. `newStartId` should be `existingStartId + existingConditionCount` if this function seeks to append new claim conditions.
3. Based on the code, the function overwrites existing claim conditions by copying the new conditions to the storage except `supplyClaimed`. But there is no check on whether `condition.currency == newCondition.currency`.
4. The function does `totalClaimableSupply += newCondition.maxClaimableSupply;` before `verifyEnoughBalance(newCondition.currency, totalClaimableSupply);`. But there is no guarantee that all new conditions have the same `currency`.
5. `self.conditionStartId` is never written to.
6. For fixes, please base them upon #3104.

giuseppecrj: Given we are the ones creating the claim conditions it was deemed not a necessary check. However I understand where you're coming from. Thanks for the PR.

PR #2648, opened 2025-03-20
refactor(contract): remove unused setups and streamline helper imports

shuhuiluo:
Eliminated redundant setup contracts and unnecessary functions in test files. Consolidated imports of `FacetHelper` and `DiamondHelper` to centralize logic, enhancing clarity and reducing code duplication.


contracts/scripts/deployments/facets/DeployDiamondCut.s.sol
```diff
@@ -8,7 +8,7 @@ pragma solidity ^0.8.23;
 //contracts
 import {Deployer} from "contracts/scripts/common/Deployer.s.sol";
 import {DiamondCutFacet} from "@towns-protocol/diamond/src/facets/cut/DiamondCutFacet.sol";
-import {FacetHelper} from "@towns-protocol/diamond/scripts/common/helpers/FacetHelper.s.sol";
```
giuseppecrj, line 11: why the change back? can you just update the one from the diamond repo?

giuseppecrj (changes requested):
we should be using diamond and facet based contracts from diamond repo

contracts/scripts/deployments/facets/DeployDiamondCut.s.sol
shuhuiluo: Yes.

giuseppecrj (approved):

PR #2659, opened 2025-03-21
updates to contract deployment system

giuseppecrj:


shuhuiluo (approved):

shuhuiluo:
> we should be using diamond and facet based contracts from diamond repo

https://github.com/towns-protocol/towns/pull/2648#pullrequestreview-2703576736

PR #2665, opened 2025-03-21
Fix broken link 

Hopium21:
Updated the incorrect file path in the inflation schedule reference to point to the correct location in the utils/ directory. Also ensured formatting consistency in the documentation. If there’s a better reference, let me know, and I’ll update it.


texuf:
contracts are in the packages folder now

PR #2666, opened 2025-03-21
feat(contract): Set stream replication factor in Stream Registry

bas-vk:
- [X] update the `Stream` storage layout and claim 1 byte for a stream replication factor
- [X] discuss how setting the stream replication factor for existing streams and implement solution
- [x] this PR made assumption that this function can be called by a configuration manager. Maybe we should pick a different role or add a new one?

This PR adds `StreamRegistry::setStreamReplicationFactor` that can be used to upgrade existing streams into replicated streams. The migration will be a 2 step process:

1. assign extra nodes to the stream in the registry while keeping the replication factor to 1. This will instruct the first node in the node list to be the leader while the other nodes in the list will start syncing miniblocks into their local state. To support this `setStreamReplicationFactor` doesn't check that the provides replication factor matches the length of the given nodes.
2. After a while when all nodes have synced the replication factor is set to the desired replication factor (probably 3 or 5)


contracts/src/river/registry/facets/stream/StreamRegistry.sol
```diff
@@ -340,6 +347,26 @@ contract StreamRegistry is IStreamRegistry, RegistryModifiers {
...
+
+    uint256 streamsCount = streamIds.length;
+
+    if (streamsCount == 0) revert(RiverRegistryErrors.BAD_ARG);
+
+    for (uint256 i; i < streamsCount; ++i) {
+      bytes32 streamId = streamIds[i];
+      _verifyStreamIdExists(streamId);
+      Stream storage stream = ds.streamById[streamId];
+      // TODO: emit StreamPlacementUpdated?
```
shuhuiluo, line 365: Yes.

contracts/src/river/registry/facets/stream/StreamRegistry.sol
```diff
@@ -340,6 +347,26 @@ contract StreamRegistry is IStreamRegistry, RegistryModifiers {
...
+      (stream.nodes, stream.replicationFactor) = (nodes, uint8(nodes.length));
```
shuhuiluo, line 366: Do we want to overwrite `stream.nodes` with `nodes`? Add a test with `replicationFactor > 1`.

shuhuiluo (changes requested):

contracts/src/river/registry/libraries/RegistryStorage.sol
```diff
@@ -9,11 +9,21 @@ import {RiverRegistryErrors} from "contracts/src/river/registry/libraries/Regist
 struct Stream {
   bytes32 lastMiniblockHash; // 32 bytes, slot 0
   uint64 lastMiniblockNum; // 8 bytes, part of slot 1
-  uint64 reserved0; // 8 bytes, part of slot 1
+  uint8 replicationFactor; // 1 byte, part of slot 1
```
giuseppecrj, line 12: we cannot add values to the middle of the struct in app storage, we have to add to the end

bas-vk: I thought it would be possible to split 64 bits into 56 and 8 bits without issues. But it gives an ABI incompatibility issue so I need to change this to use a bit mask as @shuhuiluo suggested.

giuseppecrj: Yep, that's the way to do what you want to do.

sergekh2 (approved):

shuhuiluo (approved):

PR #2676, opened 2025-03-22
Swap app registry <--> bot servers to use protobufs instead of JSON

clemire:
Move away from REST/JSON for the communication between the app registry service and bot servers and replace with REST/protobuf.


protocol/apps.proto
```diff
@@ -67,3 +70,31 @@ message GetStatusResponse {
...
+
+message AppServiceRequest {
+    repeated Envelope messages = 1;
+    repeated Envelope group_encryption_sessions_messages = 2;
+
+    message SolicitKeys {
+        repeated string session_ids = 1;
+        bytes stream_id = 2;
+    }
+
```
texuf, line 82: can you add some comments here explaining this request? I guess the thing that isn't obvious is that an undefined action is valid... and i'm wondering now if we should have a "message" action... or even rename action to payload...

texuf (approved):

protocol/apps.proto
clemire: Ah dangit, that was my other design. I read your design and thought that you wanted to always be able to piggy back messages on top of other actions...

protocol/apps.proto
```diff
@@ -67,3 +70,67 @@ message GetStatusResponse {
...
+    // SolicitKeys represents one or more messages in the same channel that cannot be forwarded to the
+    // app because the app does not have keys available to decrypt it. In this case, the app must send
+    // a message to the stream to solicit keys for it's device for the session ids that cannot yet be
+    // decrypted.
+    message SolicitKeysPayload {
+        repeated string session_ids = 1;
+        bytes stream_id = 2;
+    }
```
texuf, line 95: I don't think you need "Payload" on the nested messages, it becomes EventsPayload.MessagesPayload which is a mouthful

protocol/apps.proto
```diff
@@ -67,3 +70,67 @@ message GetStatusResponse {
...
+        SolicitKeysPayload solicitation = 2;
+    }
+}
+
+// The BatchEventsPayload will have a single set of messages and overlapping group encryption sessions
+// messages, or a set of key solicitations for session ids in the same channel. The BatchEventsPayload allows
+// many such groups to be combined into a single request in order to reduce webhook call frequency.
+message BatchEventsPayload {
+    repeated EventPayload events = 1;
+}
```
texuf, line 108: EventsPayload is probably fine (you can drop the "Batch")

protocol/apps.proto
```diff
@@ -67,3 +70,67 @@ message GetStatusResponse {
...
+// and channel messages for the bot to respond to in a single webhook call. Thus, an isolated
+// group of channel messages and encryption keys, or a request for a key solicitation, is
+// represented here. For webhook calls regarding channel content, the payload may contain one or
+// more of these events.
+message EventPayload {
+    // A MessagesPayload represents a group of user messages in a channel that qualifies for the app
+    // to be notified. The included set of group encryption sessions in this message should have
+    // all the needed ciphertexts to decrypt the set of messages sent in the same payload.
+    message MessagesPayload {
+        repeated Envelope messages = 1;
```
texuf, line 84: now that this is in a list, i think `repeated Envelope messages = 1;` can just be `Envelope message`

protocol/apps.proto
```diff
@@ -67,3 +70,67 @@ message GetStatusResponse {
...
+        repeated Envelope group_encryption_sessions_messages = 2;
```
texuf, line 85: can this just be `group_encryption_sessions`

texuf: also pretty sure we need a streamId here...

clemire: I think it's in the group_encryption_sessions_messages

protocol/apps.proto
clemire: discussed over towns, we will keep this so that we may batch multiple channel messages over a small set of ciphertexts.

protocol/apps.proto
clemire: discussed offline, we will add in stream id for clarity.

texuf (approved):

PR #2681, opened 2025-03-24
optimize getRandomNodes

giuseppecrj:


contracts/src/base/registry/facets/checker/EntitlementChecker.sol
```diff
@@ -218,21 +218,29 @@ contract EntitlementChecker is IEntitlementChecker, Facet {
...
+        uint256 rand = _pseudoRandom(i, remainingPool);
+        if (rand < count) {
+          // Use and update tracked index
+          randomNodes[i] = layout.nodes.at(indices[rand]);
+          indices[rand] = remainingPool - 1;
+        } else {
+          // Use large index directly
+          randomNodes[i] = layout.nodes.at(rand);
+        }
+        --remainingPool;
```
shuhuiluo, line 240: Does it prevent same `rand` to appear more than once when `rand >= count`?

shuhuiluo (changes requested):

contracts/src/base/registry/facets/checker/EntitlementChecker.sol
giuseppecrj: yeah, seems like current variation is good enough. Will revisit later.

PR #2686, opened 2025-03-25
Add prometheus metrics to StreamCache

rhbntt:


texuf (approved):

core/node/events/stream_cache.go
```diff
@@ -105,6 +107,23 @@ func NewStreamCache(params *StreamCacheParams) *StreamCache {
 			params.RiverChain.ChainId.String(),
 			params.Wallet.Address.String(),
 		),
+		loadStreamRecordDuration: params.Metrics.NewHistogramVecEx(
+			"stream_cache_load_duration_seconds",
+			"Load stream record duration",
+			infra.DefaultRpcDurationBucketsSeconds,
+			"chain_id", "address",
```
sergekh2, line 114: I think both chain id and address are redundant in this case:
we know chain from deployment (it's useful in generalized tx tracking, since there are multiple chains, but here it's clear what the chain is), and node is attributed during collection time anyway.

core/node/events/stream_cache.go
```diff
@@ -105,6 +107,23 @@ func NewStreamCache(params *StreamCacheParams) *StreamCache {
...
+		).WithLabelValues(
+			params.RiverChain.ChainId.String(),
+			params.Wallet.Address.String(),
+		),
+		loadStreamRecordCounter: params.Metrics.NewStatusCounterVecEx(
+			"stream_cache_load_counter",
+			"Number of stream record loads",
+			"chain_id", "address",
+		).MustCurryWith(prometheus.Labels{
+			"chain_id": params.RiverChain.ChainId.String(),
```
sergekh2, line 124: same

PR #2687, opened 2025-03-25
Introduce snapshot hash field in miniblock header

rhbntt:
This PR introduces the snapshot hash field with no logic around it. All changes around this field are going to be done in following PRs.


protocol/protocol.proto
```diff
@@ -144,6 +144,9 @@ message MiniblockHeader {
     // pointer to block with previous snapshot
     int64 prev_snapshot_miniblock_num = 7;
 
+    // hash of the snapshot.
+    bytes snapshot_hash = 8;
+
```
texuf, line 149: can we make this optional? I realize a 0 byte snapshot is essentially nil, but optional makes the frontend code much more clear what to expect.

texuf (approved):

PR #2688, opened 2025-03-25
Archiver to continuously download streams only from a hardcoded set of nodes

clemire:


core/node/rpc/archiver.go
```diff
@@ -568,16 +571,42 @@ func (a *Archiver) advanceNodeAndRetryWithDelay(stream *ArchiveStream, delay tim
...
+
 // ArchiveStream attempts to add all new miniblocks seen, according to the registry contract,
 // since the last time the stream was archived into storage.  It creates a new stream for
 // streams that have not yet been seen.
 func (a *Archiver) ArchiveStream(ctx context.Context, stream *ArchiveStream) (err error) {
+	log := logging.FromCtx(ctx)
+
+	if !isContinuousDownloadStream(stream) {
+		return
+	}
```
clemire, line 603:
```suggestion

```

PR #2689, opened 2025-03-25
simple change to testutils

giuseppecrj:


shuhuiluo (changes requested):
Can you upstream this to https://github.com/towns-protocol/diamond/blob/main/test/TestUtils.sol and import it?

PR #2715, opened 2025-03-31
Dockerized ponder gamma subgraphs

jterzis:
Dockerfile for ponder subgraph pointing to gamma network.

Also, added a docker-compose file to test with locally attached postgres.

CI job will follow this to host anointed docker images for subgraphs in ECR.

```
# build ponder image
docker build --no-cache -t river-subgraph -f packages/subgraph/Dockerfile .

# start subgraph with local attached postgrs pointing to gamma network in daemon mode
docker compose -f docker-compose-gamma.yml up -d

Navigate to http://localhost:42069/graphql to interact with the subgraph over the graphql api.
```


mechanical-turk (approved):
nice job. thanks for taking care of this

PR #2717, opened 2025-04-01
Proto changes according to snapshot migration

rhbntt:


texuf (approved):

protocol/protocol.proto
```diff
@@ -806,10 +805,8 @@ message StreamAndCookie {
     // if non-empty, contains all blocks since the latest snapshot, miniblocks[0].header is the latest snapshot
```
clemire, line 805: nit: this comment is a bit misleading

protocol/internode.proto
```diff
@@ -14,6 +14,7 @@ message MiniblockProposal {
 message AllocateStreamRequest {
     Miniblock miniblock = 1;
     bytes stream_id = 2; // stream_id should match the stream_id in the inception payload of the first event
+    optional Envelope snapshot = 3;
```
sergekh2, line 17: here and below: in proto3 optional for fields of type message is no-op (generated code is the same).
so it can be dropped for consistency.

(confusingly if has different meaning from proto2, in proto3 all fields are "optional" in proto2 sense; for scalars it defines if it's possible to distinguish between "not set" and "0")

protocol/protocol.proto
```diff
@@ -1032,6 +1034,7 @@ service StreamService {
     rpc GetLastMiniblockHash(GetLastMiniblockHashRequest) returns (GetLastMiniblockHashResponse);
     rpc AddEvent(AddEventRequest) returns (AddEventResponse);
     rpc AddMediaEvent(AddMediaEventRequest) returns (AddMediaEventResponse);
+    // TODO: rpc GetSnapshot(GetSnapshotRequest) returns (GetSnapshotResponse);
```
sergekh2, line 1037: lets just add it if we need it, i.e. TODO seems premature

sergekh2 (approved):

PR #2719, opened 2025-04-01
Do not remove inflight stream from sync

rhbntt:


packages/sdk/src/syncedStreamsLoop.ts
```diff
@@ -267,6 +267,10 @@ export class SyncedStreamsLoop {
             this.log('stream already in pending delete', streamId)
             return
         }
+        if (this.inFlightSyncCookies.has(streamId)) {
+            this.log('stream is in flight', streamId)
+            return
+        }
```
texuf, line 273: I don't think this is right. We need to fix it in the tick, something like 
```
private async tick() {
        if (this.syncState === SyncState.Syncing) {
            const pendingStreamsToDelete = this.pendingStreamsToDelete.filter(
                (x) => !this.inFlightSyncCookies.has(x),
            )
            if (
                (this.inFlightSyncCookies.size <= this.MIN_IN_FLIGHT_COOKIES &&
                    this.pendingSyncCookies.length > 0) ||
                pendingStreamsToDelete.length > 0
            ) {
                const syncId = this.syncId
                this.pendingSyncCookies.sort((a, b) => {
                    const aPriority = priorityFromStreamId(a, this.highPriorityIds)
                    const bPriority = priorityFromStreamId(b, this.highPriorityIds)
                    return aPriority - bPriority
                })
                const streamsToAdd = this.pendingSyncCookies.splice(0, this.MAX_IN_FLIGHT_COOKIES)
                const streamsToDelete = pendingStreamsToDelete.splice(
                    0,
                    this.MAX_IN_FLIGHT_STREAMS_TO_DELETE,
                )
                this.pendingStreamsToDelete = pendingStreamsToDelete.filter(
                    (x) => !streamsToDelete.find((y) => x === y),
                )
```

texuf (approved):

PR #2726, opened 2025-04-02
Pass snapshot field through components and APIs 

rhbntt:
- Updated create storage functions in the storage interface. 
- Added snapshot to the miniblocks reader function signature. It is always nil for now.
- Populated snapshot in some proto models.
- Added snapshot when saving miniblock candidate
- Updated MakeGenesisMiniblock to return snapshot 
- Replaced Snapshot field type inside WriteMiniblockData type to bytes for further use
- Added snapshot to the get miniblocks response 
- Updated RemoteMiniblockProvider interface
- Force populating snapshot to APIs


core/cmd/stream_cmd.go
```diff
@@ -647,14 +651,23 @@ func runStreamGetCmd(cmd *cobra.Command, args []string) error {
 	stream := response.Msg.GetStream()
 	fmt.Println("MBs: ", len(stream.GetMiniblocks()), " Events: ", len(stream.GetEvents()))
 
-	for _, mb := range stream.GetMiniblocks() {
-		i, err := events.NewMiniblockInfoFromProto(mb, events.NewParsedMiniblockInfoOpts().WithDoNotParseEvents(true))
+	for i, mb := range stream.GetMiniblocks() {
+		var snapshot *protocol.Envelope
+		if i == 0 {
```
clemire, line 656: Does the stream response include only one snapshot?

core/cmd/stream_cmd.go
```diff
@@ -647,14 +651,23 @@ func runStreamGetCmd(cmd *cobra.Command, args []string) error {
...
+			events.NewParsedMiniblockInfoOpts().
+				WithDoNotParseEvents(true),
+		)
 		if err != nil {
 			return err
 		}
 
-		fmt.Print(i.Ref, "  ", i.Header().GetTimestamp().AsTime().Local())
-		if i.Header().GetSnapshot() != nil {
+		fmt.Print(info.Ref, "  ", info.Header().GetTimestamp().AsTime().Local())
```
clemire, line 669: 👍🏻

core/node/events/miniblock.go
```diff
@@ -46,10 +46,13 @@ func Make_GenesisMiniblockHeader(parsedEvents []*ParsedEvent) (*MiniblockHeader,
 	}, nil
 }
 
-func MakeGenesisMiniblock(wallet *crypto.Wallet, genesisMiniblockEvents []*ParsedEvent) (*Miniblock, error) {
+func MakeGenesisMiniblock(
+	wallet *crypto.Wallet,
+	genesisMiniblockEvents []*ParsedEvent,
+) (*Miniblock, *Envelope, error) {
```
clemire, line 52: Can you either name these return parameters or document them above the function definition?

core/node/events/miniblock_info.go
```diff
@@ -16,17 +16,20 @@ import (
...
+	Proto    *Miniblock
+	Snapshot *Envelope
+
 	headerEvent        *ParsedEvent
 	useGetterForEvents []*ParsedEvent // Use events(). Getter checks if events have been initialized.
-	Proto              *Miniblock
+	snapshot           *ParsedEvent   //nolint:unused
 }
 
 // NewMiniblockInfoFromProto initializes a MiniblockInfo from a proto, applying validation based
```
clemire, line 28: I think this comment is now slightly out of date.

clemire (approved):

core/cmd/stream_cmd.go
clemire: In this case, is it guaranteed to always be block 0?

core/node/events/miniblock_info.go
```diff
@@ -249,16 +253,19 @@ func (b *MiniblockInfo) lastEvent() *ParsedEvent {
...
-	return b.Header().GetSnapshot() != nil || len(b.Header().GetSnapshotHash()) > 0
+// IsSnapshot returns non-nil but empty byte array if the miniblock is a snapshot.
+// TODO: Will be replaced with the real snapshot after DB migration. WIP.
+func (b *MiniblockInfo) IsSnapshot() []byte {
+	if b.Header().GetSnapshot() != nil || len(b.Header().GetSnapshotHash()) > 0 {
+		return make([]byte, 0)
+	}
+
+	return nil
 }
```
texuf, line 264: weird that isSnapshot nolonger returns bool - should we change this to GetSnapshot()?

texuf (approved):

PR #2727, opened 2025-04-02
allow state transition to operational from remote-only

jterzis:


shuhuiluo (approved):

giuseppecrj (approved):

giuseppecrj:
What's the order of operations in testing this? gamma first or is this a hot fix?

jterzis:
> What's the order of operations in testing this? gamma first or is this a hot fix?

hot fix

PR #2728, opened 2025-04-02
typescript 5.8

miguel-nascimento:
Small changes, yarn.lock and doc changes due to docgen script update is good amount of diff here.
Harmony PR: https://github.com/HereNotThere/harmony/pull/9365

What's new from 5.1 to 5.8?

[Inferred Type Predicates](https://devblogs.microsoft.com/typescript/announcing-typescript-5-5/#inferred-type-predicates): `xs.filter(x => x !== undefined)` now refines the type to remove the undefined from it. Reduces the usage of `!`
[Narrowing On Comparisons to Booleans](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-3.html#narrowing-on-comparisons-to-booleans)
[`using` declaration and explicit resource management support](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-2.html#using-declarations-and-explicit-resource-management) (nice in theory but in practice it's bad, don't even be hyped about it)
[Decorator Metadata](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-2.html#decorator-metadata)
[Object.groupBy/Map.groupBy](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-4.html#objectgroupby-and-mapgroupby)
[lsp level: Comma Completions for Object Members](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-2.html#comma-completions-for-object-members)
[`NoInfer` utility type](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-4.html#the-noinfer-utility-type) - tldr: better error message from libraries
[Better granular checks for branches in return expressions](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-8.html#granular-checks-for-branches-in-return-expressions)

and a lot of type level optimizations :p

Probably a good follow up for this is eslint v8 -> v9 (previous PR https://github.com/towns-protocol/towns/pull/1065 tried to do that, but omg it's a lot to do in a single take)

I also changed a bit the signature of react-sdk `useObservable` hook. The previous signature was causing issue with the docgen script, probably due to subtyping.


packages/react-sdk/package.json
```diff
@@ -59,7 +58,7 @@
     ],
     "peerDependencies": {
         "react": "^18.2.0",
-        "typescript": "^5.1.6"
```
miguel-nascimento, line 62: actually this should be 5.x

packages/dlog/src/dlog.ts
```diff
@@ -169,14 +169,26 @@ const makeDlog = (d: Debugger, opts?: DLogOpts): DLogger => {
...
+                    }
                     newArgs.push(cloneAndFormat(c, { shortenHex: true }))
                 }
             } else {
-                isSingleLineLogsMode ? fmt.push('%o ') : fmt.push('%O ')
+                if (isSingleLineLogsMode) {
+                    fmt.push('%o ')
+                } else {
+                    fmt.push('%O ')
+                }
```
texuf, line 191: why? did ternaries change?

packages/docs/sdk/react-sdk/api/connectTownsWithBearerToken.mdx
```diff
@@ -16,7 +16,7 @@ import { connectTownsWithBearerToken } from '@towns-protocol/react-sdk'
 ```ts
 function connectTownsWithBearerToken(
   token: string,
-  config: Omit<SyncAgentConfig, 'context'>,
+  config: Omit<SyncAgentConfig, "context">,
```
texuf, line 19: double quotes now?

packages/docs/sdk/react-sdk/api/useCreateChannel.mdx
```diff
@@ -14,15 +14,15 @@ import { useCreateChannel } from '@towns-protocol/react-sdk'
...
-  config?: ActionConfig<Space['createChannel']>,
+  config?: ActionConfig<Space["createChannel"]>,
 ): {
     data: string | undefined;
     error: Error | undefined;
     isPending: boolean;
     isSuccess: boolean;
     isError: boolean;
     createChannel: (channelName: string, signer: Signer, opts?: {
-        topic?: string | undefined;
```
texuf, line 25: this is a different signature... is this also a docs update?

packages/dlog/src/dlog.ts
miguel-nascimento: eslint added a new default rule about unused expressions. i.e ternary to run side effects. I disabled it on harmony, I will update and disable it here too

packages/docs/sdk/react-sdk/api/connectTownsWithBearerToken.mdx
miguel-nascimento: I think that under the hood the typescript compiler is using double quotes to stringify. We're using the TS compiler to get those signatures.

packages/docs/sdk/react-sdk/api/useCreateChannel.mdx
miguel-nascimento: `topic?: string` is the same as `topic?: string | undefined`. Typescript Eslint added a new default rule that removes the explicit undefined from the latter.

miguel-nascimento: >  is this also a docs update?

I had to update the api-extractor package, which changed a bit some stuff (i.e the single quote vs double quote in docs types)

I also had to simplify `useObservable` return type in react-sdk, since by some reason it was hitting an error where the api-extractor wasn't able to extract the type, failing to generate docs

texuf (approved):

packages/dlog/src/dlog.ts
texuf: honestly - that's pretty nice. you can leave this.

PR #2751, opened 2025-04-04
chore: add access public to npm packages

miguel-nascimento:
I tried setting it in lerna.conf, but I didn't worked. Tested locally, I was able to publish to npm with this change.


texuf:
I wonder if now that they are pushed subsequent pushes will work

texuf (approved):

PR #2771, opened 2025-04-07
revert(contract): remove use of forge fmt due to slow experience in vscode

giuseppecrj:
forge fmt is slow when using formatOnSave in vscode, will revert until a solution is found.


shuhuiluo (approved):
Lgtm.

PR #2775, opened 2025-04-08
Handle new snapshot format

rhbntt:


clemire (approved):
I'm not deep in this, but everything appears to be in order here.

texuf (approved):

PR #2797, opened 2025-04-10
Advance multi_ne RPC port

clemire:
Currently the justfile assigns a port to the app registry that is RPC_PORT + NUM_INSTANCES, taking the next available port in sequence after all stream nodes are assigned. This was causing an issue, however, because the RPC_PORT env setting for multi and multi_ne are apart by exactly 10, meaning that the app registry launched with the multi configuration attempts to take the same port as node 0 of the multi_ne config. There are many possible solutions here, but this one simply advances the lowest assigned RPC_PORT for multi_ne config by 10 to create more headroom.


sergekh2 (approved):

core/env/local/multi_ne/env.env
```diff
@@ -1,4 +1,4 @@
 RUN_ENV=multi_ne
 DISABLE_BASE_CHAIN=true
 NUM_INSTANCES=10
-RPC_PORT=5180
```
texuf, line 4: A different fix would be to add APP_REGISTRY_RPC_PORT to multi/env.env and multi_ne/env.env
I want to start running the notifications server locally as well, which could follow the same convention
This is a good quick fix though. Up to you.

texuf (approved):

core/env/local/multi_ne/env.env
clemire: I'll merge for now to unblock, we can circle back.

PR #2800, opened 2025-04-10
web3: Optimize Banned Wallet Checks with Caching and Multicall

salzbrenner:
This PR significantly optimizes the performance of checking and retrieving banned wallet information in `SpaceDapp` by introducing caching and utilizing multicall.

Before this PR, loading AX1 would result in ~1300 rpc calls related to banned wallets. Now, dependent on the user and their linked wallets, it will make a maximum of 10 calls.

Can probably further clean up banning/tokenId logic across this package and consumers in harmony, but this is enough for now imo.

**Key Changes:**

*   **Added Caching:** Implemented `SimpleCache` for:
    *   `bannedTokenIds`: Caches the list of banned token IDs for a space.
    *   `ownerOfToken`: Caches the owner address for a specific token ID.
    *   `isBannedToken`: Caches the banned status of a specific token ID.
*   **Optimized `bannedWalletAddresses`:**
    *   Refactored to first check caches (`bannedTokenIdsCache`, `ownerOfTokenCache`).
    *   Uses `multicall` to efficiently fetch owners for any uncached banned token IDs.
    *   Updates the `ownerOfTokenCache` after fetching.
*   **Optimized `walletAddressIsBanned`:**
    *   Now uses `isBannedTokenCache` for faster lookups.
*   **Cache Invalidation:**
    *   Added `updateCacheAfterBanOrUnBan` helper method.
    *   `banWalletAddress` and `unbanWalletAddress` now call this helper to clear relevant cache entries after performing the action.
*   **Optimized Entitlement Checks:**
    *   `getEntitledWalletForJoiningSpaceUncached` and `isEntitledToChannelUncached` now use the cached `walletAddressIsBanned` check via `Promise.any` for faster checks on linked wallets, avoiding fetching the full banned list unnecessarily.
*   **Refactoring:**
    *   Moved cache key classes (`EntitlementRequest`, `BannedTokenIdsRequest`, `OwnerOfTokenRequest`, `IsTokenBanned`) to `Keyable.ts`.

### DIFF
--- a/packages/web3/src/ISpaceDapp.ts
+++ b/packages/web3/src/ISpaceDapp.ts
@@ -374,4 +374,5 @@ export interface ISpaceDapp {
         receipt: ContractReceipt,
         senderAddress: string,
     ) => TipEventObject | undefined
+    updateCacheAfterBanOrUnBan: (spaceId: string, tokenId: ethers.BigNumber) => void
 }
--- a/packages/sdk/src/tests/multi/entitlements/channelsWithEntitlements.test.ts
+++ b/packages/sdk/src/tests/multi/entitlements/channelsWithEntitlements.test.ts
@@ -53,6 +53,27 @@ describe('channelsWithEntitlements', () => {
             ),
         ).resolves.toBeFalsy()
 
+        // unban alice
+        const unbanTx = await bobSpaceDapp.unbanWalletAddress(
+            spaceId,
+            alicesWallet.address,
+            bobProvider.wallet,
+        )
+        await unbanTx.wait()
+
+        // Wait 5 seconds for the caches to expire
+        await new Promise((f) => setTimeout(f, 5000))
+
+        await expect(
+            aliceSpaceDapp.isEntitledToChannel(
+                spaceId,
+                channelId!,
+                alice.userId,
+                Permission.Read,
+                getXchainConfigForTesting(),
+            ),
+        ).resolves.toBeTruthy()
+
         const doneStart = Date.now()
         // kill the clients
         await bob.stopSync()
--- a/packages/sdk/src/tests/multi/entitlements/spaceWithEntitlements.test.ts
+++ b/packages/sdk/src/tests/multi/entitlements/spaceWithEntitlements.test.ts
@@ -64,6 +64,25 @@ describe('spaceWithEntitlements', () => {
         )
         expect(entitledWallet).toBeUndefined()
 
+        // unban alice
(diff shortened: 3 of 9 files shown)


packages/web3/src/ISpaceDapp.ts
```diff
@@ -374,4 +374,5 @@ export interface ISpaceDapp {
         receipt: ContractReceipt,
         senderAddress: string,
     ) => TipEventObject | undefined
+    updateCacheAfterBanOrUnBan: (spaceId: string, tokenId: ethers.BigNumber) => void
```
clemire, line 377: This hook seems like an implementation detail of the SpaceDapp and I think it doesn't really need to be exposed to harmony, is that correct?

clemire (changes requested):
Could you add 1 or 2 tests to spaceWithEntitlements.test.ts and channelWithEntitlements.test.ts to validate the behavior of the banning cache, including how it clears?

packages/web3/src/ISpaceDapp.ts
salzbrenner: no, i need it for userops which doesn't use SpaceDapp writes

salzbrenner:
> Could you add 1 or 2 tests to spaceWithEntitlements.test.ts and channelWithEntitlements.test.ts to validate the behavior of the banning cache, including how it clears?

Yeah, will do. Thought there were banning tests already. I ran these changes against harmony/lib integration tests and it was fine, so should be all good

clemire:
> > Could you add 1 or 2 tests to spaceWithEntitlements.test.ts and channelWithEntitlements.test.ts to validate the behavior of the banning cache, including how it clears?
> 
> Yeah, will do. Thought there were banning tests already. I ran these changes against harmony/lib integration tests and it was fine, so should be all good

I think you're right. I was thinking, it might make sense to test more complicated behaviors, but this could be good enough already since they do actually wait for the node cache to clear as well.

clemire (approved):

salzbrenner:
> > > Could you add 1 or 2 tests to spaceWithEntitlements.test.ts and channelWithEntitlements.test.ts to validate the behavior of the banning cache, including how it clears?
> > 
> > 
> > Yeah, will do. Thought there were banning tests already. I ran these changes against harmony/lib integration tests and it was fine, so should be all good
> 
> I think you're right. I was thinking, it might make sense to test more complicated behaviors, but this could be good enough already since they do actually wait for the node cache to clear as well. I'll leave it up to you!

I extended the mentioned tests

PR #2806, opened 2025-04-11
Support new snapshot format in SDK

rhbntt:


packages/sdk/src/sign.ts
```diff
@@ -217,6 +231,27 @@ export const unpackEnvelope = async (
...
+
+    const sn = fromBinary(SnapshotSchema, snapshot.event)
+    let hash = snapshot.hash
+
+    if (opts?.disableHashValidation !== true) {
+        hash = riverSnapshotHash(snapshot.event)
+        check(bin_equal(hash, snapshot.hash), 'Snapshot id is not valid', Err.BAD_EVENT_ID)
+    }
+
+    // TODO: Check signature?
```
texuf, line 250: in order to check the signature we need the header that the snapshot was originally linked to... is that accessible from this point in the code? can we pass it to this function?

packages/sdk/src/client.ts
```diff
@@ -2193,6 +2193,7 @@ export class Client
             miniblocks,
             'backward',
         )
+        await this.persistenceStore.saveSnapshots(streamIdAsString(streamId), snapshots)
```
texuf, line 2197: We don't want to save these snapshots in persistence - the client only cares about the latest snapshot.

packages/sdk/src/persistenceStore.ts
```diff
@@ -105,12 +106,15 @@ export interface IPersistenceStore {
...
     ): Promise<ParsedMiniblock[]>
+    saveSnapshot(streamId: string, miniblockNum: bigint, snapshot: Snapshot): Promise<void>
+    saveSnapshots(streamId: string, snapshots: Record<string, Snapshot>): Promise<void>
 }
 
 export class PersistenceStore extends Dexie implements IPersistenceStore {
     cleartexts!: Table<{ cleartext: Uint8Array | string; eventId: string }>
     syncedStreams!: Table<{ streamId: string; data: Uint8Array }>
     miniblocks!: Table<{ streamId: string; miniblockNum: string; data: Uint8Array }>
+    snapshots!: Table<{ streamId: string; miniblockNum: string; data: Uint8Array }>
```
texuf, line 117: we don't want to save these snapshots - it's gigs of data that the client will never read.

packages/sdk/src/client.ts
texuf: we do save the snapshots rolling forward though... 
[https://github.com/towns-protocol/towns/blob/638fc7a132adb1b8cfd34f3b990e7ddb93c6e4a3/packages/sdk/src/syncedStream.ts#L1](https://github.com/towns-protocol/towns/blob/638fc7a132adb1b8cfd34f3b990e7ddb93c6e4a3/packages/sdk/src/syncedStream.ts#L172)

packages/sdk/src/streamUtils.ts
```diff
@@ -106,6 +106,7 @@ export function parsedMiniblockToPersistedMiniblock(
 ) {
     if (direction === 'backward') {
         miniblock.header.snapshot = undefined
+        miniblock.header.snapshotHash = undefined
```
texuf, line 109: i think you can delete this line `        miniblock.header.snapshotHash = undefined` - it's fine to know that the header has a snapshot, we were only deleting the snapshot because it's so huge.

packages/sdk/src/sign.ts
```diff
@@ -217,6 +231,31 @@ export const unpackEnvelope = async (
...
+    let hash = snapshot.hash
+
+    if (opts?.disableHashValidation !== true) {
+        hash = riverSnapshotHash(snapshot.event)
+        check(bin_equal(hash, snapshot.hash), 'Snapshot id is not valid', Err.BAD_EVENT_ID)
+    }
+
+    const doCheckEventSignature = opts?.disableSignatureValidation !== true
+    if (doCheckEventSignature) {
+        checkEventSignature(headerEvent, hash, snapshot.signature)
```
texuf, line 253: would love to either see a comment here that the headerEvent contains the createrAddress of the snapshot OR, call this like so 
```
checkEventSignature({
        creatorAddress: headerEvent.creatorAddress
        delegateSig: headerEvent.delegateSig
        delegateExpiryEpochMs: headerEvent.delegateExpiryEpochMs
    },
    hash,
    snapshot.signature
}
```
for clarity

texuf (approved):

PR #2816, opened 2025-04-12
Fix handling of empty genesis miniblocks and add recovery logic for c…

sergekh2:
…orrupted stream records.


core/node/events/stream_cache.go
```diff
@@ -477,6 +447,18 @@ func (s *StreamCache) loadStreamRecordImpl(
 		return nil, err
 	}
 
+	if record.StreamId() != streamId || !slices.Contains(record.Nodes(), s.params.Wallet.Address) {
```
bas-vk, line 450: Why the check for `record.StreamId() != streamId`?

bas-vk (approved):

Issue #2818, opened 2025-04-13
Local failing streams are not retried on remotes.


sergekh2:
### Description

In forwarder code, if stream is not local, and remote fails, it failovers to another replica.

However, if streams is local, and fails (while generally doesn't happen, happened recently due to data corruption bug), there is no retry to remote. However, in this case if it would failover to remote, it would successfully return data.

Tricky part: which error codes should failover? (Same problem exists in regular failover in forwarder.)

PR #2821, opened 2025-04-14
Fallback to remotes on local failure

rhbntt:


core/node/rpc/forwarder.go
```diff
@@ -24,6 +25,21 @@ const (
...
+		Err_UNKNOWN,
+		Err_DEADLINE_EXCEEDED,
+		Err_NOT_FOUND,
+		Err_RESOURCE_EXHAUSTED,
+		Err_ABORTED,
+		Err_UNIMPLEMENTED,
+		Err_INTERNAL,
+		Err_UNAVAILABLE,
+		Err_DATA_LOSS,
+		Err_BUFFER_FULL,
```
bas-vk, line 39: `Err_STREAM_RECONCILIATION_REQUIRED`

texuf (approved):

core/node/rpc/forwarder.go
```diff
@@ -45,6 +62,10 @@ func allowNoQuorum[T any](req *connect.Request[T]) bool {
 	return req.Header().Get(RiverAllowNoQuorumHeader) == RiverHeaderTrueValue
 }
 
+func shouldFallbackToRemotes(err error) bool {
+	return slices.Contains(shouldFallbackToRemotesErrCodes, AsRiverError(err).Code)
+}
```
bas-vk, line 67: Error codes are used together with `RiverErrorImpl`. A better place is put this functionality is in the base package next to `RiverErrorImpl`. This is more localised and other parts of the code base can use it when they use `RiverErrorImpl`. Now it's private to the rpc package.

Suggestion: delete `shouldFallbackToRemotesErrCodes` and `shouldFallbackToRemotes`. Introduce `base#IsOperationRetriableOnRemotes(err) bool` that follows the same pattern as some of the existing error checks.

core/node/base/error.go
```diff
@@ -464,3 +466,21 @@ func TruncateErrorToConnectLimit(err error) error {
...
+		protocol.Err_NOT_FOUND,
+		protocol.Err_RESOURCE_EXHAUSTED,
+		protocol.Err_ABORTED,
+		protocol.Err_UNIMPLEMENTED,
+		protocol.Err_INTERNAL,
+		protocol.Err_UNAVAILABLE,
+		protocol.Err_DATA_LOSS,
+		protocol.Err_BUFFER_FULL,
+		protocol.Err_STREAM_RECONCILIATION_REQUIRED,
+	}, AsRiverError(err).Code)
```
bas-vk, line 485: If `AsRiverError` receives an unsupported error type it defaults to `protocol.Err_UNKNOWN` making it retriable on remotes which might not always be true.

Everywhere `RiverErrorImpl` should be used. I think if `errors.As(err, &myRiverErr)` and `myRiverErr.Code` is in the list it should return true. This is not perfect but more precise.

bas-vk (approved):

core/node/base/error.go
clemire: I know this is already merged... but: it would be helpful to have the above discussion in comments here to provide context when reading the code.

clemire:
I also need to remember this advice, but please @rbhnt provide a short PR explanation that includes any specific outage information this PR is a response to! It would help me to stay up to date on what's already fixed for replication and give a little more context since I missed the conversations you and @bas-vk had when you were working closely together.

clemire:
> @clemire check the linked development issue (#2818), the description is inside the issue. Would you like to get some more context?

Ah, my mistake! I didn't think to check there, thank you!

PR #2831, opened 2025-04-15
App Registry Server: Filter sent messages with forward settings

clemire:
3 forward settings supported:
- no messages
- MENTIONS_REPLIES_REACTIONS - the logic is the same here as it is for notifications
- all messages - include all events in the channel, including Membership events like Key Solicitations, Usernames, etc

As part of this PR, we also disallow forwarding for bot-authored messages.


protocol/apps.proto
```diff
@@ -15,17 +15,39 @@ import "protocol.proto";
...
 message RegisterRequest {
     // public key of the app
     bytes app_id = 1;
 
     // public key of the app owner
     bytes app_owner_id = 2;
+
+    // The setting for determining which messages the app registry service should forward to the app. If unspecified,
+    // the default setting will be used.
+    optional ForwardSettingValue forward_setting = 3;
```
texuf, line 50: why optional?

texuf: I think it would be worth it to make a proper Settings message, A SetSettingsRequest/Response, where:
```
message AppSettings {
    ForwardSettingValue forward = 1
}
```

I would bet money that we will have at least one more setting added to this service.

texuf (approved):

PR #2844, opened 2025-04-16
ponder imports typings via @towns-protocol/contracts/typings

mechanical-turk:


packages/contracts/package.json
```diff
@@ -5,6 +5,9 @@
   "installConfig": {
     "hoistingLimits": "dependencies"
   },
+  "files": [
+    "./typings/index.ts"
+  ],
```
shuhuiluo, line 10:
```suggestion
  "files": [
    "src/**/*.sol",
    "scripts/**/*.sol",
    "./typings/index.ts",
    "README.md"
  ],
```

PR #2847, opened 2025-04-17
Introduce snapshots interval in miniblocks config

rhbntt:
This PR introduces a new onchain setting to control snapshot retention frequency. The setting determines how often snapshots are persisted (e.g. every N-th miniblock). It supports delayed activation and avoids aggressive pruning when the setting changes—retaining at least one snapshot per historical range. This lays the groundwork for efficient stream trimming while ensuring backward compatibility and preventing operator error.


bas-vk (commented):
Add `StreamSnapshotIntervalInMiniblocks` to `func TestSetOnChain(t *testing.T)`

bas-vk (approved):
Please add a comment that 0 means trimming is disabled.

texuf (approved):

PR #2849, opened 2025-04-17
update local ponder setup

giuseppecrj:


jterzis (approved):
tested on gamma, looks like you solved the import issue with @towns-protocol/contracts/typings, thanks.

PR #2855, opened 2025-04-17
support per-chunk-encryption in stream-metadata-service

erikolsson:


packages/stream-metadata/src/riverStreamRpcClient.ts
```diff
@@ -113,19 +113,34 @@ async function mediaContentFromStreamView(
...
-		mediaInfo.chunks.reduce((totalLength, chunk) => totalLength + chunk.length, 0),
-	)
-	let offset = 0
-	mediaInfo.chunks.forEach((chunk) => {
-		data.set(chunk, offset)
-		offset += chunk.length
-	})
+	let decrypted: Uint8Array
+	if (mediaInfo.perChunkEncryption) {
+		decrypted = new Uint8Array()
```
miguel-nascimento, line 118: can't you initialize this buffer with the total chunk size?
 
then instead of creating a new Uint8Array in every iteration of the loop you would just append the existing one

PR #2859, opened 2025-04-18
feat: tip support in sync agent timeline

miguel-nascimento:
TODO add sendTip and add tests


miguel-nascimento:
Closed in favor of https://github.com/towns-protocol/towns/pull/2879, since its moving the entire timeline into towns sdk (adding proper support for fully read markers)

PR #2863, opened 2025-04-18
Snapshots trimming mechanism

rhbntt:


core/node/storage/pg_snapshot_trimmer.go
```diff
@@ -0,0 +1,129 @@
...
+	if n == 0 {
+		return nil
+	}
+
+	// find max seq and compute cutoff below which we may nullify
+	maxSeq := snapshotSeqs[n-1]
+	cutoff := maxSeq - minKeep
+
+	// group by bucketID = seq / retentionInterval
+	buckets := make(map[int64][]int64, n)
```
sergekh2, line 103: It seems it should be possible to do in one pass without extra data structure?
I.e. on forward pass over snapshot numbers you always know which bucket you are in from index, if new bucket -> update bucket number, skip current snapshot number; otherwise add snapshot number to nullify; do not add last snapshot to nullify list (before processing slice input list to [:len(-1)]

core/node/storage/pg_snapshot_trimmer.go
```diff
@@ -0,0 +1,238 @@
...
+	streamId StreamId,
+	lastMbNum int64,
+	retentionInterval int64,
+) error {
+	// Collect all miniblocks with a snapshot for the given stream starting from the last processed miniblock number.
+	// Genesis miniblock must not be trimmed.
+	snapshotMiniblockRows, err := tx.Query(
+		ctx,
+		st.storage.sqlForStream(
+			"SELECT seq_num FROM {{miniblocks}} WHERE stream_id = $1 AND seq_num > $2 AND snapshot IS NOT NULL",
```
sergekh2, line 137: as a general observation - this is non-index read due to IS NOT NULL, and as such is relatively expensive, i.e. it needs to be scheduled quite rarely. It would be cool to avoid it entirely for non-updated streams. Maybe we can put something is es table? (Probably we can start with current implementation and implement it as upgrade on top) (this will also eliminate the need to keep extra data in memory)

think few million streams - loop in trimStreams will take multiple hours.

core/node/storage/pg_snapshot_trimmer.go
```diff
@@ -0,0 +1,238 @@
...
+	// Reset snapshot field for the given miniblocks
+	if _, err = tx.Exec(
+		ctx,
+		st.storage.sqlForStream(
+			"UPDATE {{miniblocks}} SET snapshot = NULL WHERE stream_id = $1 AND seq_num = ANY($2)",
+			streamId,
+		),
+		streamId, toNullify,
+	); err != nil {
+		fmt.Println(err)
```
sergekh2, line 180:
```suggestion
		
```

core/node/storage/pg_stream_store.go
```diff
@@ -1347,6 +1359,7 @@ func (s *PostgresStreamStore) WriteMiniblocks(
 		return RiverError(Err_INTERNAL, "No miniblocks to write").Func("pg.WriteMiniblocks")
 	}
 	if prevMinipoolGeneration != miniblocks[0].Number {
+		fmt.Println(prevMinipoolGeneration, miniblocks[0].Number)
```
sergekh2, line 1362:
```suggestion
		
```

core/node/storage/pg_snapshot_trimmer.go
```diff
@@ -0,0 +1,238 @@
...
+
+// trimStream trims the snapshots for the given stream.
+func (st *snapshotTrimmer) trimStream(
+	ctx context.Context,
+	tx pgx.Tx,
+	streamId StreamId,
+	lastMbNum int64,
+	retentionInterval int64,
+) error {
+	// Collect all miniblocks with a snapshot for the given stream starting from the last processed miniblock number.
```
sergekh2, line 132: `lockStream` should be called here - otherwise this tx can get inconsistent view of the data for this stream if other txs are modifying it. (or rather updates are going to be applied when other txes potentially modified stream data)

core/node/storage/pg_snapshot_trimmer.go
```diff
@@ -0,0 +1,273 @@
...
+	return st, nil
+}
+
+// monitorWorkerPool monitors the worker pool and handles shutdown
+func (st *snapshotTrimmer) monitorWorkerPool(ctx context.Context) {
+	for {
+		select {
+		case <-ctx.Done():
+			st.log.Info("Worker pool stopped due to context cancellation")
+			st.workerPool.StopWait()
```
sergekh2, line 82: it shouldn't be StopWait here and below - pending task should be canceled, i.e. close should be fast.

Also, it doesn't really own worker polls (since it's passed as param), but closes it anyway?

core/node/storage/pg_stream_store.go
```diff
@@ -2240,6 +2273,23 @@ func (s *PostgresStreamStore) getLastMiniblockNumberTx(
 	return maxSeqNum, nil
 }
 
+// getLowestStreamMiniblock retrieves the lowest miniblock number for a given stream
+func (s *PostgresStreamStore) getLowestStreamMiniblock(
```
sergekh2, line 2277:
```suggestion
func (s *PostgresStreamStore) getLowestStreamMiniblockTx(
```

core/node/storage/pg_snapshot_trimmer.go
```diff
@@ -0,0 +1,273 @@
...
+	}
+
+	maxSeq := snapshotSeqs[n-1]
+	cutoff := maxSeq - minKeep
+
+	var toNullify []int64
+	var seenBucket int64 = -1
+
+	// Reverse‐scan: newest first
+	for i := n - 1; i >= 0; i-- {
```
sergekh2, line 253: I think it should keep first snapshot in the bucket, reason for that: it's possible to rebuild latter snapshot from the preceding snapshot and events, but if oldest snapshot is deleted, it's not possible to rebuild it.

core/node/storage/pg_snapshot_trimmer.go
```diff
@@ -0,0 +1,273 @@
...
+	ctx context.Context,
+	tx pgx.Tx,
+	streamId StreamId,
+) {
+	// Retention interval could be changed at any time so the current value must be used
+	retentionInterval := int64(st.config.Get().StreamSnapshotIntervalInMiniblocks)
+	if retentionInterval == 0 {
+		// If retention interval is 0, it means that snapshots trimming is disabled
+		return
+	}
```
sergekh2, line 111: I think it shouldn't schedule if workerpool if "full", i.e. if there are 10k items already scheduled, do not schedule new one - it's not keeping up (otherwise it can blow out memory).

True for stream trimmer as well.

core/node/storage/pg_snapshot_trimmer.go
```diff
@@ -0,0 +1,273 @@
...
+	var mbNum int64
+	if _, err = pgx.ForEachRow(snapshotMiniblockRows, []any{&mbNum}, func() error {
+		mbs = append(mbs, mbNum)
+		return nil
+	}); err != nil {
+		return err
+	}
+
+	// Determine which miniblocks should be nullified
+	toNullify := determineSnapshotsToNullify(mbs, task.retentionInterval, st.minKeep)
```
sergekh2, line 207: judging from experience with delete, this update might need to be limited in size in one tx: otherwise potentially it can take very long time.

core/node/storage/pg_snapshot_trimmer.go
```diff
@@ -0,0 +1,263 @@
...
+	maxSeq := snapshotSeqs[n-1]
+	cutoff := maxSeq - minKeep
+
+	var toNullify []int64
+	var lastBucket int64 = -1
+
+	for _, seq := range snapshotSeqs {
+		// skip anything in the protected tail
+		if seq > cutoff {
+			continue
```
sergekh2, line 251:
```suggestion
			break
```

sergekh2 (approved):

PR #2875, opened 2025-04-20
Update useObservable to memo the callback function

texuf:
otherwise it recreates it every frame and loops 🤦


miguel-nascimento (approved):
🤦‍♂️

PR #2879, opened 2025-04-20
WIP - move timelines store to river and drop parsed timeline events

texuf:
River timelines, no more parsed stream events. This still needs to be broken up more, but plane is landing and wanted to get it up in case anyone wants to take a look


packages/sdk/src/streams-view/streamsView.ts
```diff
@@ -0,0 +1,72 @@
...
+import { StreamChange } from '../streamEvents'
+import { LocalTimelineEvent, StreamTimelineEvent } from '../types'
+import { TimelineStore, TimelineStoreDelegate } from './timelineStore'
+import { SnapshotCaseType } from '@towns-protocol/proto'
+
+export type StreamsViewDelegate = TimelineStoreDelegate
+
+// a view of all the streams
+export class StreamsView {
+    readonly timelineStore: TimelineStore
```
texuf, line 12: timeline store should probably be timeline view? jtol

PR #2890, opened 2025-04-23
refactor(script): simplify logic for block time and mining settings

shuhuiluo:
Simplified logic for setting block time and mining modes in local scripts. Made `BASE_EXECUTION_CLIENT` explicit for conditional automine and interval mining adjustments, improving clarity and maintainability.

### DIFF
--- a/scripts/deploy-base-contracts.sh
+++ b/scripts/deploy-base-contracts.sh
@@ -39,4 +39,4 @@ make interact-any-local context=$RIVER_ENV rpc=base_anvil contract=InteractClaim
 # Utils
 make deploy-any-local context=$RIVER_ENV rpc=base_anvil type=utils contract=DeployMember
 make deploy-any-local context=$RIVER_ENV rpc=base_anvil type=utils contract=DeployMockNFT
-make deploy-any-local context=$RIVER_ENV rpc=base_anvil type=utils contract=DeployEntitlementGatedExample 
+make deploy-any-local context=$RIVER_ENV rpc=base_anvil type=utils contract=DeployEntitlementGatedExample
--- a/scripts/deploy-contracts.sh
+++ b/scripts/deploy-contracts.sh
@@ -42,10 +42,8 @@ fi
 # make deploy-any-local context=$RIVER_ENV rpc=base_anvil type=utils contract=DeployEntrypoint
 # make deploy-any-local context=$RIVER_ENV rpc=base_anvil type=utils contract=DeployAccountFactory
 
-# Only anvil supports automine but this might be a local geth node
-if [ "${BASE_EXECUTION_CLIENT}" != "geth_dev" ]; then
-    cast rpc evm_setAutomine true --rpc-url $BASE_ANVIL_RPC_URL
-fi
+# Deploying contracts with automine is faster
+cast rpc evm_setAutomine true --rpc-url $BASE_ANVIL_RPC_URL
 cast rpc evm_setAutomine true --rpc-url $RIVER_ANVIL_RPC_URL
 
 # Deploy base contracts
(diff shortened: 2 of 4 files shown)


scripts/deploy-base-contracts.sh
```diff
@@ -13,12 +13,19 @@ set +a
...
+: ${BASE_EXECUTION_CLIENT:?}
 
 # Build if not called with nobuild
 if [ "${1-}" != "nobuild" ]; then
     yarn turbo build --filter=@towns-protocol/contracts
 fi
 
+# Deploying contracts with automine is faster
+# Only anvil supports automine but this might be a local geth node
+if [ "${BASE_EXECUTION_CLIENT}" != "geth_dev" ]; then
```
giuseppecrj, line 25: we should not be moving to using geth.

giuseppecrj (changes requested):

scripts/deploy-base-contracts.sh
shuhuiluo: Done.

scripts/deploy-base-contracts.sh
```diff
@@ -19,6 +19,9 @@ if [ "${1-}" != "nobuild" ]; then
     yarn turbo build --filter=@towns-protocol/contracts
 fi
 
+# Deploying contracts with automine is faster
+cast rpc evm_setAutomine true --rpc-url $BASE_ANVIL_RPC_URL
```
giuseppecrj, line 23: isn't automine the default?

shuhuiluo: Yes.

texuf (approved):

giuseppecrj (approved):

scripts/deploy-base-contracts.sh
```diff
@@ -39,4 +42,7 @@ make interact-any-local context=$RIVER_ENV rpc=base_anvil contract=InteractClaim
 # Utils
 make deploy-any-local context=$RIVER_ENV rpc=base_anvil type=utils contract=DeployMember
 make deploy-any-local context=$RIVER_ENV rpc=base_anvil type=utils contract=DeployMockNFT
-make deploy-any-local context=$RIVER_ENV rpc=base_anvil type=utils contract=DeployEntitlementGatedExample 
\ No newline at end of file
+make deploy-any-local context=$RIVER_ENV rpc=base_anvil type=utils contract=DeployEntitlementGatedExample
+
+# Restore to interval mining
+cast rpc evm_setIntervalMining $RIVER_BLOCK_TIME --rpc-url $BASE_ANVIL_RPC_URL
```
salzbrenner, line 48: this is actually going to make Userops_tests.yml in harmony much slower, or maybe even break because we can't evm_setIntervalMining to 0.1, which is what that ci job does. 

Optimizing that ci job was actually why i made this whole script separate in the first place - I moved these deployments out of deploy-contracts.sh

I think this script is always going to be called from deploy-contracts, with the exception of the mentioned ci job, so can we just keep the automining management in deploy-contracts?

scripts/deploy-contracts.sh
```diff
@@ -42,21 +42,14 @@ fi
...
-# Only anvil supports automine but this might be a local geth node
-if [ "${BASE_EXECUTION_CLIENT}" != "geth_dev" ]; then
-    cast rpc evm_setAutomine true --rpc-url $BASE_ANVIL_RPC_URL
-fi
-cast rpc evm_setAutomine true --rpc-url $RIVER_ANVIL_RPC_URL
-
 # Deploy base contracts
 "$SCRIPT_DIR/deploy-base-contracts.sh" nobuild
 
+cast rpc evm_setAutomine true --rpc-url $RIVER_ANVIL_RPC_URL
```
salzbrenner, line 48: related to my other comment, if we just move this to line 44 then it should be fine

scripts/deploy-base-contracts.sh
shuhuiluo: Thanks. Fixed.

scripts/deploy-contracts.sh
```diff
@@ -42,10 +42,8 @@ fi
 # make deploy-any-local context=$RIVER_ENV rpc=base_anvil type=utils contract=DeployEntrypoint
 # make deploy-any-local context=$RIVER_ENV rpc=base_anvil type=utils contract=DeployAccountFactory
 
-# Only anvil supports automine but this might be a local geth node
-if [ "${BASE_EXECUTION_CLIENT}" != "geth_dev" ]; then
-    cast rpc evm_setAutomine true --rpc-url $BASE_ANVIL_RPC_URL
-fi
+# Deploying contracts with automine is faster
+cast rpc evm_setAutomine true --rpc-url $BASE_ANVIL_RPC_URL
```
salzbrenner, line 46: duplicated line below?

salzbrenner (approved):

scripts/deploy-contracts.sh
salzbrenner: nvm

PR #2905, opened 2025-04-28
node: use ForAllStreamsOnNode instead of ForAllStreams

bas-vk:
Use `StreamRegistry::getPaginatedStreamsOnNode` instead of `StreamRegistry::getPaginatedStreams` where possible. This will only load streams from the current node instead of iterating over all streams and grab streams on the local node.

### DIFF
--- a/core/node/registries/river_registry_contract.go
+++ b/core/node/registries/river_registry_contract.go
@@ -358,39 +358,50 @@ var ZeroBytes32 = [32]byte{}
 func (c *RiverRegistryContract) callGetPaginatedStreams(
 	ctx context.Context,
 	blockNum crypto.BlockNumber,
+	node *common.Address,
 	start int64,
 	end int64,
-) ([]river.StreamWithId, bool, error) {
+) ([]river.StreamWithId, error) {
 	if c.Settings.SingleCallTimeout > 0 {
 		var cancel context.CancelFunc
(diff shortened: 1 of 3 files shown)


texuf (approved):

core/node/registries/river_registry_contract.go
```diff
@@ -422,6 +464,21 @@ func waitForBackoff(ctx context.Context, bo backoff.BackOff) bool {
...
+// ForAllStreamsOnNode calls the given cb for all streams that are registered in the river registry at the given block
+// num and on the given node. If cb returns false ForAllStreamsOnNode returns.
+func (c *RiverRegistryContract) ForAllStreamsOnNode(
+	ctx context.Context,
+	blockNum crypto.BlockNumber,
+	node common.Address,
+	cb func(*river.StreamWithId) bool,
+) error {
+	if c.Settings.ParallelReaders > 1 {
+		return c.forAllStreamsOnNodeParallel(ctx, blockNum, node, cb)
```
sergekh2, line 476: I think it's possible to have only one version of the core pagination logic:
extend `forAllStreamsParallel` and `forAllStreamsSingle` to take node address as an argument. Pass it down to `callGetPaginatedStreams`; there if it's 0 call `GetPaginatedStreams`, otherwise `GetPaginatedStreamsOnNode`. 

This way this change will be more compact, and critically there will be only one version of pagination logic, which is more maintainable.

sergekh2 (approved):

PR #2906, opened 2025-04-28
SpaceDapp: use multicall in getJoinSpacePriceDetails

salzbrenner:
reduced call time on my machine (dev build) from about 1200ms -> 300ms


texuf (approved):
nice

PR #2909, opened 2025-04-28
Fix potential erroneous false entitlement check result

clemire:
If a nested entitlement contains two on-chain checks and chain unavailability causes rpc requests to lag until the context times out, these could have erroneously come back as a false entitlement result instead of an error result. I think it's very likely this would happen any time a chain or rpc provider is partially unavailable. This PR repairs the error for both AND and OR operators in entitlement trees.

### DIFF
--- a/core/xchain/entitlement/entitlement.go
+++ b/core/xchain/entitlement/entitlement.go
@@ -27,16 +27,16 @@ func (e *Evaluator) EvaluateRuleData(
 	return e.evaluateOp(ctx, opTree, linkedWallets)
 }
 
-// isEntitlementEvaluationError returns true iff the error is the result of a failure when evaluating
+// isNoncancelationError returns true iff the error is the result of a failure when evaluating
 // an entitlement. It ignores context cancellations, which can occur when a operation evaluation
 // short-circuits because the other child returned a definitive answer.
-func isEntitlementEvaluationError(err error) bool {
+func isNoncancelationError(err error) bool {
 	return err != nil && !errors.Is(err, context.Canceled)
 }
 
 // logIfEntitlementError conditionally logs an error if it was not a context cancellation.
 func logIfEntitlementError(ctx context.Context, err error) {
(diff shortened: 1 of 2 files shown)


core/xchain/entitlement/entitlement.go
```diff
@@ -45,15 +45,24 @@ func logIfEntitlementError(ctx context.Context, err error) {
...
 		return rightErr
 	}
+
+	// If both contexts were cancelled, assume this is due to rpc timeout and return this as an error.
+	// Note: for nested entitlements, we only manually trigger a cancellation if one of the entitlement
+	// checks resulted in a definitive answer, so we should never see two cancellations unless the parent
+	// context was itself cancelled.
+	if errors.Is(leftErr, context.Canceled) && errors.Is(rightErr, context.Canceled) {
+		return fmt.Errorf("%w; %w", leftErr, rightErr)
+	}
```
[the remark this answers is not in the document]
texuf: oh wait i get it.

core/xchain/entitlement/entitlement.go
```diff
@@ -27,34 +27,45 @@ func (e *Evaluator) EvaluateRuleData(
...
-	return nil
+
+	// If either error is nil, allow the other error to dictate the response
+	if leftErr == nil {
+		return rightErr
+	}
+
+	if rightErr == nil {
+		return leftErr
+	}
```
texuf, line 65: I don't understand this - it cancels out the checks above, the whole thing is now just `return leftErr || rightErr`

clemire: No, it prioritizes returning non-cancellation errors over errors that are related to deadlines or cancellations.

1. If both are noncancellation errors, compose and return
2. if either are noncancellation errors, return the noncancellation error
3. If only one error is (possibly) non-nil, then return the other error, which is either nil or a cancellation error
4. If both errors are cancellation errors, compose them and return

texuf (approved):

PR #2916, opened 2025-04-29
refactor(contract): burn towns points on claim

giuseppecrj:


packages/contracts/src/airdrop/drop/DropClaimLib.sol
```diff
@@ -142,19 +148,31 @@ library DropClaimLib {
...
-            // Compute the keccak256 hash of the previous hash (stored at memory location 0) and
-            // store it
-            // in the leaf variable
-            leaf := keccak256(0, 0x20)
+    function createLeaf(
+        address member,
+        uint256 amount,
+        uint256 points
+    ) internal pure returns (bytes32 leaf) {
+        assembly {
```
shuhuiluo, line 156:
```suggestion
        assembly ("memory-safe") {
```

> While we recommend to always respect Solidity’s memory model, inline assembly allows you to use memory in an incompatible way. Therefore, moving stack variables to memory and additional memory optimizations are, by default, globally disabled in the presence of any inline assembly block that contains a memory operation or assigns to Solidity variables in memory.

https://docs.soliditylang.org/en/latest/assembly.html#memory-safety

packages/contracts/src/airdrop/drop/DropClaimLib.sol
```diff
@@ -142,19 +148,31 @@ library DropClaimLib {
...
+            // Load free memory pointer
+            let ptr := mload(0x40)
```
shuhuiluo, line 158: You can cache `mload(0x40)` to a stack variable, use memory [0, 0x60) and restore 0x40 later. `mstore(0x40, add(ptr, 0x60))` is unnecessary because the allocated memory won't be used again.

shuhuiluo (changes requested):

packages/contracts/test/airdrop/DropFacet.t.sol
```diff
@@ -245,27 +263,35 @@ contract DropFacetTest is BaseSetup, IDropFacetBase, IOwnableBase, IRewardsDistr
     }
 
     // claimWithPenalty
-    function test_claimWithPenalty_fuzz(ClaimData[] memory claimData) external {
+    function test_fuzz_claimWithPenalty(ClaimData[] memory claimData) external {
```
shuhuiluo, line 266:
```suggestion
    /// forge-config: default.fuzz.runs = 64
    function test_fuzz_claimWithPenalty(ClaimData[] memory claimData) external {
```
In-line test configuration to confine fuzz runs on tests with unbound array inputs.
https://book.getfoundry.sh/reference/config/inline-test-config

shuhuiluo (changes requested):

packages/contracts/test/airdrop/DropFacet.t.sol
giuseppecrj: This is dope! thanks @shuhuiluo

giuseppecrj:
@coderabbitai review

packages/contracts/src/airdrop/drop/DropClaimLib.sol
```diff
@@ -142,19 +148,32 @@ library DropClaimLib {
...
-            // in the leaf variable
-            leaf := keccak256(0, 0x20)
+            // Cache the free memory pointer
+            let origPtr := mload(0x40)
+            let ptr := 0x00
+
+            // Pack address + amount + points into memory [0, 0x60)
+            mstore(ptr, shl(96, member)) // member at ptr
+            mstore(add(ptr, 0x20), amount) // amount at ptr+32
+            mstore(add(ptr, 0x40), points) // points at ptr+64
```
shuhuiluo, line 164:
```suggestion
        assembly ("memory-safe") {
            let fmp := mload(0x40)
            mstore(0, member)
            mstore(0x20, amount)
            mstore(0x40, points)
            leaf := keccak256(0, 0x60)
            mstore(0, leaf)
            leaf := keccak256(0, 0x20)
            mstore(0x40, fmp)
        }
```
I'm afraid this conflicts with OpenZeppelin's JS library since they use `abi.encode` instead of `abi.encodePacked`. The current implementation is also not `abi.encodePacked`. If you want `abi.encodePacked`, it should be
```solidity
            mstore(ptr, shl(96, member)) // member at ptr
            mstore(add(ptr, 0x14), amount) // amount at ptr+20
            mstore(add(ptr, 0x34), points) // points at ptr+52
```
https://github.com/OpenZeppelin/merkle-tree?tab=readme-ov-file#leaf-hash

packages/contracts/test/airdrop/DropFacet.t.sol
```diff
@@ -40,6 +41,7 @@ contract DropFacetTest is BaseSetup, IDropFacetBase, IOwnableBase, IRewardsDistr
     struct ClaimData {
         address claimer;
         uint16 amount;
+        uint16 points;
     }
```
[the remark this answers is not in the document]
shuhuiluo: Good catch.

shuhuiluo (approved):
Lgtm.

PR #2926, opened 2025-04-30
Fixed snapshot miniblock index selection in MakeRemoteStreamView

rbhnt:


core/node/events/stream_view.go
```diff
@@ -134,7 +134,7 @@ func MakeRemoteStreamView(stream *StreamAndCookie) (*StreamView, error) {
 		}
 	}
 
-	snapshot := miniblocks[snapshotIndex].GetSnapshot()
+	snapshot := miniblocks[0].GetSnapshot()
```
texuf, line 137: this doesn't seem right to me, but just in case, if you change this you need to also fix up the index in the return value
```
	return &StreamView{
		streamId:      streamId,
		blocks:        miniblocks,
		minipool:      newMiniPoolInstance(minipoolEvents, generation, eventNumOffset),
		snapshot:      snapshot,
		snapshotIndex: snapshotIndex,
	}, nil
```

core/node/events/snapshot.go
```diff
@@ -903,14 +915,18 @@ func findUserMembership(
...
 		streamId,
 		bytes.Compare,
 		func(membership *UserPayload_UserMembership) []byte {
 			return membership.StreamId
 		},
 	)
+	if err != nil {
+		return nil, AsRiverError(err).Func("findUserMembership")
+	}
+	return md, nil
```
texuf, line 929: thanks for adding these errors - as far as i know, these errors should just be ignored and not cause any disruption in the snapshotting flow.

core/node/events/stream_view.go
texuf: at some point it was not like that though: https://github.com/towns-protocol/towns/blob/e98b70c542815c6527efad0b11a3412212aedfb2/core/node/events/stream_view.go

texuf (approved):

core/node/events/stream_view.go
sergekh2: I don't recall why ATM. Different data sources? We should be able to unify them.

sergekh2: Re: this fix. Seems unrelated to the problem?

PR #2929, opened 2025-04-30
Ping remote syncer before considering as an available fallback node

rhbntt:


core/node/rpc/sync/client/syncer_set.go
```diff
@@ -339,32 +343,28 @@ func (ss *SyncerSet) modify(ctx context.Context, req ModifyRequest) error {
...
-				for _, streamIDRaw := range modifySync.GetRemoveStreams() {
-					req.RemovingFailureHandler(&SyncStreamOpStatus{
-						StreamId: streamIDRaw,
-						Code:     int32(rvrErr.Code),
-						Message:  rvrErr.GetMessage(),
-					})
+				var err error
+				syncer, err = ss.newRemoteSyncer(nodeAddress)
+				if err != nil {
+					rvrErr := AsRiverError(err).Tag("remoteSyncerAddr", syncer.Address())
```
clemire, line 352: nit: if there is an error here, are we guaranteed that syncer is not nil and has valid fields? Perhaps it would be better to use `nodeAddress` here instead of `syncer.Address()` for annotating the error.

core/node/rpc/sync/client/syncer_set.go
```diff
@@ -214,7 +209,8 @@ func (ss *SyncerSet) modify(ctx context.Context, req ModifyRequest) error {
 
 	modifySyncs := make(map[common.Address]*ModifySyncRequest)
 
-	// group cookies by node address
+	// Group modify sync request by the remote syncer.
+	// Identifying which node to use for the given streams.
```
clemire, line 213: 🔥

clemire (approved):

PR #2932, opened 2025-04-30
Add user/{userid}/tips route to user metadata service

texuf:
returns tips for any user, cached every 30 minutes
add test

### DIFF
--- a/packages/stream-metadata/src/routes/userTips.ts
+++ b/packages/stream-metadata/src/routes/userTips.ts
@@ -1,0 +1,95 @@
+import { FastifyReply, FastifyRequest } from 'fastify'
+import { StreamPrefix, StreamStateView, makeStreamId } from '@towns-protocol/sdk'
+import { z } from 'zod'
+
+import { getStream } from '../riverStreamRpcClient'
+import { isValidEthereumAddress } from '../validators'
+
+const paramsSchema = z.object({
+	userId: z.string().min(1, 'userId parameter is required'),
+})
+
+export interface UserTips {
+	sent: Record<string, string>
+	received: Record<string, string>
+	sentCount: Record<string, string>
+	receivedCount: Record<string, string>
+}
+
+export interface UserTipsResponsePayload {
+	tips: UserTips
+}
+
+const CACHE_CONTROL = {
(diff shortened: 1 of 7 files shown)


packages/stream-metadata/src/routes/userTips.ts
```diff
@@ -0,0 +1,95 @@
...
+
+const paramsSchema = z.object({
+	userId: z.string().min(1, 'userId parameter is required'),
+})
+
+export interface UserTips {
+	sent: Record<string, string>
+	received: Record<string, string>
+	sentCount: Record<string, string>
+	receivedCount: Record<string, string>
```
miguel-nascimento, line 16: what exactly is the sent / received field?
a map that stores what? userId? eth value? 

how about tip currency?

packages/stream-metadata/tests/integration/userTips.test.ts
```diff
@@ -0,0 +1,156 @@
...
+		const aliceRoute = `user/${alicesUserId}/tips`
+		log(`Fetching Alice's tips from ${baseURL}/${aliceRoute}`)
+		const aliceResponse = await axios.get<UserTipsResponsePayload>(`${baseURL}/${aliceRoute}`)
+
+		expect(aliceResponse.status).toBe(200)
+		expect(aliceResponse.headers['content-type']).toContain('application/json')
+		expect(aliceResponse.data.tips.received).toEqual(tipAmount.toString())
+		expect(aliceResponse.data.tips.receivedCount).toEqual('1')
+		expect(aliceResponse.data.tips.sent).toEqual('0')
+		expect(aliceResponse.data.tips.sentCount).toEqual('0')
```
miguel-nascimento, line 103: received is the tip amount received in the ETH_CURRENCY, what if someone tip in a different currency? are we filtering it in the stream view userContent to be only ETH_CURRENCY?

texuf: yes that's why it's a map :) right now we only tip in eth in the app, but you can tip in anything

packages/stream-metadata/src/routes/userTips.ts
miguel-nascimento: got it! thought it was userId 😅

PR #2935, opened 2025-04-30
tests: remove timeout when debugging

miguel-nascimento:
We don't want to resume a debugger and get a timeout error. 

That's why this PR changes the tests timeout when debugging.


texuf (approved):
Nice!!!!

PR #2937, opened 2025-04-30
chore: migrate libolm to vodozemac

miguel-nascimento:
This migrates our Olm implementation from`libolm` to the newer and maintained Rust one: `vodozemac`

Should close TOWNS-25923

## Why:

1. libolm got [deprecated in July 2024](https://gitlab.matrix.org/matrix-org/olm#important-libolm-is-now-deprecated)
2. the wasm packaging is bad. It was blocking us to ship 100% ESM builds. That's why we have esbuild in stream metadata and stress tests.

I did our own bindings to vodozemac: [@towns-protocol/vodozemac](https://github.com/towns-protocol/vodozomac-bindings), still yet to be published to npm.

It correctly bundles WebAssembly, enabling us to: 
- remove the esbuild from our apps build step. 
- allow Nextjs users to use our sdk
- possibly usage of `bun` on our apps like stream-metadata and future bots

## Migration steps

Vodozemac can pickle and unpickle accounts in the libolm (pickle is undocumented btw)
However, I couldn't find a pickle in libolm format for `Session` and `GroupSessions`

Still to be done.


packages/encryption/src/encryptionDevice-vodozemac.ts
```diff
@@ -0,0 +1,969 @@
...
+    HybridGroupSessionKeySchema,
+    PlainMessage,
+} from '@towns-protocol/proto'
+import { exportAesGsmKeyBytes, generateNewAesGcmKey } from './cryptoAesGcm'
+import { Dexie } from 'dexie'
+import { create, fromBinary, toBinary } from '@bufbuild/protobuf'
+
+const log = dlog('csb:encryption:encryptionDevice')
+
+const PICKLE_KEY = 'DEFAULT_PICKLE_KEY_1234567890___'
```
miguel-nascimento, line 40: Changed to a string that's 32 bytes.

packages/encryption/src/encryptionDevice-vodozemac.ts
miguel-nascimento, line 1: I can replace the current libolm encryptionDevice with the vodozemac impl, this way github will render a nice diff here. Lmk you guys want this.

packages/encryption/src/encryptionDevice.ts
```diff
@@ -583,26 +566,27 @@ export class EncryptionDevice {
         sessionKey: string,
         keysClaimed: Record<string, string>,
         _exportFormat: boolean,
-        extraSessionData: GroupSessionExtraData = {},
+        extraSessionData: GroupSessionExtraDataVodozemac = {},
     ): Promise<void> {
+        await initAsync()
```
texuf, line 571: ^ this initAsync seems out of place

texuf (commented):
Plan to ship
- can we drop the "Vodozemac" on the new functions, add "OlmLib" to the old functions
- ensure that we can load existing hybrid and group encryption sessions from crypto db and use them with vodozemac
- A) ensure that old messages in channel can still be decrypted when loading existing sessions from crypto db
- B) ensure that old messages in channel can be decrypted after soliciting keys
- ensure that messages can be encrypted and subsequently decrypted by a) other web clients using vodozemac, b) other web clients still using olm, c) ios clients (which we don't plan on upgrading)
- a) figure out pickle differences for accounts - if the user has to create a new account, this is fine, as long as we can read pickled sessions (maybe there's a matrix account migrator we can implement?)
- b) figure out pickle differences for sessions - hybrid group sessions are not pickled, but old olm group sessions are, and we would like to still be able to decrypt messages with these.

it might be possible to write a javascript parser for old pickled sessions (will need to cut and paste from matrix libs) or find matrix session migration code

ALSO - spend a tiny amount of time, not that you're a packaging expert, to see if we can properly package the previous olm lib

PR #2938, opened 2025-04-30
chore: move `@connectrpc/connect-node` to peer dependency

miguel-nascimento:
This allow us to correctly handle and support both web and node environments.

Previously, this was making web clients downloading (and building!) node dependencies.


packages/sdk/src/rpcCommon.ts
```diff
@@ -1,25 +1,32 @@
...
-            createConnectTransport: createConnectTransportNode,
-            // eslint-disable-next-line import/no-extraneous-dependencies, @typescript-eslint/no-var-requires, @typescript-eslint/no-require-imports
-        } = require('@connectrpc/connect-node')
-        // eslint-disable-next-line @typescript-eslint/no-unsafe-call
-        return createConnectTransportNode({ ...options, httpVersion: '2' }) as Transport
+export async function createHttp2ConnectTransport(
+    options: ConnectTransportOptionsWeb,
+): Promise<Transport> {
+    try {
+        if (isNodeEnv) {
```
texuf, line 15: the previous code ensured that we use the browser version in the test suite - which might be why the tests are now failing?

texuf (approved):

PR #2940, opened 2025-04-30
wrap the tipping transaction so that it retries during tests

texuf:
do we want this?


salzbrenner (approved):
if it's causing issues, sure

PR #2942, opened 2025-05-01
add `getStreamMetadataUrl` in sdk

miguel-nascimento:
So we can use it in other places such as the bot framework.


packages/sdk/src/riverConfig.ts
```diff
@@ -128,3 +128,18 @@ export function makeRiverConfig(inEnvironmentId?: string) {
...
+        case 'alpha':
+            return 'https://alpha.river.delivery'
+        case 'gamma':
+            return 'https://gamma.river.delivery'
+        case 'omega':
+            return 'https://river.delivery'
+        case 'delta':
+            return 'https://delta.river.delivery'
+        default:
+            return 'http://localhost:3002'
```
texuf, line 143: this one can be 3003 if you're running local_multi_ne

PR #2948, opened 2025-05-01
ci: bump actions/setup-go to v5

reject-i:
Updates GitHub workflows to use `actions/setup-go@v5`, bringing the Node 20-based runtime and keeping the workflows on the latest maintained major version.

[Ref](https://github.com/actions/setup-go?tab=readme-ov-file#v5)


texuf:
Closing for now, @sergekh2 please re-open if you think this is useful.

PR #2949, opened 2025-05-01
[WIP] Shared syncers

rhbntt:
Implemented a shared remote syncer between node A and B for stream sync operations


core/node/rpc/sync/client/message_router.go
```diff
@@ -0,0 +1,199 @@
...
+	syncIDs, _ := r.streamSubscriptions.LoadOrStore(streamID, xsync.NewMap[string, struct{}]())
+	syncIDs.Store(syncID, struct{}{})
+	r.streamSubscriptions.Store(streamID, syncIDs)
+}
+
+func (r *messageRouter) removeStreamFromSubscription(syncID string, streamID StreamId) {
+	if subscribers, ok := r.streamSubscriptions.Load(streamID); ok {
+		subscribers.Delete(syncID)
+		if subscribers.Size() == 0 {
+			r.streamSubscriptions.Delete(streamID)
```
bas-vk, line 83: There is a race condition with `messageRouter#addStreamToSubscription` that can call `Store` after the size check and before the delete.

In general the usage of streamSubscriptions looks racy over the various functions unless messageRouter will be locked externally.

core/node/rpc/sync/client/syncer_set.go
```diff
@@ -434,57 +484,162 @@ func (ss *SyncerSet) modify(ctx context.Context, req ModifyRequest) error {
...
+			SyncOp:   SyncOp_SYNC_UPDATE,
+			Stream:   sc,
+			StreamId: cookie.GetStreamId(),
+		}); err != nil {
+			return AsRiverError(err).Func("startSyncingByCookie")
+		}
+	}
+
+	// Add the stream to the subscription
+	ss.router.addStreamToSubscription(syncID, streamID)
```
bas-vk, line 581: Does the client miss an update when a new event was added in between `GetStreamSince` was called and the stream was added to the subscription?

PR #2954, opened 2025-05-01
Add miniblockNum to addEvent requests

texuf:


sergekh2 (approved):

sergekh2 (approved):

packages/sdk/src/client.ts
```diff
@@ -2466,7 +2480,27 @@ export class Client
...
+
+            if (errorContains(err, Err.MINIBLOCK_TOO_NEW)) {
+                this.logInfo('RETRYING event after MINIBLOCK_TOO_NEW response', {
+                    syncStats: this.streams.stats(),
+                    retryCount,
+                    prevMiniblockHash,
+                    prevMiniblockNum,
+                })
+                await new Promise((resolve) => setTimeout(resolve, 1000))
+                return await this.makeEventWithHashAndAddToStream(
```
sergekh2, line 2492: technically there is no need to make new event in this case, but this should work too

PR #2969, opened 2025-05-02
web3: simplify shims with typechain factories

salzbrenner:
Before, all shims had to provide 3 files: "contract" and "interface" from typechain, and the json ABI.

Typechain already packages all of this and makes it available via their factories

This PR updates the BaseContractShim to accept a typed contract using these factories. We don't need to import json ABIs, or other typechain typings, although a few shims still import some extra types.

There were some mismatches in several shim files between the typings get imported, and the abi that was used to create the contract interface. see [channel shim](https://github.com/towns-protocol/towns/pull/2969/files#diff-3c190a76fe6634fc5bdaf22d9abd983661ff25f3486f6335396590dd8f7cd336L2) for example. I corrected this for I believe all shims by adding the abi to the generated types, and using that for the factory.


texuf (approved):
Nice

shuhuiluo (approved):

packages/web3/src/BaseContractShim.ts
```diff
@@ -1,63 +1,65 @@
...
 
 const logger = dlogger('csb:BaseContractShim')
 
-// V2 smart contract shim
-// todo: replace BaseContractShim with this when refactoring is done
-export class BaseContractShim<
-    T_DEV_CONTRACT extends ethers.Contract,
-    T_DEV_INTERFACE extends ethers.utils.Interface,
-> {
+export class BaseContractShim<T_CONTRACT extends ethers.Contract> {
```
shuhuiluo, line 10:
```suggestion
export class BaseContractShim<
  F = unknown,
  T_CONTRACT extends ethers.Contract = F extends { connect: Connect<unknown> } 
    ? ContractType<typeof F.connect> 
    : ethers.Contract
> {
```
Consider this for simplicity. Use it as
```typescript
export class IDropFacetShim extends BaseContractShim<IDropFacet__factory>
```

salzbrenner: I'll try it out!

shuhuiluo (approved):

PR #2972, opened 2025-05-03
Switch yarn to use global cache and hardlinks

sergekh2:


texuf (approved):

.yarnrc.yml
```diff
@@ -1,6 +1,16 @@
...
+
+enableImmutableCache: false
+
+enableImmutableInstalls: true
+
+enableTelemetry: false
+
+nmMode: hardlinks-global
 
 nodeLinker: node-modules
```
miguel-nascimento, line 15: https://yarnpkg.com/configuration/yarnrc#nodeLinker
we could make it similar to pnpm too

sergekh2: Haven't checked if all our packages are ready for plug-n-play. There are some not fully specified/inconsistent deps through our configs. Ideally we should fix them and go to yarn4 or pnpm (which is stricter about deps, so we can't upgrade without fixing).

miguel-nascimento: I was talking about using `nodeLinker: pnpm`. It behaves similarly, docs say:
> If pnpm, a node-modules will be created using symlinks and hardlinks to a global content-addressable store.

Instead of copying from the global cache, it will use a symlink.

.yarnrc.yml
```diff
@@ -1,6 +1,16 @@
+checksumBehavior: throw
+
 compressionLevel: mixed
 
-enableGlobalCache: false
+enableGlobalCache: true
+
+enableImmutableCache: false
+
+enableImmutableInstalls: true
```
miguel-nascimento, line 9: can we mark this as false? without this, if you change versions on package.json and hit yarn to install it wont work.

.yarnrc.yml
miguel-nascimento: I tried `nodeLinker: pnpm`. 
It doesn't really that well currently, since our monorepo has a bunch of unlisted dependencies (mostly on `packages/generated`)
Let's skip it for now

miguel-nascimento (approved):

.yarnrc.yml
sergekh2: in theory then you should run with env var override to update yarn.lock, but either value is fine with me

miguel-nascimento: I like the default one (when you dont provide the value in .yarnrc.yml). It allows without env on local, but on CI it blocks.

PR #2703, opened 2025-03-27
feat(contract): introduce `SwapRouter` for token swaps with fees

shuhuiluo:
Adds `SwapRouter` contract to facilitate token swaps through whitelisted routers, with integrated fee collection for treasury and posters. Implements swap execution, validation, fee distribution, and balance handling alongside supporting storage and interface.

PR #2980, opened 2025-05-05
add swap router shims

erikolsson:
stacked on #2703


packages/web3/src/v3/Space.ts
```diff
@@ -96,6 +98,7 @@ export class Space {
         this.tipping = new ITippingShim(address, provider)
         this.review = new IReviewShim(address, provider)
         this.treasury = new ITreasuryShim(address, provider)
+        this.swapRouter = new ISwapRouterShim(address, provider)
```
salzbrenner, line 101: so every space gets a swap router facet?
just double checking b/c in Towns you were asking about a swapRouter address, which suggests to me it's not...or are there multiple?

PR #2987, opened 2025-05-06
refactor(contract): surface channel service errors

shuhuiluo:
Replaced `ChannelService` error imports with `IChannelBase` and updated generated ABIs and typings to reflect new error definitions.


texuf (approved):
Thank you

giuseppecrj:
Why was this merged? ChannelService is separate from ChannelBase since one is the facet and the other is the library used across facets? IChannelBase.ChannelService doesn't make sense.

shuhuiluo:
> Why was this merged? ChannelService is separate from ChannelBase since one is the facet and the other is the library used across facets? IChannelBase.ChannelService doesn't make sense.

This is to unblock pull-river on harmony. Errors declared in a library are not exported in ABI when using with `CustomRevert`. Also arguably unlike internal functions, errors and events belong to the ABI which justifies putting them in an interface.

PR #2991, opened 2025-05-06
Implement miniblock candidate production slowdown if no progress is made

sergekh2:


core/node/storage/pg_stream_store.go
```diff
@@ -1333,6 +1333,59 @@ func (s *PostgresStreamStore) readMiniblockCandidateTx(
...
+	if err := tx.QueryRow(
+		ctx,
+		s.sqlForStream(
+			"SELECT COUNT(*) FROM {{miniblock_candidates}} WHERE stream_id = $1 AND seq_num = $2",
+			streamId,
+		),
+		streamId,
+		miniblockNumber,
+	).Scan(&count); err != nil {
+		// if errors.Is(err, pgx.ErrNoRows) {
```
clemire, line 1381: This is a nit, but: if we get a database or pgx-related error on this query which is not pgx.ErrNoRows, we might mistakenly assume the candidate count here is 0 instead of realizing there was an error. Could you branch here on error type?

PR #2993, opened 2025-05-07
Space stream trimming

rhbntt:
Space streams (10xxx) can grow indefinitely due to encryption key solicitation requests that continue to loop. This leads to excessive database growth - for example, one stream was observed to take up 150GB. Since space streams don't require pagination support, we can safely trim old miniblocks.


core/node/storage/pg_stream_store.go
```diff
@@ -34,8 +34,8 @@ type PostgresStreamStore struct {
 
 	numPartitions int
 
-	//
 	esm *ephemeralStreamMonitor
+	st  *streamTrimmer
```
bas-vk, line 38: Please use more descriptive names for fields.

core/node/storage/pg_stream_trimmer.go
```diff
@@ -0,0 +1,258 @@
...
+	// Start the stream trimmer
+	st.start(ctx)
+
+	return st, nil
+}
+
+// start starts the stream trimmer
+func (t *streamTrimmer) start(ctx context.Context) {
+	t.wg.Add(1)
+	defer t.wg.Done()
```
bas-vk, line 55: Should this be in the go routine?

core/node/storage/pg_stream_trimmer.go
```diff
@@ -0,0 +1,258 @@
...
+		defer ticker.Stop()
+
+		for {
+			select {
+			case <-ctx.Done():
+				t.log.Info("Stream trimmer stopped due to context cancellation", "err", ctx.Err())
+				return
+			case <-t.stop:
+				t.log.Info("Stream trimmer stopped due to stop signal")
+				t.close()
```
bas-vk, line 69: Do we need this stop method with the wait group?

core/node/storage/pg_stream_trimmer.go
```diff
@@ -0,0 +1,258 @@
...
+
+	if len(streams) == 0 {
+		t.log.Debug("No space streams to trim")
+		return
+	}
+
+	t.log.Infow("Starting trim operation", "totalStreams", len(streams))
+
+	// Create a worker pool
+	streamCh := make(chan StreamId, len(streams))
```
bas-vk, line 113: We use `github.com/gammazero/workerpool` in other places.

core/node/storage/pg_stream_trimmer.go
```diff
@@ -0,0 +1,242 @@
...
+	for {
+		// Use CTID for efficient deletion
+		rows, err := tx.Query(
+			ctx,
+			t.store.sqlForStream(
+				`WITH rows_to_delete AS (
+					SELECT ctid
+					FROM {{miniblocks}}
+					WHERE seq_num < $1
+					LIMIT $2
```
bas-vk, line 180: This must include the stream id as criteria.

core/node/storage/pg_stream_trimmer.go
```diff
@@ -0,0 +1,242 @@
...
+	}
+
+	// Calculate the miniblock number to keep (last snapshot - miniblocksToKeep)
+	miniblockToKeep := lastSnapshotMiniblock - t.miniblocksToKeep
+	if miniblockToKeep <= 0 {
+		return nil // Nothing to trim
+	}
+
+	// Delete miniblocks in batches
+	const batchSize = 1000
```
bas-vk, line 170: I can't say how fast this deletion is. But for large streams this can probably take a long time while having a lock on the stream.

A different approach is to give the stream trimmer a worker pool with a reduced worker count to pace deletions (for instance 4). Drop the for loop here and if there are more miniblocks to delete for the stream post a new trim task to the worker pool. The worker pool will continue to work on other streams in the mean time and when its this task turn will delete the next batch of miniblocks. In the mean time the stream is unlocked and can progress.

And instead of scheduling every hour new trim tasks have some kind of accounting that keeps track for which streams there is a pending deletion task to prevent having multiple trim tasks scheduled for the same stream. And every time a new miniblock is written with a number > `t.miniblocksToKeep` schedule a new miniblock trim task for the stream.

This approach distributes the load more evenly while having short stream lock durations.

core/node/storage/pg_stream_trimmer.go
```diff
@@ -0,0 +1,237 @@
...
+		if err := t.processTrimTask(task); err != nil {
+			t.log.Errorw("Failed to process trim task",
+				"stream", task.streamId,
+				"err", err,
+			)
+		}
+
+		// Remove the pending task mark
+		t.pendingTasksLock.Lock()
+		delete(t.pendingTasks, streamId)
```
bas-vk, line 186: In the following scenario:

task is picked up by the worker pool (`scheduleTrimTask` already returned), `processTrimTask` deletes batch size miniblocks and schedules a new trim task. `t.pendingTasks[streamId]` is not yet deleted (that happens afterwards) so the task is dropped due to a duplicate? And when `processTrimTask` returns the stream is deleted from pendingTasks on this line?

bas-vk:
We also have stream reconcile logic. This syncs remote stream data into a nodes local state. This fetches all miniblocks from genesis. With stream trimming this logic also needs to be updated that it won't reconcile from genesis but only imports the last N miniblocks.

core/env/local/multi/on_chain.csv
```diff
@@ -16,3 +16,4 @@
 "stream.getMiniblocksMaxPageSize",0,200
 "stream.ephemeralStreamTTLMs",0,600000
 "stream.snapshotIntervalInMiniblocks",0,0
+"stream.spaceStreamMiniblocksToKeep",0,1000
```
sergekh2, line 19: lets do per-type settings in the same way as stream.minEventsPerSnapshot

core/node/storage/pg_stream_trimmer.go
```diff
@@ -0,0 +1,236 @@
...
+	}
+
+	// Calculate the miniblock number to keep
+	miniblockToKeep := lastSnapshotMiniblock - t.miniblocksToKeep
+	if miniblockToKeep <= 0 {
+		return nil // Nothing to trim
+	}
+
+	// Delete miniblocks in a single batch.
+	const batchSize = 1000
```
sergekh2, line 118: lets make it configurable

core/node/storage/pg_stream_trimmer.go
```diff
@@ -0,0 +1,236 @@
...
+	rows, err := tx.Query(
+		ctx,
+		t.store.sqlForStream(
+			`WITH rows_to_delete AS (
```
sergekh2, line 122: please change to num-based delete by the primary key

core/node/storage/pg_stream_trimmer.go
```diff
@@ -0,0 +1,267 @@
...
+}
+
+// scheduleStreamsTrimming schedules trimming for all trimmable streams.
+func (t *streamTrimmer) scheduleStreamsTrimming(ctx context.Context) error {
+	return t.store.txRunner(
+		ctx,
+		"streamTrimmer.scheduleStreamsTrimming",
+		pgx.ReadWrite,
+		func(ctx context.Context, tx pgx.Tx) error {
+			rows, err := tx.Query(ctx, "SELECT stream_id FROM es WHERE ephemeral = false")
```
sergekh2, line 243: Lets add SKIP LOCKED here so this tx doesn't conflict with other txes on es table

sergekh2 (approved):

core/node/storage/pg_stream_trimmer.go
```diff
@@ -0,0 +1,267 @@
...
+	// Calculate the highest miniblock number to keep
+	miniblockToKeep := lastSnapshotMiniblock - task.miniblocksToKeep
+	if miniblockToKeep <= 0 {
+		return nil // Nothing to trim
+	}
+
+	// Calculate the range to delete starting from miniblock 0
+	var startSeq int64
+	if task.lastDeletedMiniblock == 0 {
+		startSeq = miniblockToKeep - int64(t.trimmingBatchSize)
```
sergekh2, line 144: this part is confusing, if nothing was deleted before it seems that it will delete some mbs in the middle of the range?
Lets always delete from the beginning of the range so range in continuous. 
It seems the easier way is to select min number from db always and do calculation from there. Also eliminated the need to track lastDeletedMb potentially.

PR #3005, opened 2025-05-07
Add a filter option to snapshot repair in the migration tool

clemire:


core/tools/migrate_db/river_migrate_db.go
```diff
@@ -1728,31 +1740,46 @@ func restoreSnapshotIndices(
 		})
 	}
 
+	if verbose {
+		fmt.Println("Work submitted")
+	}
+
```
clemire, line 1746:
```suggestion
```

PR #3015, opened 2025-05-08
Fix order of operations issue in sdk when adding tips

texuf:
?? wasn't ??ing, adds were taking presidence

### DIFF
--- a/packages/sdk/src/streamStateView_User.ts
+++ b/packages/sdk/src/streamStateView_User.ts
@@ -108,8 +108,8 @@ export class StreamStateView_User extends StreamStateView_AbstractContent {
                             return
                         }
                         const currency = utils.getAddress(bin_toHexString(event.currency))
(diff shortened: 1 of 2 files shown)


packages/sdk/src/streamStateView_User.ts
```diff
@@ -108,8 +108,8 @@ export class StreamStateView_User extends StreamStateView_AbstractContent {
                             return
                         }
                         const currency = utils.getAddress(bin_toHexString(event.currency))
-                        this.tipsSent[currency] = this.tipsSent[currency] ?? 0n + event.amount
-                        this.tipsSentCount[currency] = this.tipsSentCount[currency] ?? 0n + 1n
```
miguel-nascimento, line 112: that wasn't causing a lint issue??

texuf: nupe.

PR #3021, opened 2025-05-09
Check client availability in remote syncer

rhbntt:
A remote node might be down so the remote syncer creation process is going to be timed out in this case. As a result, the entire modify sync request is going to fail. 

This PR adds a logic to make sure the remote node is available by sending the info request with a short timeout. Fixed a couple of bugs in the remote syncer, namely keep alive mechanism and node unavailability handling.

### DIFF
--- a/core/node/rpc/sync/client/syncer_set.go
+++ b/core/node/rpc/sync/client/syncer_set.go
@@ -4,11 +4,10 @@ import (
 	"context"
 	"slices"
 	"sync"
-	"sync/atomic"
 	"time"
 
 	"github.com/ethereum/go-ethereum/common"
-	"github.com/puzpuzpuz/xsync/v4"
+	"github.com/linkdata/deadlock"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/trace"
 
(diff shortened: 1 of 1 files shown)


texuf (commented):
does this actually change functionality or just refactor the error handling?

core/node/rpc/sync/client/remote.go
```diff
@@ -54,9 +54,25 @@ func newRemoteSyncer(
...
 
-	if !responseStream.Receive() {
+	// Create a timer for the first Receive
+	timer := time.NewTimer(5 * time.Second)
+	defer timer.Stop()
+
+	firstMsgChan := make(chan bool, 1)
+	go func() {
+		firstMsgChan <- responseStream.Receive()
+	}()
```
bas-vk, line 64: close firstMsgChan

core/node/rpc/sync/client/remote.go
```diff
@@ -54,9 +54,25 @@ func newRemoteSyncer(
 		return nil, err
 	}
 
-	if !responseStream.Receive() {
+	// Create a timer for the first Receive
+	timer := time.NewTimer(5 * time.Second)
```
bas-vk, line 58: Do you have some metrics that indicates that 5s is a reasonable time for this?

PR #3029, opened 2025-05-09
chore: update ethereum-cryptography to 3.2

miguel-nascimento:
This PR ugprade ehtereum-cryptography from 1.2 to 3.2.

Mostly, the changes needed is reported here: https://github.com/ethereum/js-ethereum-cryptography#from-v1-to-v2
Also, check https://github.com/paulmillr/noble-curves#upgrading

This is needed so we can continue moving towards the ESM future


packages/sdk/src/sign.ts
```diff
@@ -475,9 +475,12 @@ export async function riverSign(
...
-    // TODO(HNT-1380): why async sign doesn't work in node? Use async sign in the browser, sync sign in node?
-    const [sig, recovery] = signSync(hash, privateKey, { recovered: true, der: false })
-    return pushByteToUint8Array(sig, recovery)
+    const sig = secp256k1.sign(hash, privateKey)
+    const sigBytes = sig.toCompactRawBytes()
+    if (sig.recovery === undefined) {
+        throw new Error('Signature recovery bit is undefined')
+    }
+    return pushByteToUint8Array(sigBytes, sig.recovery)
 }
```
miguel-nascimento, line 484: There's no need for this function to be `async`. Is it fine if I remove and use it sync everywhere?

miguel-nascimento: also, now all secp256k1 fn are sync
https://github.com/paulmillr/noble-curves#upgrading

sergekh2: if there is no async version anymore we can do it.
it would be advantageous to have these async for app performance, but I guess it's not possible with this specific lib.

sergekh2 (approved):

packages/sdk/src/sign.ts
```diff
@@ -23,8 +23,8 @@ import {
     EventRefSchema,
     SnapshotSchema,
 } from '@towns-protocol/proto'
-import { assertBytes } from 'ethereum-cryptography/utils'
-import { recoverPublicKey, signSync, verify } from 'ethereum-cryptography/secp256k1'
+import { abytes } from '@noble/hashes/utils'
+import { secp256k1 } from 'ethereum-cryptography/secp256k1'
```
sergekh2, line 27: ethereum-cryptography/secp256k1 just forwards to @noble/curves/secp256k1, should we just use @noble/curves/secp256k1 directly? 

https://github.com/ethereum/js-ethereum-cryptography/blob/main/src/secp256k1.ts
```js
export { secp256k1 } from "@noble/curves/secp256k1";
```

miguel-nascimento: yup I think it’s better

packages/sdk/src/sign.ts
miguel-nascimento: even previously, it was fake async. I'll change the function signature to non-asnyc in a future PR.

PR #3046, opened 2025-05-12
feat: implement custom membership duration functionality

giuseppecrj:
### Description

This PR enhances the membership duration functionality by allowing spaces to set custom membership durations while respecting platform-defined maximum limits.

### Changes

- Added duration verification logic with `_verifyDuration()` to ensure durations don't exceed platform maximums
- Implemented custom duration storage in the membership storage layout
- Updated `_getMembershipDuration()` to first check for a custom duration before falling back to platform defaults
- Added `_setMembershipDuration()` method to update the duration with verification
- Included duration handling in the membership initialization process

### Checklist

- [x] Tests added where required
- [x] Documentation updated where applicable
- [x] Changes adhere to the repository's contribution guidelines

PR #3053, opened 2025-05-12
Check membership expiration during entitlement checks

salzbrenner:
### Description

Entitlement checks determine whether the user's balance for a membership NFT is greater than 0. However, when a token is expired, they will still have a balance (see MembershipRenew.t.sol). Entitlement checks need to change to make sure a wallet contains an unexpired token.

### Changes

- generates go bindings for MembershipFacet contract
- adds GetMembershipStatus to SpaceContractV3 which returns a MembershipStatus struct
- existing SpaceContract.IsMember updated to use GetMembershipStatus
- Updated checkEntitlement to use GetMembershipStatus. If a wallet has a membership nft that is not expired, or in the special case that they are joining a space (not a member yet), entitlement passes, otherwise it fails.

### Checklist

- [x] Tests added where required - added membershipRenewals.test in sdk package. looking at existing go tests, I didn't see any tests needed adding, as entitlement check tests use mocking, @clemire wdyt?

### DIFF
--- a/core/node/auth/auth_impl.go
+++ b/core/node/auth/auth_impl.go
@@ -1043,7 +1043,7 @@ func (ca *chainAuth) checkEntitlement(
 	}()
 
 	isMember := false
-	isExpired := false
+	isExpired := true
 	var membershipError error = nil
 
 	// This loop will wait on at least one true result, and will exit if the channel is closed,
@@ -1051,9 +1051,12 @@ func (ca *chainAuth) checkEntitlement(
 	for result := range isMemberResults {
 		if result.status.IsMember {
 			isMember = true
-			isExpired = result.status.IsExpired
-			isMemberCancel() // Cancel all other goroutines
-			break
+			// if not expired, cancel other checks, otherwise continue
+			if !result.status.IsExpired {
+				isExpired = false
+				isMemberCancel()
+				break
+			}
 		}
 	}
 
--- a/core/node/auth/auth_impl.go
+++ b/core/node/auth/auth_impl.go
@@ -426,13 +426,19 @@ func (ca *chainAuth) areLinkedWalletsEntitled(
 ) (bool, error) {
 	log := logging.FromCtx(ctx)
 	if args.kind == chainAuthKindSpace {
-		log.Debugw("isWalletEntitled", "kind", "space", "args", args)
+		log.Debugw("isWalletEntitled", "kind", "space", "args", args, "isExpired", *isExpired)
+		if *isExpired {
+			return false, nil
+		}
 		return ca.isEntitledToSpace(ctx, cfg, args)
 	} else if args.kind == chainAuthKindChannel {
-		log.Debugw("isWalletEntitled", "kind", "channel", "args", args)
+		log.Debugw("isWalletEntitled", "kind", "channel", "args", args, "isExpired", *isExpired)
+		if *isExpired {
+			return false, nil
+		}
 		return ca.isEntitledToChannel(ctx, cfg, args)
 	} else if args.kind == chainAuthKindIsSpaceMember {
-		log.Debugw("isWalletEntitled", "kind", "isSpaceMember", "args", args)
+		log.Debugw("isWalletEntitled", "kind", "isSpaceMember", "args", args, "isExpired", *isExpired)
 		return !*isExpired, nil // is space member is checked by the calling code in checkEntitlement
 	} else {
 		return false, RiverError(Err_INTERNAL, "Unknown chain auth kind").Func("isWalletEntitled")
--- a/core/node/auth/auth_impl.go
+++ b/core/node/auth/auth_impl.go
@@ -422,6 +422,7 @@ func (ca *chainAuth) areLinkedWalletsEntitled(
 	ctx context.Context,
 	cfg *config.Config,
 	args *ChainAuthArgs,
+	isExpired *bool,
 ) (bool, error) {
 	log := logging.FromCtx(ctx)
 	if args.kind == chainAuthKindSpace {
@@ -432,7 +433,7 @@ func (ca *chainAuth) areLinkedWalletsEntitled(
 		return ca.isEntitledToChannel(ctx, cfg, args)
 	} else if args.kind == chainAuthKindIsSpaceMember {
 		log.Debugw("isWalletEntitled", "kind", "isSpaceMember", "args", args)
-		return true, nil // is space member is checked by the calling code in checkEntitlement
+		return !*isExpired, nil // is space member is checked by the calling code in checkEntitlement
 	} else {
 		return false, RiverError(Err_INTERNAL, "Unknown chain auth kind").Func("isWalletEntitled")
 	}
@@ -892,11 +893,11 @@ func (ca *chainAuth) checkMembershipUncached(
 	_ *config.Config,
 	args *ChainAuthArgs,
 ) (CacheResult, error) {
-	isMember, err := ca.spaceContract.IsMember(ctx, args.spaceId, args.principal)
+	membershipStatus, err := ca.spaceContract.GetMembershipStatus(ctx, args.spaceId, args.principal)
 	if err != nil {
-		return boolCacheResult(false), err
+		return &membershipStatusCacheResult{status: nil}, err
 	}
-	return boolCacheResult(isMember), nil
+	return &membershipStatusCacheResult{status: membershipStatus}, nil
 }
 
 func (ca *chainAuth) checkMembership(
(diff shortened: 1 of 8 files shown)


giuseppecrj:
@salzbrenner contract work is merge to main in #3046

core/node/auth/auth_impl.go
```diff
@@ -1039,13 +1043,15 @@ func (ca *chainAuth) checkEntitlement(
 	}()
 
 	isMember := false
+	isExpired := false
 	var membershipError error = nil
 
 	// This loop will wait on at least one true result, and will exit if the channel is closed,
 	// meaning all checks have terminated, or if at least one check was positive.
 	for result := range isMemberResults {
```
clemire, line 1051: This point is kind of a nit in that I'm not sure how often we will see it right now, but I think there's technically a bug here. If a user has an expired membership token in one wallet and an unexpired token in another wallet, I believe there's a chance that this user would be evaluated here as unentitled depending on the order of what membership status is first returned from the contract queries for each linked wallet. Can you update this loop to only consider the user unentitled if all of their memberships are expired?

clemire: I'm not sure actually exactly what the contract does if a single wallet has > 1 token and only 1 of them is expired, I'm curious about that too tbh.

packages/sdk/src/tests/multi/entitlements/membershipRenewals.test.ts
```diff
@@ -0,0 +1,181 @@
...
+            spaceId,
+            alicesWallet.address,
+            getXchainConfigForTesting(),
+        )
+        expect(entitledWallet).toBeDefined()
+        // same here
+        const hasMembership = await aliceSpaceDapp.hasSpaceMembership(spaceId, entitledWallet!)
+        expect(hasMembership).toBe(true)
+
+        await expect(alice.joinStream(spaceId)).rejects.toThrow(/7:PERMISSION_DENIED/)
```
clemire, line 57: I think it's possible that this test could flake based on node membership result caching. We may end up needing to wait here for this test to consistently pass.

packages/sdk/src/tests/multi/entitlements/membershipRenewals.test.ts
```diff
@@ -0,0 +1,181 @@
...
+        await expectUserCanJoin(
+            spaceId,
+            channelId,
+            'bob',
+            bob,
+            bobSpaceDapp,
+            bobsWallet.address,
+            bobProvider.wallet,
+        )
+        // When bob's join event is added to the stream, it should trigger a scrub, and Alice
```
clemire, line 107: I am confused. Where is alice's wallet unlinked?

packages/sdk/src/tests/multi/entitlements/membershipRenewals.test.ts
```diff
@@ -0,0 +1,181 @@
...
+        // should be booted from the stream since she unlinked her entitled wallet.
```
clemire, line 168: comment here looks out of date.

packages/sdk/src/tests/multi/entitlements/membershipRenewals.test.ts
```diff
@@ -0,0 +1,181 @@
...
+        await expect(bob.joinStream(channelId)).resolves.not.toThrow()
+        const userStreamView = (await alice.waitForStream(makeUserStreamId(alice.userId))).view
+
+        // wait for a few seconds to make sure the scrub has been triggered
+        await new Promise((f) => setTimeout(f, 5_000))
```
clemire, line 173: Can we waitFor on the membership change instead to minimize the extra time it takes for this test to run?

core/node/auth/auth_impl.go
clemire: Regarding the specific scenario I mentioned above, i think it would be good to have this be codified and in e2e tests so that it will at least sporadically fail in CI if there is a true bug in our implementation.

clemire (changes requested):
You're right, I don't think we have any membership unit tests, so far we've done everything as e2e. I suggested an additional test scenario, lmk what you think.

Good job btw! I know the go code can be gnarly and also outside of your usual wheelhouse. Thanks for fixing this!

core/node/auth/auth_impl.go
salzbrenner: ok, changed as suggested, and added a test in which a user has 2 membership nfts, one expired and one not

packages/sdk/src/tests/multi/entitlements/membershipRenewals.test.ts
salzbrenner: copy pasted

packages/sdk/src/tests/multi/entitlements/membershipRenewals.test.ts
salzbrenner: there is no membership change in this test - it's testing a user is still joined. is there another way to wait for any caches/scrubs to have taken effect?

packages/sdk/src/tests/multi/entitlements/membershipRenewals.test.ts
salzbrenner: the membership duration is 10s, and we wait for it to expire earlier in the test, so that should invalidate the cache do you think I need to change the placement, to be after the leaveStream event?

side note, 10s for duration, I know it's kind of long for these tests to wait, but I think that number will reduce flakiness

clemire: Nah if you counted it, I'm good. I didn't read it too too carefully.

packages/sdk/src/tests/multi/entitlements/membershipRenewals.test.ts
clemire: Ah, I think the comment is confusing then.

```and Alice
        // should be booted from the stream since her membership has expired```

File: packages/sdk/src/tests/multi/entitlements/membershipRenewals.test.ts [hunk 6]
```diff
@@ -0,0 +1,200 @@
...
+            ).toBe(true)
+        })
+
+        // Clean up
+        await alice.stopSync()
+        await bob.stopSync()
+    })
+
+    test('user with unexpired membership is not bounced from a channel after scrub is triggered', async () => {
+        const {
```
clemire, line 126: Sorry I'm just reading this again and picking up a bit more now... Are you get all of these values (`alice`, `bob`, ...) as return values from `createTownWithRequirements`? I think these may be doing duplicate work.

packages/sdk/src/tests/multi/entitlements/membershipRenewals.test.ts
```diff
@@ -0,0 +1,200 @@
...
+
+        // link a wallet to alice
+        const eoaWallet = ethers.Wallet.createRandom()
+        await linkWallets(aliceSpaceDapp, aliceProvider.wallet, eoaWallet)
+
+        // have alice join the channel
+        await expectUserCanJoinChannel(alice, aliceSpaceDapp, spaceId, channelId)
+
+        // Wait for membership to expire (and for channel to be eligible for scrubbing)
+        await new Promise((f) => setTimeout(f, WAIT_TIME + 2_000))
```
clemire, line 162: Nit: for the sake of simplicity, since we've already added a constant buffer to WAIT_TIME, we might just want to either use WAIT_TIME alone here as it is, or use it alone and add a couple more seconds to it.

packages/sdk/src/tests/multi/entitlements/membershipRenewals.test.ts
```diff
@@ -0,0 +1,200 @@
...
+            channelId,
+            'alice',
+            alice,
+            aliceSpaceDapp,
+            alicesWallet.address,
+            aliceProvider.wallet,
+        )
+
+        await expectUserCanJoinChannel(alice, aliceSpaceDapp, spaceId, channelId)
+        // Wait 5 seconds so the channel stream will become eligible for scrubbing
```
clemire, line 95: I think this comment is inaccurate, can you update it?

packages/sdk/src/tests/multi/entitlements/membershipRenewals.test.ts
```diff
@@ -0,0 +1,200 @@
...
+            aliceSpaceDapp,
+            alicesWallet.address,
+            aliceProvider.wallet,
+        )
+
+        // wait for membership to expire
+        await new Promise((resolve) => setTimeout(resolve, WAIT_TIME))
+        await alice.leaveStream(spaceId)
+
+        // space dapp does not check expiration, so we'll still get an entitled wallet
```
clemire, line 48: 👍🏻 This is helpful, TY!

clemire (approved):

PR #3059, opened 2025-05-13
fix: channelScrubbing.test.ts

salzbrenner:
### Description

channelScrubbing.test wasn't testing loss entitlement anything because the final waitFor had no assertion

### Changes

- added assertion
- changed the joiner to be a member that was not already joined to the space, which kicked off the scrubber

### Checklist


clemire (approved):
Thank you! That one is a bit subtle.

PR #3078, opened 2025-05-15
chore: pin `account-abstraction` dev dependency

miguel-nascimento:
https://app.towns.com/messages/8843de0513cf436e2b9afc754e71f51dc80a35f141764d12d9535a77d9d62060#5b8fe75afd684abae3688acf53b3ef62b7bba901783a7988a2c92f13cf43b25e

### DIFF
--- a/packages/contracts/package.json
+++ b/packages/contracts/package.json
@@ -38,7 +38,7 @@
     "@prb/test": "^0.6.4",
     "@towns-protocol/prettier-config": "workspace:^",
     "@wagmi/cli": "^2.2.0",
-    "account-abstraction": "github:eth-infinitism/account-abstraction#cc3893bcaf2272c163ce89d5eb9eadb8e6b52db7",
+    "account-abstraction": "github:eth-infinitism/account-abstraction#v0.7.0",
     "forge-std": "github:foundry-rs/forge-std#v1.9.7",
     "prettier": "^3.5.3",
     "prettier-plugin-solidity": "^1.4.2",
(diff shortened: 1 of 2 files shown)


packages/contracts/package.json
```diff
@@ -38,7 +38,7 @@
     "@prb/test": "^0.6.4",
     "@towns-protocol/prettier-config": "workspace:^",
     "@wagmi/cli": "^2.2.0",
-    "account-abstraction": "github:eth-infinitism/account-abstraction",
+    "account-abstraction": "github:eth-infinitism/account-abstraction#cc3893bcaf2272c163ce89d5eb9eadb8e6b52db7",
```
shuhuiluo, line 41:
```suggestion
    "account-abstraction": "github:eth-infinitism/account-abstraction#v0.8.0",
```
A release tag is more readable.

shuhuiluo (changes requested):

packages/contracts/package.json
miguel-nascimento: nice call, I'm pinning to v0.7, since it was the version that we we're using. We can update in another PR

shuhuiluo (approved):

nachoiacovino:
why don't we just install the package instead?

shuhuiluo:
> why don't we just install the package instead?

The npm package is not published by the same org.

nachoiacovino:
but they suggest you install it in their readme, I don't know if pinning will fix it didn't seem to fix the issue for the crypto package

shuhuiluo:
> but they suggest you install it in their readme, I don't know if pinning will fix it didn't seem to fix the issue for the crypto package

They didn't put a npm badge in the README. I would rather trust the GitHub tag than the npm package name.

This is not related to crypto-lib. That one is weird.

PR #3080, opened 2025-05-15
Refactor isEntitled to return result instead of bool

texuf:
Add stub for Evan to add reason into the result
Reasoning: We want to differentiate between leave membership events which were initiated by the user, by the scrubber when entitlements were lost, and by the scrubber when membership has expired.

### DIFF
--- a/core/node/auth/auth_impl.go
+++ b/core/node/auth/auth_impl.go
@@ -46,8 +46,29 @@ type ChainAuth interface {
 		6B. For channels, the space contract method `IsEntitledToChannel` is called for each linked wallet. If any of the
 			linked wallets are entitled to the channel, the permission check passes. Otherwise, it fails.
 	*/
-	IsEntitled(ctx context.Context, cfg *config.Config, args *ChainAuthArgs) (bool, error)
+	IsEntitled(ctx context.Context, cfg *config.Config, args *ChainAuthArgs) (IsEntitledResult, error)
 	VerifyReceipt(ctx context.Context, cfg *config.Config, receipt *BlockchainTransactionReceipt) (bool, error)
+}
+
+type isEntitledResult struct {
+	isAllowed bool
+	isExpired bool
+}
+
+type IsEntitledResult interface {
+	IsEntitled() bool
+	Reason() string
+}
+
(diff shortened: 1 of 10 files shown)


core/node/auth/auth_impl_cache.go
```diff
@@ -25,11 +25,13 @@ type entitlementCache struct {
 
 type CacheResult interface {
 	IsAllowed() bool
+	IsExpired() bool
```
clemire, line 28: `IsAllowed` is actually used by the implementation of the entitlementCache, but `IsExpired` is not. Is it possible to move this lower to a single specific struct? Because ofc not all structs are actually using it to communicate anything meaningful, and it's not needed by the cache itself.

You can actually cast the specific returned interface value to a specific struct implementation instead to get access to this method. I think that might be better to control the maintainability of our caching implementation until we have a chance to refactor it.

core/node/rpc/create_stream.go
```diff
@@ -114,11 +114,11 @@ func (s *Service) createStream(ctx context.Context, req *CreateStreamRequest) (*
 
 	// check entitlements
 	if csRules.ChainAuth != nil {
-		isEntitled, err := s.chainAuth.IsEntitled(ctx, s.config, csRules.ChainAuth)
+		isEntitledResult, err := s.chainAuth.IsEntitled(ctx, s.config, csRules.ChainAuth)
 		if err != nil {
 			return nil, nil, err
 		}
-		if !isEntitled {
```
clemire, line 121: Same here, I'd like to include the reason in the returned error if possible.

clemire (changes requested):

core/node/auth/auth_impl.go
```diff
@@ -527,6 +555,10 @@ func (scr *entitlementCacheResult) IsAllowed() bool {
 	return scr.allowed
 }
 
+func (scr *entitlementCacheResult) Reason() EntitlementResultReason {
```
clemire, line 558: very big nit: since this is not "simpleCacheResult", can I trouble you to change these two method receiver names to "ecr"?

clemire (approved):
👍🏻

PR #3081, opened 2025-05-15
add MembershipReason to MemberPayload and UserPayload

salzbrenner:
### Description

Adds a MembershipReason to the membership/user payloads. For now  it only defines `MR_NONE` and `MR_NOT_ENTITLED`, to be updated with an expired reason later.

This allows for better tracking of why users leave or are removed.

For now, the only time a reason applies will be removal via scrubber.

### Checklist

- [x] Tests updated where required


core/node/rules/can_add_event.go
```diff
@@ -1485,6 +1485,22 @@ func (ru *aeUserMembershipRules) validUserMembershipTransition() (bool, error) {
...
+		// at this time, the only reasons are for the scrubber so we need to make sure the membership op is leave
+		if ru.userMembership.Op != MembershipOp_SO_LEAVE {
+			return false, RiverError(Err_PERMISSION_DENIED, "reasons should be undefined for non-leave events", "op", ru.userMembership.Op)
+		}
+		userAddress, err := shared.GetUserAddressFromStreamId(*ru.params.streamView.StreamId())
+		if err != nil {
+			return false, err
+		}
+		// reasons should only be set by the scrubber at this time
+		if bytes.Equal(ru.params.parsedEvent.Event.CreatorAddress, userAddress.Bytes()) {
```
clemire, line 1498: For this, I think a better check here would be, is the event creator an actual node? As there are many non-user addresses that are also invalid.

clemire: There is a `creatorIsValidNode()` existing pattern that you can use here. I think you will need to define it on `aeUserMembershipRules`, but there's a lot of examples in the file. Let me know if you have questions!

clemire (changes requested):
There is also a membership_scrub_test.go which has an ObservingEventAdder that is used to observe events the scrubber emits. I think it would ideally also be good if you could update these tests to validate that the observed events emitted from the scrubber are all membership leave events with the expected author/reason. We currently do not validate even the event type but just the count.

Vibecoding-wise, you may be actually be able to do this pretty quickly by adding in the context of the rpc folder tests to cursor, and asking the agent to include validation that incoming payloads are membership leaves with the scrubber reason, in `*ObservingEventAdder.AddEventPayload`.

clemire (approved):

PR #3082, opened 2025-05-15
Add membership renewal functionality and expiration checks

salzbrenner:
### Description

This PR introduces handling expired memberships in SpaceDapp.

### Changes

- Added membership renewal functionality via `renewMembership` method
- Replaced `hasSpaceMembership` with more detailed `getMembershipStatus` that includes expiration information
- Enhanced entitlement checks to properly handle expired memberships
- Optimized token ownership queries using multicall for better performance
- Fixed stress test code to use the new membership status API
- Added tests for membership renewal and expiration scenarios

### Checklist

### DIFF
--- a/packages/sdk/src/tests/multi/entitlements/membershipRenewals.test.ts
+++ b/packages/sdk/src/tests/multi/entitlements/membershipRenewals.test.ts
@@ -16,14 +16,14 @@ import { MembershipOp } from '@towns-protocol/proto'
 import { ethers } from 'ethers'
 
 const SHORT_MEMBERSHIP_DURATION = 20 // seconds
-const WAIT_TIME = SHORT_MEMBERSHIP_DURATION * 1_000 + 1_000
+const WAIT_TIME = SHORT_MEMBERSHIP_DURATION * 1_000 + 500
 
 // this test is long because it has to wait for the membership to expire in real time
 // but the membership duration has to be long enough such that other actions/assertions can be made
 // it tests both node entitlement checks for expired memberships, as well as client entitlement checks via space dapp
 
 describe('membershipRenewals', () => {
-    test('expired membership is not allowed to join', async () => {
+    test.concurrent('expired membership is not allowed to join', async () => {
         const { spaceId, channelId, alice, bob, aliceSpaceDapp, aliceProvider, alicesWallet } =
             await createTownWithRequirements({
                 everyone: true,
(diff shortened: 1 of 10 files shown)


packages/web3/src/space-dapp/SpaceDapp.ts
```diff
@@ -847,11 +847,26 @@ export class SpaceDapp {
...
+        if (permission === Permission.JoinSpace) {
+            // if you're joining a space with an expired membership, you're not entitled to join
+            if (isMember && isExpired) {
+                return
+            }
+        }
+        // otherwise you're trying to do something with an expired membership
+        else if (isExpired) {
+            return
+        }
```
clemire, line 868: Is there a logical difference between `isExpired` == true and `isMember && isExpired == true`? If isExpired is true, then isMember is always also true too, is that right?

If that's true, does it make sense to handle space joins separately? Or can we just say, if you're expired, go home?

packages/web3/src/space/Space.ts
```diff
@@ -480,12 +483,128 @@ export class Space {
...
+            expirations = await this.multicall.makeCalls({
+                encoder: () =>
+                    tokenIds.map((id) => this.membership.encodeFunctionData('expiresAt', [id])),
+                decoder: (result) => {
+                    return this.membership.decodeFunctionResult('expiresAt', result)[0].toBigInt()
+                },
+            })
+        } catch (error) {
+            log.error('getMembershipStatus::error', { error })
+            // Error evaluating expirations, assume not expired
```
clemire, line 529: This strategy is the opposite of how we consider entitlement errors usually - we don't grant permissions on error.

packages/sdk/src/tests/multi/entitlements/membershipRenewals.test.ts
```diff
@@ -42,155 +46,243 @@ describe('membershipRenewals', () => {
...
+                carolsWallet.address,
+                carolProvider.wallet,
+            )
+            // When carol's join event is added to the stream, it should trigger a scrub, and Alice
+            // should not be booted from the stream since her she has an additional token that is not expired
+            await expect(carol.joinStream(channelId)).resolves.not.toThrow()
+            const userStreamView = (await alice.waitForStream(makeUserStreamId(alice.userId))).view
+
+            // wait for any caches to be invalidated
+            await new Promise((f) => setTimeout(f, 5_000))
```
clemire, line 207: I don't think we're taking any actions after we do this wait that require the caches to be cleared; I think we can remove it.

packages/sdk/src/tests/multi/entitlements/membershipRenewals.test.ts
```diff
@@ -42,155 +46,243 @@ describe('membershipRenewals', () => {
...
+                channelId,
+                alice,
+                carol,
+                aliceSpaceDapp,
+                aliceProvider,
+                alicesWallet,
+                carolSpaceDapp,
+                carolProvider,
+                carolsWallet,
+            } = await createTownWithRequirements({
```
clemire, line 142: Your pushback from the previous PR on why you preferred not to use createTownWithRequirements for accounts made sense to me, btw. Totally fine if you have a reason like that to keep it as is, just FYI!

clemire (commented):
Questions, comments

packages/sdk/src/tests/multi/entitlements/membershipRenewals.test.ts
```diff
@@ -15,11 +15,15 @@ import { makeUserStreamId } from '../../../id'
 import { MembershipOp } from '@towns-protocol/proto'
 import { ethers } from 'ethers'
 
-const SHORT_MEMBERSHIP_DURATION = 10 // seconds
-const WAIT_TIME = SHORT_MEMBERSHIP_DURATION * 1_000 + 1_000
+const SHORT_MEMBERSHIP_DURATION = 20 // seconds
+const WAIT_TIME = SHORT_MEMBERSHIP_DURATION * 1_000 + 500
+
+// this test is long because it has to wait for the membership to expire in real time
```
clemire, line 21: +1 helpful

packages/web3/src/space/Space.ts
salzbrenner: Good call, I adjusted

packages/web3/src/space-dapp/SpaceDapp.ts
salzbrenner: Great point!!

PR #3085, opened 2025-05-15
Simplify tiny part of isExpired check in go entitlements

texuf:


salzbrenner (approved):

core/node/auth/auth_impl.go
```diff
@@ -1122,44 +1115,53 @@ func (ca *chainAuth) checkEntitlement(
 			"rootKey", args.principal, "wallets", len(wallets)).LogError(log)
 	}
 
-	result, err := ca.areLinkedWalletsEntitled(ctx, cfg, args, &isExpired)
+	if isExpired {
```
salzbrenner, line 1118: isExpired inits as true right?
what happens with `args.kind == chainAuthKindIsSpaceMember`. with this, a new member joining a space will be denied entry?

texuf: 1. yes, that's how it reads, but..
2. how would the tests pass?

clemire (approved):

core/node/auth/auth_impl.go
texuf: this is fine, the function 1st checks to see if you're not a member, if you are a member, checks to see if it's expired, if it's not expired, makes sure you're still entitled

texuf: I rebased to remove formatting issues

PR #3087, opened 2025-05-15
Move false entitlement result on >> linked wallets higher in the method to avoid unnecessary work

clemire:
### Checklist


core/node/auth/auth_impl.go
```diff
@@ -1020,6 +1020,13 @@ func (ca *chainAuth) checkEntitlement(
 		return boolCacheResult(false), nil
 	}
 
+	// Now that we know the user is a member of the space, we can check entitlements.
```
clemire, line 1023:
```suggestion
```

texuf (approved):

texuf (commented):
Seems like this check should be in the contract and we shouldn’t let people brick their accounts by linking one too many wallets

clemire:
> Seems like this check should be in the contract and we shouldn’t let people brick their accounts by linking one too many wallets

True dat, let me check.

clemire:
> > Seems like this check should be in the contract and we shouldn’t let people brick their accounts by linking one too many wallets
> 
> True dat, let me check.

@texuf looks like we do check it, see https://github.com/towns-protocol/towns/blob/9d69c64791481bf67065a72fb73703f0f1376e06/packages/contracts/src/factory/facets/wallet-link/WalletLinkBase.sol#L435. the MAX_LINKED_WALLETS constant appears to be checked everywhere a wallet is added.

PR #3098, opened 2025-05-16
Enable stream trimming for user setting stream

rhbntt:


core/env/local/multi/on_chain.csv
```diff
@@ -18,3 +18,4 @@
 "stream.snapshotIntervalInMiniblocks",0,0
 "stream.defaultStreamTrimmingMiniblocksToKeep",0,0
 "stream.streamTrimmingMiniblocksToKeep.10",0,100
+"stream.streamTrimmingMiniblocksToKeep.a5",0,100
```
texuf, line 21: please make this 1000k blocks so that we have some buffer here. 100 is nowhere near enough. 
The space should be >200, 100 might just be one snapshot.

sergekh2 (approved):

PR #3115, opened 2025-05-20
feat(dApp) add space delegation

ttillberg:
Enables space delegation and provides access to node operator getters, both required for delegating staking

- Expose BaseRegistry
- Add space delegation helper

### DIFF
--- a/packages/web3/src/space-dapp/SpaceDapp.ts
+++ b/packages/web3/src/space-dapp/SpaceDapp.ts
@@ -54,7 +54,7 @@ import { UserEntitlementShim } from '../space/entitlements/UserEntitlementShim'
 import { RiverAirdropDapp } from '../airdrop/RiverAirdropDapp'
 import { BaseChainConfig } from '../utils/IStaticContractsInfo'
 import { WalletLink, INVALID_ADDRESS } from '../wallet-link/WalletLink'
-import { UNKNOWN_ERROR } from '../BaseContractShim'
+import { OverrideExecution, UNKNOWN_ERROR } from '../BaseContractShim'
 import { PricingModules } from '../pricing-modules/PricingModules'
 import { dlogger, isTestEnv } from '@towns-protocol/dlog'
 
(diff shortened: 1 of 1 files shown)


packages/web3/src/space-dapp/SpaceDapp.ts
```diff
@@ -1898,4 +1901,35 @@ export class SpaceDapp {
...
+        const spaceAddress = space.Address
+
+        return wrapTransaction(
+            () =>
+                this.baseRegistry.spaceDelegation
+                    .write(signer)
+                    .addSpaceDelegation(spaceAddress, operatorAddress),
+            txnOpts,
+        )
+    }
```
salzbrenner, line 1934:
```suggestion
       public async addSpaceDelegation<T = ContractTransaction>(args: {
        spaceId: `0x${string}`
        operatorAddress: `0x${string}`
        signer: ethers.Signer
        txnOpts?: TransactionOpts
        overrideExecution?: OverrideExecution<T>
    }) {
        const { spaceId, operatorAddress, signer, txnOpts, overrideExecution } = args
        const space = this.getSpace(spaceId)
        if (!space) {
            throw new Error(`Space with spaceId "${spaceId}" is not found.`)
        }
        const spaceAddress = space.Address

        return this.baseRegistry.spaceDelegation.executeCall({
            signer,
            functionName: 'addSpaceDelegation',
            args: [spaceAddress, operatorAddress],
            overrideExecution,
            transactionOpts: txnOpts,
        })
    }
```

This is a new pattern I'm experimenting with, try it out? To allow for the userops class to pass in it's own execution function, so we don't have to duplicate the body of the write method from the space dapp to the userop class
For this method not really a big deal, but other methods that might calculate data internally for the write, or busting caches etc, this should make it easier to avoid discrepancies between the 2 classes

packages/web3/src/space-dapp/SpaceDapp.ts
```diff
@@ -1898,4 +1901,34 @@ export class SpaceDapp {
...
+     * @param args.spaceId - The space id
+     * @param args.operatorAddress - The operator address
+     * @returns The transaction
+     */
+    public async addSpaceDelegation<T = ContractTransaction>(args: {
+        spaceId: string
+        operatorAddress: string
+        signer: ethers.Signer
+        transactionOpts?: TransactionOpts
+        overrideExecution?: (calldata: string) => Promise<T>
```
salzbrenner, line 1917: overrideExecution?: (args: { calldata: string; value?: bigint }) => Promise<T>
or
overrideExecution?: OverrideExecution<T>

should fix build

PR #3116, opened 2025-05-20
hold setUsername until session is available

erikolsson:
Let's avoid injecting more new sessions than necessary into the streams. 100 members == large space.


packages/sdk/src/client.ts
```diff
@@ -1153,11 +1155,28 @@ export class Client
...
+            this.pendingUsernames.set(streamId, username)
+            const prevTimeout = this.pendingUsernameTimeouts.get(streamId)
+            if (prevTimeout) clearTimeout(prevTimeout)
+            const timeout = setTimeout(() => {
+                void this.setUsername(streamId, username, true)
+            }, 60000)
+            this.pendingUsernameTimeouts.set(streamId, timeout)
+            return
+        }
+        this.pendingUsernameTimeouts.delete(streamId)
```
texuf, line 1178: you need a `this.pendingUsernames.delete(streamId)` here

PR #3120, opened 2025-05-20
chore: remove rust toolchain from ci

miguel-nascimento:
lets see if it passes :)

my guess: leftover from mls-rs encryption


shuhuiluo (approved):
Lgtm.

PR #3145, opened 2025-05-22
Modified handling of isNewDevice

texuf:
This should be able to go in without affecting needing iOS updates yet

I think it was a mistake to not pass session ids in the initial key solicitations, we end up in a situation where we can ping pong solicitations without actually satisfying the request. The contents of this change

- isNewDevice becomes a suggestion instead of a state
- all key requests now require session ids
- if isNewDevice is true, clients will still send all known sessions, but explicitly state in the fulfillment if they fulfilled any of the requested ids.

this should tighten the reqest loop and prevent at least one back and forth between clients in some cases.

update test

### DIFF
--- a/packages/encryption/src/decryptionExtensions.ts
+++ b/packages/encryption/src/decryptionExtensions.ts
@@ -339,7 +339,7 @@ export abstract class BaseDecryptionExtensions {
                 if (keySolicitation.deviceKey === this.userDevice.deviceKey) {
                     continue
                 }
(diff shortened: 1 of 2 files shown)


texuf:
> Hm! If we deploy this, won't new iOS devices be unable to solicit keys from web? iOS will send isNewDevice and '[]' session ids right now if I'm not mistaken

that's fine! then it will just work the same way as before.

PR #3154, opened 2025-05-23
auto retry blockchain txs with 0 confirmations

erikolsson:
When the node's view of the blockchain world is lagging behind, the client is unable to post blockchain related events.
This PR makes sure that we retry those AddEvent calls if needed. 
Also does some minor refactoring to make the catch clauses a bit less repetitive.

image

### DIFF
--- a/packages/sdk/src/client.ts
+++ b/packages/sdk/src/client.ts
@@ -2515,7 +2515,8 @@ export class Client
                 // for blockchain transactions: if we get a permission denied error, and the error message contains 'Transaction has 0 confirmations',
                 // we need to retry the event, since the node is likely lagging behind, and waiting a few seconds will allow it to catch up
                 errorContains(err, Err.PERMISSION_DENIED) &&
-                errorContainsMessage(err, 'Transaction has 0 confirmations.') &&
+                (errorContainsMessage(err, 'Transaction has 0 confirmations.') ||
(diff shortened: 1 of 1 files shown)


packages/sdk/src/client.ts
```diff
@@ -2503,17 +2507,19 @@ export class Client
...
-                    isDefined(localId) ? eventId : undefined,
-                    cleartext,
-                    tags,
-                    retryCount + 1,
                 )
+            } else if (
+                // for blockchain transactions: if we get a permission denied error, and the error message contains 'Transaction has 0 confirmations',
+                // we need to retry the event, since the node is likely lagging behind, and waiting a few seconds will allow it to catch up
+                errorContains(err, Err.PERMISSION_DENIED) &&
+                errorContainsMessage(err, 'Transaction has 0 confirmations') &&
```
bas-vk, line 2518: The message ends with a `.` in the node. But this will work.

The node end can also be improved and the client. It currently retrieves the tx receipt, if fails -> error (not converted to RiverErr) and the client won't retry. If the receipt isn't found the error should be mapped to a `Err.NOT_FOUND` and also retried because the client knows that the receipt is available but the client is lagging behind. I can make this change in the node and push it on this branch if you want.

texuf (approved):

PR #3162, opened 2025-05-25
Shared syncer set backfilling mechanism

rhbntt:
This PR adds a way to catch up a new client on a stream that’s already being synced by others. When a user starts syncing and adds a stream with a custom cookie, but that stream is already live in the shared syncer set, the node will first send all the missed changes since that cookie. Once it has sent those past updates, it will continue sending new updates as they happen. To do this, the node makes a backfill request to the stream syncer, which then delivers the stream delta to the shared channel before forwarding it on to the client.

### DIFF
--- a/core/node/rpc/sync/operation.go
+++ b/core/node/rpc/sync/operation.go
@@ -446,7 +446,7 @@ func (syncOp *StreamSyncOperation) process(cmd *subCommand) error {
 			return RiverError(Err_CANCELED, "sync operation cancelled").
 				Tags("syncId", syncOp.SyncID)
 		}
-	case <-time.After(30 * time.Second):
+	case <-time.After(30 * time.Second): // TODO: make this configurable
 		err := RiverError(Err_DEADLINE_EXCEEDED, "sync operation command queue full").
 			Tags("syncId", syncOp.SyncID)
 		syncOp.log.Errorw("Sync operation command queue full", "error", err)
--- a/core/node/rpc/sync/subscription/manager.go
+++ b/core/node/rpc/sync/subscription/manager.go
@@ -40,7 +40,7 @@ type Manager struct {
 	// subscriptions is a map of stream IDs to subscriptions.
 	subscriptions map[StreamId][]*Subscription
 	// stopped is a flag that indicates whether the manager is stopped (1) or not (0).
-	stopped uint32
+	stopped atomic.Bool
 	// otelTracer is the OpenTelemetry tracer used for tracing individual sync operations.
 	otelTracer trace.Tracer
 }
@@ -78,7 +78,7 @@ func NewManager(
 
 // Subscribe creates a new subscription with the given sync ID.
 func (m *Manager) Subscribe(ctx context.Context, cancel context.CancelCauseFunc, syncID string) (*Subscription, error) {
-	if atomic.LoadUint32(&m.stopped) == 1 {
+	if m.stopped.Load() {
 		return nil, RiverError(Err_UNAVAILABLE, "subscription manager is stopped").Tag("syncId", syncID)
 	}
 
@@ -96,7 +96,7 @@ func (m *Manager) Subscribe(ctx context.Context, cancel context.CancelCauseFunc,
 // start starts the subscription manager and listens for messages from the syncer set.
 func (m *Manager) start() {
 	defer func() {
-		atomic.StoreUint32(&m.stopped, 1)
+		m.stopped.Store(true)
 		m.cancelAllSubscriptions(m.globalCtx.Err())
 	}()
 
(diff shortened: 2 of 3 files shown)


core/node/rpc/sync/subscription/manager.go
```diff
@@ -165,7 +171,17 @@ func (m *Manager) distributeMessage(msg *SyncStreamsResponse) {
...
 				m.sLock.Unlock()
-			} else {
+			} else if msg.GetTargetSyncId() == "" || msg.GetTargetSyncId() == subscription.syncID {
+				if initializing, _ := subscription.initializingStreams.Load(streamID); initializing {
+					// The given stream is in initialization state for the given subscription.
+					if msg.GetSyncOp() == SyncOp_SYNC_UPDATE && msg.GetTargetSyncId() == subscription.syncID {
+						// Backfill of the stream targeted specifically to the given subscription
+						subscription.initializingStreams.Delete(streamID)
+					} else {
+						return
```
texuf, line 181: Will this early return skip the wg.done() call below?

core/node/rpc/sync/subscription/subscription.go
```diff
@@ -50,7 +54,11 @@ func (s *Subscription) Send(msg *SyncStreamsResponse) {
 		return
 	}
 
-	err := s.Messages.AddMessage(proto.Clone(msg).(*SyncStreamsResponse))
+	msg = proto.Clone(msg).(*SyncStreamsResponse)
+	if msg.GetTargetSyncId() != "" {
+		msg.TargetSyncId = ""
+	}
```
texuf, line 60: Why do we need to clear the targetSyncId

protocol/protocol.proto
```diff
@@ -994,12 +996,20 @@ message RemoveStreamFromSyncResponse {}
 
 // ModifySyncRequest adds or removes streams from an existing sync session.
 message ModifySyncRequest {
+  message Backfill {
+    // backfill_streams is the list of sync cookies to backfill
+    repeated SyncCookie streams = 1;
+    string sync_id = 2;
+  }
```
texuf, line 1003: How is this different than the add streams list and why would it need a unique syncid?

core/node/events/stream.go
```diff
@@ -776,6 +776,59 @@ func (s *Stream) addEventLocked(
 	return newSV, nil
 }
 
+// Backfill retrieves the stream content since the given cookie and sends it to the callback.
+func (s *Stream) Backfill(
```
bas-vk, line 780: Instead of calling this `Backfill` can it be renamed to something like `UpdatesSinceCookie`? Backfill makes sense in the context of subscriptions but this is on the Stream type and I think renaming it would be more clear for users that don't see it in the context of a stream update subscription.

core/node/events/stream.go
```diff
@@ -776,6 +776,59 @@ func (s *Stream) addEventLocked(
...
+	ctx context.Context,
+	cookie *SyncCookie,
+	cb func(*StreamAndCookie),
+) error {
+	if !s.IsLocal() {
+		return RiverError(
+			Err_NOT_FOUND,
+			"stream not found",
```
bas-vk, line 788: Rename this to `stream not local`.

core/node/events/stream.go
```diff
@@ -776,6 +776,59 @@ func (s *Stream) addEventLocked(
...
+			Err_BAD_SYNC_COOKIE,
+			"bad stream id",
+			"cookie.StreamId",
+			cookie.StreamId,
+			"s.streamId",
+			s.streamId,
+		)
+	}
+
+	view, err := s.lockMuAndLoadView(ctx)
```
bas-vk, line 816: There is `Stream#GetViewIfLocal` that returns the view only when it's local.

core/node/rpc/sync/subscription/subscription.go
```diff
@@ -27,19 +31,23 @@ type Subscription struct {
 	cancel context.CancelCauseFunc
 	// syncID is the subscription/sync ID
 	syncID string
-	// Messages is the channel for the subscription messages
-	Messages *dynmsgbuf.DynamicBuffer[*SyncStreamsResponse]
 	// closed is the indicator of the subscription status. 1 means the subscription is closed.
-	closed  uint32
+	closed uint32
```
sergekh2, line 35: same

core/node/rpc/sync/operation.go
```diff
@@ -437,7 +446,7 @@ func (syncOp *StreamSyncOperation) process(cmd *subCommand) error {
 			return RiverError(Err_CANCELED, "sync operation cancelled").
 				Tags("syncId", syncOp.SyncID)
 		}
-	case <-time.After(10 * time.Second):
+	case <-time.After(30 * time.Second):
```
sergekh2, line 449: I wonder eventually if we might need setting for this.

sergekh2 (approved):

PR #3163, opened 2025-05-25
[wip] run packages without esbuild

miguel-nascimento:
This PR changes the build pipeline to push cjs and esm, correctly changes tsconfig to use `moduleResolution: bundler` to lib packages and `moduleResolution: NodeNext` for apps 

Possible follow up for this PR: moving apps like stream-metadata, xchain-monitor, and other "runnable" packages to an apps folder, like [this example from Turborepo](https://github.com/vercel/turborepo/tree/main/examples/kitchen-sink)
- it makes a clear distinction of what is supposed to run / whats supposed to be bundled as a lib


packages/sdk/src/rpcCommon.ts
miguel-nascimento, line 1: ill move this into a different change - sorry

PR #3169, opened 2025-05-26
chore: move runnable packages to `apps` folder

miguel-nascimento:
Still confusing to me to be honest. We have servers, docs, monitoring, stress scripts.. 😵‍💫

Will leave it in draft for a while, need to think if its better to have everything at packages or break a bit..


texuf:
i think we want to keep packages for now, but i see how this makes sense.

PR #3171, opened 2025-05-27
node: implement new strategy for choosing stream nodes

bas-vk:
Currently the method to pick a set of nodes for a new stream is randomised. The result is that nodes that just joined the platform get as many new streams assigned to them as nodes that already have a lot of streams assigned. This results in an unbalanced situation that doesn't resolve it self.

This PR implements a new approach for picking nodes to assign a stream to. It picks N candidates pseudo-randomly from the operational nodes and selects from these candidates replication factor best nodes. With best defined as nodes that has the least number of streams assigned to them. It tries to select nodes from a unique set of operators for correct distribution.

It introduces a new on-chain config setting (default value=1) that is used to determine the number of candidates to select the repl factor best nodes from (replication factor+extra_candidates).

### DIFF
--- a/core/node/stream/distribution.go
+++ b/core/node/stream/distribution.go
@@ -136,7 +136,24 @@ func NewDistributor(
 		}
 	})
 
-	// update internal statistic for stream node calculation
+	// update node registry state for stream node selection
+	nodeRegistryChainMonitor := crypto.NewNodeRegistryChainMonitor(chainMonitor, riverRegistry.Address)
+	nodeRegistryChainMonitor.OnNodeAdded(blockNumber+1,
+		func(context.Context, *river.NodeRegistryV1NodeAdded) {
+			d.Reload()
+		})
+	nodeRegistryChainMonitor.OnNodeRemoved(blockNumber+1,
+		func(context.Context, *river.NodeRegistryV1NodeRemoved) {
+			d.Reload()
+		})
+	nodeRegistryChainMonitor.OnNodeStatusUpdated(
+		blockNumber+1,
+		func(context.Context, *river.NodeRegistryV1NodeStatusUpdated) {
+			d.Reload()
+		},
+	)
+
+	// update internal statistic for stream node selection
 	chainMonitor.OnContractWithTopicsEvent(blockNumber+1, riverRegistry.Address,
 		[][]common.Hash{{riverRegistry.StreamRegistryAbi.Events["StreamUpdated"].ID}}, d.onStreamUpdate)
 


core/node/stream/distribution.go
```diff
@@ -0,0 +1,399 @@
...
+	"github.com/towns-protocol/towns/core/node/crypto"
+	"github.com/towns-protocol/towns/core/node/logging"
+	. "github.com/towns-protocol/towns/core/node/protocol"
+	"github.com/towns-protocol/towns/core/node/registries"
+	. "github.com/towns-protocol/towns/core/node/shared"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/linkdata/deadlock"
+	"github.com/twmb/murmur3"
```
sergekh2, line 22: github.com/cespare/xxhash/v2 is already used in other parts of the code, maybe lets use it here as well to standardize on a single hash?

core/node/stream/distribution.go
```diff
@@ -0,0 +1,399 @@
...
+	streamID StreamId,
+	replFactor int,
+) ([]common.Address, error) {
+	impl := d.impl.Load()
+	if len(impl.nodes) < replFactor {
+		return nil, ErrInsufficientNodesAvailable
+	}
+
+	cfg := d.cfg.Get()
+	candidateCount := int(cfg.StreamDistribution.CandidatesCount)
```
sergekh2, line 154: lets refactor this setting to be "extra candidate count".

it's very possible that soon we'll have streams with different replication factors: for example, metadata streams probably going to use replication factor 7 or such.

with extra candidate count then "actual candidate count" can be computed as `replFactor` + "extra candidate count"

sergekh2 (approved):

core/node/stream/distribution.go
```diff
@@ -0,0 +1,496 @@
...
+					if node, found := impl.nodes[nodeAddr]; found {
+						node.streamCount.Add(1)
+					}
+				}
+			}
+		}
+	}
+}
+
+func (d *streamsDistributor) Reload() {
```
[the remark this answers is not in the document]
bas-vk: Because I want to ensure that reloading only happens when needed. For instance if 1 river block contains multiple modifications to the registry I want to prevent loading the nodes multiple times. And loading nodes takes some time so it is possible that if consecutive river blocks contain changes I don't want to miss them.

core/node/stream/distribution.go
bas-vk: Missed that xxhash was already used. It provides a good distribution so I switched to that.

PR #3198, opened 2025-05-30
adding protobuf dependencies to the vscode ~Start Local Dev deps

mechanical-turk:
### Description

Add "Install Protobuf Dependencies" to the `~ Start Local Dev` task.
Also, pulling other dependency subtasks out of `Stage 1`, into a separate sub-task called "Dependencies" - patten-matching harmony.

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -9,6 +9,21 @@
   },
   "tasks": [
     {
+      "label": "Dependencies",
+      "dependsOn": [
+        "CheckVersions",
+        "FoundryUp",
+        "YarnInstall",
+        "Install Protobuf Dependencies",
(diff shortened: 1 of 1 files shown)


.vscode/tasks.json
```diff
@@ -8,6 +8,21 @@
...
+        "YarnInstall",
+        "Install Protobuf Dependencies",
+        "BuildProtobufs",
+      ],
+      "group": {
+        "kind": "build",
+        "isDefault": true
+      },
+      "dependsOrder": "sequence"
+    },
```
texuf, line 25: is this slow if it's already installed? if so can you just add it to checkVersions? otherwise i'm fine with this.

PR #3213, opened 2025-06-02
ci: bump golangci-lint-action to v8

eeemmmmmm:
Maintenance update: switch all jobs to golangci-lint-action@v8 for better performance and compatibility. Pure maintenance, behavior unchanged. More details in the [v8.0.0 release](https://github.com/golangci/golangci-lint-action/releases/tag/v8.0.0).


sergekh2:
we upgrade entire go toolchain at once

PR #3223, opened 2025-06-03
autogenerated docs for npm packages

miguel-nascimento:
Use the commits to review the changes.

Recently found typedoc, used by [TanStack](https://github.com/TanStack/config/blob/ed9cc9fca96220f66d7cf774d5204eaaa48e451d/packages/typedoc-config/src/index.js) and some Vercel libraries.

It's not perfect, and I 100% prefer a more human approach to this, but it gets the job done for a quick check and move us towards `llm.txt` if we can narrow down the output.

Typedoc generates markdown from JSDocs, and it automatically look into types to attach them in the markdown (parameters and return types)

Coderabbit has an option that autogenerates JSDocs for PRs. We can use it to build auto-healing docs.

Q: do we want to publish those auto-generated docs into [docs.towns.com](https://docs.towns.com)? I personally think we need to throw JSDocs everywhere first. Maybe a good task for Devin.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -8510,8 +8510,6 @@ __metadata:
     solady: ^0.1.14
     solhint: ^5.0.5
     turbo: ^2.0.12
-    typedoc-plugin-frontmatter: ^1.3.0
-    typedoc-plugin-markdown: ^4.6.4
   languageName: unknown
   linkType: soft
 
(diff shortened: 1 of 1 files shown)


miguel-nascimento:
A very interesting thing is that the docs show how much bloat we're exporting it 🙃

miguel-nascimento:
not yet, will hook it into docs.towns.com and add it in the npm publish script

PR #3226, opened 2025-06-03
yarn build added to vscode ~Start Local Dev steps

mechanical-turk:
### Description

The `~Start Local Dev` vscode script has implicit assumptions about what's currently present in the local file system. For example, the playground assumes that the react-sdk is built. But the `~Start Local Dev` script does not actually build this package, resulting in a broken JS experience with errors. This PR adds a `yarn build` step to the workflow, which makes sure these implicit dependencies are explicitly built.

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -14,6 +14,7 @@
         "CheckVersions",
         "FoundryUp",
         "YarnInstall",
+        "YarnBuild",
         "BuildProtobufs",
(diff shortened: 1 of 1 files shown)


texuf (commented):
I guess i never saw an issue because the watches should build everything that is needed... 
I'm fine with this going in though, but if you add it you should be able to remove the build protobufs step here since that is included there

PR #3227, opened 2025-06-03
Add rejected event hashes to miniblock header

sergekh2:
### Description

For metadata streams there is a need to have a record of events that were rejected from inclusion into miniblock (i.e. it's analogous to failed tx record in blockchain block)


texuf (approved):
nice

PR #3228, opened 2025-06-03
[wip] sdk-connector package to solve node/web connectrpc imports

miguel-nascimento:
WIP

The idea is to use in `packages/sdk` as a common.

So you can use `createClient` and if its node (solved by package.json exports field) its gonna import from `packages/sdk-connector/src/node/index.ts`, while keeping the function sync (no need for async + dynamic import call)

If its node, we want to set httpVersion to 2.

Currently blocked - need to ship tsconfig moduleResolution changes first to support bundler in libraries and nodenext in apps.


packages/sdk-connector/package.json
```diff
@@ -0,0 +1,61 @@
+{
+    "name": "@towns-protocol/sdk-connector",
```
miguel-nascimento, line 2: Ill probably change the name to rpc-connector instead
`@towns-protocol/rpc-connector` reads way better

PR #3235, opened 2025-06-04
move node:crypto usage in sdk to a isolated package

miguel-nascimento:
This allow us to bundle correctly sdk to both browser and web.
Also allow us to remove the `vite-plugin-node-polyfills` from clients.

Stacked on top of 
- #3228 
- #3163 

Requires vitest update to 3.2.1 #3237 
Requires @ethereumjs/util update to v10 (latest, has browser support) #3236


texuf (commented):
can you add a unit test that it equals either a Buffer or a Uint8Array in each environment?

texuf:
we're going to think about creating another version of the tests, one for node, one for web. 
is it fine if the sdk uses web and the stress and stream-metadata use node?

PR #3240, opened 2025-06-04
Quck cleanup in react-sdk

texuf:
makes it easier to refactor the observables in upcoming timeline change


texuf:
oops, too soon

PR #3264, opened 2025-06-06
fix ci while arc-runner bug is fixed upstream

jterzis:
arc-runner pods keep crashing in GKE with an upstream error fetching from GH backend. This is blocking CI progress. Let's use gh hosted vm's until this issue is fixed. It is likely an issue with GH backend or gh-runner image as nothing else has changed with our deployment which runs a Helm chart for the runner system in GKE.
Screenshot 2025-06-06 at 10 13 06 AM

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -211,7 +211,7 @@ jobs:
                   access_token: ${{ github.token }}
 
             - name: Install Binaries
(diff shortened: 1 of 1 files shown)


clemire:
I wonder if this will coincidentally fix the super super flaky CI tests on the main branch...

jterzis:
As of this comment, the bug with our gke hosted gh runners is still happening (can track https://github.com/actions/runner/issues/3899#issuecomment-2950892061) so will merge this to unblock CI.

PR #3266, opened 2025-06-06
Revert "Introduce a delay when closing a sync session runner"

clemire:
Reverts towns-protocol/towns#3179

Looks like I introduced a typo here and missed it, and it caused a bug.


clemire:
Failed test is already flaky on main - I'll go ahead and merge this.

PR #3308, opened 2025-06-10
Code for creating metadata streams

sergekh2:


core/node/shared/ids.go
```diff
@@ -291,9 +291,6 @@ func MetadataStreamIdFromShard(shard uint64) StreamId {
 }
 
 func ShardForStreamId(streamId StreamId, shardMask uint64) uint64 {
```
bas-vk, line 293: Should this function check that the given stream id isn't a metadata stream id?

sergekh2: seems a bit of overkill to check it here

PR #3358, opened 2025-06-17
chore: dont clean `.turbo` in `yarn-clean.sh`

miguel-nascimento:
Let's say that you moved into a branch that contains changes to the `sdk` package and you run the `kill all clean and start local dev` vscode task. This task will call `yarn-clean.sh`, but If we clean the `.turbo` folder, you will need to rebuild **all packages**. Including contracts, which takes some time.

If we keep `.turbo`, we can reuse build cache of other packages and turbo will only re-build the `sdk` package.


miguel-nascimento:
You're right, but we have it in `kill all clean and start local dev` vscode task, which at least I use when I switch branches

texuf:
> You're right, but we have it in `kill all clean and start local dev` vscode task, which at least I use when I switch branches

Just use the restart local dev instead? Or make a new task. Please close this.

PR #3405, opened 2025-06-21
Move the decryption extensions back to the sdk package

texuf:
I hope this doesn’t break anyone, we only moved it here because the encryption package was the first we released and we wanted it to contain our encryption/decryption protocol - but having it away from the rest of the sdk makes things unnecessarily hard and also it’s worthless in isolation without the greater context of the streams, syncing, and persistence. It’s also a huge bottleneck in the app right now and we need to figure out something else.
I did not re-combine with the “ClientDecryptionExtensions” class - i can do that later in a separate commit


texuf:
> > It’s also a huge bottleneck in the app right now and we need to figure out something else.
> 
> Unrelated with this PR, but would love to hear why!

on my local client, which i haven't reset in a very long time, i see 
```
csb:decryption:A3c7..942c:20QA..oQgc status: working queues: priorityTasks: 0, newGroupSession: 0, ownKeySolicitations: 16 encryptedContent: 11377, streamsMissingKeys: 0, keySolicitations: 42059  +2m
```
that's 11k items enqueued to decrypt, and 42k key solicitations that i haven't replied to yet! It clicks through them one by one, but they are all in memory in the app! we need to figure out a faster way to get through these and we need to get them out of memory and into local storage.

PR #3421, opened 2025-06-24
Protocol and SDK updates to support bots, take 2

clemire:
Squash of the following:

Refactor to move bot entitlement check before entitlement data is fetched for space or channel.

Checkpoint - need to refactor to avoid fetching entitlements for bots.

Add typescript types for account interface contract change.

Remove console logs.

Partial repair of tests -> use bot client address as app id with app registry service.

yarn lint fixes.

Fix syncpack and syntax issues.

All e2e tests pass.

Tests almost done.

clean up some yarn lint errors.

Yarn prettier.

Checkpoint. fixed errors returned from space contract so that failed event adds due to contract errors are not propogated. An installed bot can join a space.

Fix missing creator id.

Fix stream creation entitlement checks for bot users.

WIP 4.

WIP.

WIP.

WIP

### Checklist


core/node/auth/auth_impl.go
```diff
@@ -1056,60 +1161,142 @@ func (ca *chainAuth) checkStreamIsEnabled(
...
+				"principal", args.principal,
+				"foundAppAddress", appAddress,
+				"isBot", isBot,
+			)
+
+		var reason EntitlementResultReason
+		if args.kind == chainAuthKindIsBot {
+			isEntitled := isBot && appAddress == args.appAddress
+			if !isEntitled {
+				reason = EntitlementResultReason_IS_NOT_BOT
```
texuf, line 1200: should we have a third reason for app address mismatch? that seems like it would be a logic error somewhere else, maybe we should at least log it.

texuf: also i think these would be more clear
```
if !isEntitled {
    return boolCacheResult(false, EntitlementResultReason_IS_NOT_BOT)
} else {
    return boolCacheResult(true)
}
```
for both

core/node/auth/auth_impl.go
```diff
@@ -1210,48 +1405,73 @@ func (ca *chainAuth) checkEntitlement(
...
+		return boolCacheResult{false, reason}, nil
 	}
 
-	result, cacheHit, err := ca.membershipCache.executeUsingCache(
-		ctx,
-		cfg,
-		&args,
-		ca.checkMembershipUncached,
-	)
+	cacheResult, err := ca.checkIsBot(ctx, args)
```
texuf, line 1435: I would call this variable isBotResult instead of cacheResult

core/node/rules/can_add_event.go
```diff
@@ -1657,13 +1734,19 @@ func (ru *aeMembershipRules) channelMembershipEntitlements() (*auth.ChainAuthArg
 		return nil, err
 	}
 
+	var appAddress common.Address
+	if permissionUser == common.Address(ru.membership.UserAddress) {
+		appAddress = common.BytesToAddress(ru.membership.AppAddress)
+	}
```
texuf, line 1740: why is this if statement here? if there's an app address on the membership we should always check it right?

packages/sdk/src/client.ts
```diff
@@ -565,84 +586,90 @@ export class Client
...
         userStreamId: string | Uint8Array,
-        metadata: { spaceId: Uint8Array } | undefined,
+        metadata?: { spaceId?: Uint8Array; appAddress?: Uint8Array },
     ): Promise<ParsedStreamResponse> {
         const userEvents = [
             await makeEvent(
                 this.signerContext,
                 make_UserPayload_Inception({
                     streamId: streamIdAsBytes(userStreamId),
+                    appAddress: metadata?.appAddress ?? bin_fromHexString(''),
```
texuf, line 596: appAddress is optional in the protocol, you shouldn't need `?? bin_fromHexString`

packages/sdk/src/utils.ts
```diff
@@ -18,6 +25,42 @@ export function safeSet<O extends Record<any, any>, K extends keyof O>(
...
+
+export const ethereumAddressToBytes = (address: string): Uint8Array => bin_fromHexString(address)
+
+export const ethereumAddressFromBytes = (bytes: Uint8Array): string => bin_toHexString(bytes)
+
+export const ethereumAddressAsString = (address: string | Uint8Array): string =>
+    typeof address === 'string' ? address : ethereumAddressFromBytes(address)
+
+export const ethereumAddressAsBytes = (address: string | Uint8Array): Uint8Array =>
+    typeof address === 'string' ? ethereumAddressToBytes(address) : address
```
texuf, line 47: we have these address conversions already in id.ts

packages/sdk/src/client.ts
```diff
@@ -565,84 +586,90 @@ export class Client
...
-            metadata: metadata,
+            metadata: encoded,
         })
         return unpackStream(response.stream, this.opts?.unpackEnvelopeOpts)
     }
 
     private async createUserSettingsStream(
         inUserSettingsStreamId: string | Uint8Array,
-        metadata: { spaceId: Uint8Array } | undefined,
+        metadata?: { spaceId?: Uint8Array; appAddress?: Uint8Array },
```
texuf, line 655: these signatures should be `createUserXStream(streamId, appId, metadata)`
the metadata is passed along with the request, the app id is part of the inception, lets not mix them

core/node/rules/can_add_event.go
```diff
@@ -1481,6 +1488,63 @@ func (ru *aeMembershipRules) requireStreamParentMembership() (*DerivedEvent, err
 	}, nil
 }
 
+// ownerChainAuthForInviter validates that the inviter on the UserMembership event has space ownership.
+// For bots, we expect the user membership event to be derived from a user membership action posted
+// by the space owner; this authorization is required to ensure that bots are added to spaces or channels
+// directly by space owners.
+func (ru *aeUserMembershipRules) ownerChainAuthForInviter() (*auth.ChainAuthArgs, error) {
```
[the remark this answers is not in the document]
clemire: but what about channels? don't we want to restrict channel installation to owner-only?

core/node/rules/can_add_event.go
clemire: no, sometimes the permissions check is done against the inviter and not the app itself. In this case we would not want to create a chain auth with the app address because the app address in the chain auth is considered to be linked to the principal wallet.

clemire: I added a comment!

core/node/auth/auth_impl.go
clemire: Adding in a 3rd reason here makes this piece of code a little bit longer and more complex, and I'd really prefer to keep it contained with the actual registry check if possible, so we don't litter the calling function with specific nuances of bot auth.

does this sound reasonable? I can still refactor if you prefer.

core/node/rules/can_add_event.go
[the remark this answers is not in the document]
texuf: that is a good point. Okay good addition!

core/node/rules/can_add_event.go
texuf: okay please update `getPermissionForMembershipOp` to return an app address as well

packages/sdk/src/client.ts
```diff
@@ -1979,14 +2023,15 @@ export class Client
...
-        await this.initStream(streamId)
+        const stream = await this.initStream(streamId)
         check(isDefined(this.userStreamId))
         return this.makeEventAndAddToStream(
             this.userStreamId,
             make_UserPayload_UserMembershipAction({
                 op: MembershipOp.SO_JOIN,
                 userId: addressFromUserId(userId),
                 streamId: streamIdAsBytes(streamId),
+                streamParentId: stream.view.getContent().getStreamParentIdAsBytes(),
```
texuf, line 2034: did you start using the streamParentId anywhere?

packages/sdk/src/client.ts
```diff
@@ -565,84 +599,94 @@ export class Client
...
             await makeEvent(
                 this.signerContext,
                 make_UserSettingsPayload_Inception({
                     streamId: userSettingsStreamId,
+                    appAddress,
                 }),
             ),
         ]
 
+        const encoded = metadata ? stripUndefinedMetadata(metadata) : metadata
```
texuf, line 685: do we still need this encoding change? probably not a bad thing, just asking.

packages/sdk/src/client.ts
clemire: Since we now manually have space owners manually join bot users to spaces, we had to update this event to satisfy the JOIN constraints for derived membership events.

packages/sdk/src/client.ts
clemire: nope. I'll take it out.

PR #3442, opened 2025-06-25
Validate stream ID in the sync request 

rhbntt:
The node panics when an invalid stream ID is provided in the sync stream request. The PR updates the validation logic to make sure that the provided stream ID is valid.

### DIFF
--- a/core/node/rpc/sync/subscription/subscription.go
+++ b/core/node/rpc/sync/subscription/subscription.go
@@ -175,11 +175,9 @@ func (s *Subscription) addStream(cookie *SyncCookie) (shouldAdd bool, shouldBack
 		if !slices.ContainsFunc(subscriptions, func(sub *Subscription) bool {
 			return sub.syncID == s.syncID
 		}) {
-			if cookie.GetMinipoolGen() > 0 || len(cookie.GetPrevMiniblockHash()) > 0 {
-				// The given stream should be backfilled and then start syncing.
-				shouldBackfill = true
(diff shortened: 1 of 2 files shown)


core/node/rpc/sync/client/syncer_set.go
```diff
@@ -670,46 +672,115 @@ func (ss *SyncerSet) rmStream(streamID StreamId) {
 }
 
 // Validate checks the modify request for errors and returns an error if any are found.
-func (mr *ModifyRequest) Validate() error {
+func (mr *ModifyRequest) Validate() (bool, error) {
```
bas-vk, line 675: Add a description that when the returned bool is false the client receives the error through a callback so the caller must stop execution and return nil. And when the returned error isn't nil that caller is responsible to report the error to the client.

core/node/rpc/sync/client/syncer_set.go
```diff
@@ -670,46 +672,118 @@ func (ss *SyncerSet) rmStream(streamID StreamId) {
...
+				Message:  "Duplicate stream in removing operation",
+			})
+			invalid = true
+			continue
+		}
+		seen[streamId] = struct{}{}
+	}
+
+	// Prevent passing the same stream to both add and remove operations
+	if !invalid {
```
bas-vk, line 767: I'm not sure if this is correct? If `invalid=true` the agreement is that the caller stops execution and returns nil. In this case it can miss reports of streams that are both in the add and remove request?

bas-vk (approved):

PR #3490, opened 2025-06-28
Add event filtering support to GetMiniblocks

sergekh2:
Adds exclusion filter by both payload and content names.


core/node/rpc/get_miniblocks.go
```diff
@@ -64,3 +69,116 @@ func (s *Service) localGetMiniblocks(
 
 	return connect.NewResponse(resp), nil
 }
+
+// applyExclusionFilter applies exclusion filters to a miniblock, returning a new MiniblockInfo
+// with filtered events and partial flag set if any events were excluded
+func (s *Service) applyExclusionFilter(info *MiniblockInfo, exclusionFilter []*EventFilter) (*MiniblockInfo, error) {
```
bas-vk, line 75: There is no way this function could fail so there is no need to return an error.

bas-vk (approved):

PR #3513, opened 2025-07-01
Ephemeral key solicitations / fulfillments

erikolsson:
stacked on #3509


packages/sdk/src/tests/multi_ne/ephemeralEvents.test.ts
```diff
@@ -0,0 +1,58 @@
...
+
+        const stream = await alice.waitForStream(streamId)
+        expect(stream.view.getMembers().joinedUsers).toEqual(new Set([alice.userId, bob.userId]))
+
+        await waitFor(() => {
+            return stream.view.timeline.length == 6
+        })
+        const events = stream.view.timeline
+        console.log(events)
+    })
```
texuf, line 57: Not sure this really tests anything. Solicitations don't end up in the timeline any more.

PR #3514, opened 2025-07-01
sdk: decouple indexeddb from crypto store

miguel-nascimento:
This allow us to remove `fake-indexeddb` and use a in memory solution for apps that are not running in the browser (bots, stream-metadata, stress tests)

Would love to make the client more modular, so the user could override this "service" (i.e adding real database to store those keys if needed)

As a side effect, this PR enables bot developers to run the examples using `bun`.


miguel-nascimento:
@coderabbitai review

texuf (approved):

PR #3553, opened 2025-07-07
Filter out membership, key solicitation and fulfillment events in scrollback 

rhbntt:


packages/sdk/src/makeStreamRpcClient.ts
```diff
@@ -133,13 +135,15 @@ async function fetchMiniblocksFromRpc(
     fromInclusive: bigint,
     toExclusive: bigint,
     omitSnapshots: boolean,
+    exclusionFilter: { payload: string, content: string }[] | undefined,
```
texuf, line 138: payload can be typed as `PayloadCaseType`

texuf: for content you can add 
```
export type ContentCaseType = Extract<StreamEvent['payload']['value'], { content: any }>['content']['case']
```
to the types.ts file in the proto package

texuf (approved):

PR #3556, opened 2025-07-07
client support for bot metadata

miguel-nascimento:
This PR adds support for bot metadata in the `playground` and fixes the bot uber test

TODO: we should add some hooks to the `react-sdk` and the sync agent. Added a ticket and will work on those later.

image

image

image

image


clemire (commented):
Are there any issues with the bot avatar if the url is ipfs?

miguel-nascimento:
Yeah, we can change this if we have any issues in the playground. We’re not adding this validation in the sdk level

PR #3562, opened 2025-07-08
Fixed DOWNSTREAM_NETWORK_ERROR error when adding event

rhbntt:
### Problem

When a client requests to add an event to a stream:

1. The request is sent to node **A**.
2. Node **A** determines the request should be forwarded to the stream node **B** and adds a header to disable further forwarding.
3. Node **B** fails to add the event locally (e.g., unknown stream) and attempts to forward the request to another stream node **C** ([code](https://github.com/towns-protocol/towns/blob/main/core/node/rpc/forwarder.go#L559-L562)).
4. Node **C** receives the request, but forwarding is disabled by the header, so the check fails ([code](https://github.com/towns-protocol/towns/blob/main/core/node/rpc/forwarder.go#L568)).
5. The client receives a downstream network error.

### Root Cause

The forwarding logic does not account for the case where a node that receives a forwarded request (with forwarding disabled) cannot process it locally and tries to forward it again, resulting in a dead-end and a network error for the client.

### Solution

This PR addresses the issue by adding a retry mechanism when adding events. It uses an already used utility function with the retry logic. 

### Testing 

Performed 50+ runs of unit tests that produced the given issue with 100% success.

### DIFF
--- a/core/node/app_registry/service.go
+++ b/core/node/app_registry/service.go
@@ -453,7 +453,7 @@ waitLoop:
 			streamResponse, err := utils.PeerNodeRequestWithRetries(
 				ctx,
 				nodes,
-				func(ctx context.Context, stub protocolconnect.StreamServiceClient) (*connect.Response[GetStreamResponse], error) {
+				func(ctx context.Context, stub protocolconnect.StreamServiceClient, _ common.Address) (*connect.Response[GetStreamResponse], error) {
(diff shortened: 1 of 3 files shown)


texuf:
Can you describe what the fix is and how you tested it in your pr description?

clemire (approved):

PR #3575, opened 2025-07-10
Optimize shared syncer and add traces

rhbntt:
- Added traces
- Optimization in shared syncer

### DIFF
--- a/core/node/rpc/sync/subscription/distributor.go
+++ b/core/node/rpc/sync/subscription/distributor.go
@@ -67,17 +67,9 @@ func (d *distributor) DistributeBackfillMessage(streamID StreamId, msg *SyncStre
 	}
 
 	targetSyncID := msg.GetTargetSyncIds()[0]
-
-	// Check if the subscription is still associated with this stream
-	var subscription *Subscription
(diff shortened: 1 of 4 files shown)


core/node/rpc/sync/client/syncer_set.go
```diff
@@ -571,28 +580,31 @@ func (ss *SyncerSet) DebugDropStream(ctx context.Context, streamID StreamId) err
...
 	stream, err := ss.streamCache.GetStreamNoWait(ctx, streamID)
 	if err != nil {
 		return common.Address{}, false
 	}
 
-	// 2. Try local node if stream is local
 	remotes, isLocal := stream.GetRemotesAndIsLocal()
+
+	// 1. Try node from cookie
+	if addrRaw := cookie.GetNodeAddress(); len(addrRaw) > 0 {
```
sergekh2, line 591: If this code is moved here then optimization from not calling GetStreamNoWait is lost. We might drop node address from cookie then.

(It's interesting question if this stream record is going to in the cache anyway or not, i.e. if it's called here, what is percentage of loads vs cache hits).

PR #3589, opened 2025-07-11
prevent sending empty fulfillments

erikolsson:
I'm seeing empty fulfillments in the wild — not sure what's going on here


texuf (changes requested):
Because i monkeyed with the queues i'm pretty sure we're now sending fulfillments for streams that aren't up to date, which are not matching anything.

texuf (changes requested):
please fix it in the calling code

packages/sdk/src/clientDecryptionExtensions.ts
```diff
@@ -350,6 +350,10 @@ export class ClientDecryptionExtensions extends BaseDecryptionExtensions {
         ephemeral = false,
     }: KeyFulfilmentData & { ephemeral?: boolean }): Promise<{ error?: unknown }> {
         const userAddress = userIdToAddress(userId)
+        if (sessionIds.length === 0) {
+            this.log.debug('tried to send empty key fulfillment')
+            return {}
+        }
```
texuf, line 356: this was a valid case, updated code here: https://github.com/towns-protocol/towns/pull/3773

PR #3610, opened 2025-07-15
feat: add GetMiniblockNumberRanges for backwards reconciliation gap detection

sergekh2:
## Summary

- Implements `GetMiniblockNumberRanges` method in the storage layer to identify gaps in miniblock sequences
- Uses an optimized SQL window function query for efficient gap detection
- Returns structured `MiniblockRange` type with `StartInclusive` and `EndInclusive` fields

## Context

This PR implements the storage layer functionality needed for backwards reconciliation as described in the [backwards reconciliation feature documentation](https://github.com/towns-protocol/towns/blob/main/core/docs/features/backwards-reconciliation.md). Specifically, it addresses the TODO item:

> Extend storage layer to report gaps in the stream miniblocks, so gaps can be filled

### Storage Interface

Added `GetMiniblockNumberRanges` method to the `StreamStorage` interface that:
- Takes a stream ID and starting miniblock number
- Returns all continuous ranges of miniblock numbers present in storage
- Each range is represented as a `MiniblockRange` struct with explicit start/end fields

### Query Optimization

The implementation uses an efficient SQL window function approach:
```sql
SELECT 
    MIN(seq_num) AS start_range,
    MAX(seq_num) AS end_range
FROM (
    SELECT 
        seq_num,
        seq_num - ROW_NUMBER() OVER (ORDER BY seq_num) AS grp
    FROM miniblocks
    WHERE stream_id = $1 AND seq_num >= $2
) AS subquery
GROUP BY grp
ORDER BY start_range
```

This approach:
- Identifies gaps by comparing sequence numbers with row positions
- Groups continuous sequences efficiently
- Leverages PostgreSQL's window functions for optimal performance
- Uses `pgx.CollectRows` for efficient result scanning

### Performance

Performance test shows excellent results: ~2ms for 5000 miniblocks across 5 ranges with gaps.

## Test Plan

- [x] Unit tests covering various scenarios:
  - Empty/non-existent streams
  - Continuous sequences without gaps
  - Sequences with multiple gaps
  - Non-zero starting sequences
  - Large gaps between miniblocks
- [x] Integration tests with `WritePrecedingMiniblocks`
- [x] Performance test with 5000 miniblocks

## Next Steps

This foundation enables the implementation of backwards reconciliation logic in the events layer, where the gap information can be used to determine which miniblocks need to be fetched from remote replicas.


core/node/storage/pg_stream_store.go
```diff
@@ -2475,6 +2475,74 @@ func (s *PostgresStreamStore) getLowestStreamMiniblockTx(
...
+func (s *PostgresStreamStore) GetMiniblockNumberRanges(
+	ctx context.Context,
+	streamId StreamId,
+	startMiniblockNumberInclusive int64,
+) ([]MiniblockRange, error) {
+	var ranges []MiniblockRange
+	err := s.txRunner(
+		ctx,
+		"GetMiniblockNumberRanges",
+		pgx.ReadWrite,
```
bas-vk, line 2489: This can be reduced to read only.

sergekh2: unfortunately no: pg considers taking a lock "writing" even for "FOR SHARE" read locks.

PR #3615, opened 2025-07-15
Notifications: do not track streams on init

jontra:
### Description

Track only user settings streams on init. start tracking DM, GDM, and CHANNEL streams only after a miniblock creation.
this way, cold streams (which are not active), will not be tracked.
This feature is protected with config

### Changes

Added notifications.coldStreamsEnabled config
Added isInit parameter to TrackStream callback in stream tracker
Pass the correct miniblock hash to start the sync 

### Checklist

### DIFF
--- a/core/node/crypto/config.go
+++ b/core/node/crypto/config.go
@@ -58,7 +58,6 @@ const (
 	StreamUserSettingStreamTrimmingMiniblocksToKeepConfigKey = "stream.streamTrimmingMiniblocksToKeep.a5"
 	StreamEnableNewSnapshotFormatConfigKey                   = "stream.enableNewSnapshotFormat"
 	ServerEnableNode2NodeAuthConfigKey                       = "server.enablenode2nodeauth"
-
 	// StreamBackwardsReconciliationThresholdConfigKey is the threshold in miniblocks that determines
 	// whether to use backwards or forward reconciliation. If a stream is behind by more than this
 	// number of miniblocks, backwards reconciliation is used; otherwise forward reconciliation is used.
--- a/core/node/track_streams/multi_sync_runner.go
+++ b/core/node/track_streams/multi_sync_runner.go
@@ -124,7 +124,7 @@ func (ssr *syncSessionRunner) AddStream(
 	ssr.mu.Unlock()
 
 	logging.FromCtx(ctx).
-		Infow("Adding stream with cookie",
+		Debugw("Adding stream with cookie",
 			"stream", record.streamId,
 			"minipoolGen", record.minipoolGen,
 			"prevMiniblockHash", record.prevMiniblockHash,
@@ -266,7 +266,6 @@ func (ssr *syncSessionRunner) applyUpdateToStream(
 		}
 	}
 
-	// Reset historical content processing after the first update
 	record.applyHistoricalContent.Enabled = false
 
 	// Update record cookie for restarting sync from the correct position after relocation
--- a/core/node/track_streams/streams_tracker.go
+++ b/core/node/track_streams/streams_tracker.go
@@ -258,7 +258,6 @@ func (tracker *StreamsTrackerImpl) OnStreamLastMiniblockUpdated(
 	if !tracker.filter.TrackStream(event.GetStreamId(), false) {
 		return
 	}
-	logging.FromCtx(ctx).Infow("Stream last miniblock updated", "event", event)
 	if err := tracker.AddStream(event.GetStreamId(), ApplyHistoricalContent{true, event.LastMiniblockHash[:]}); err != nil {
 		logging.FromCtx(ctx).Errorw("Failed to add stream on miniblock update",
 			"streamId", event.GetStreamId(),


core/node/crypto/config.go
```diff
@@ -155,6 +157,10 @@ type OnChainSettings struct {
 	// Options: 1 means enabled, 0 means disabled.
 	ServerEnableNode2NodeAuth uint64 `mapstructure:"server.enablenode2nodeauth"`
 
+	// NotificationsColdStreamsEnabled indicates whether cold streams are enabled for notifications.
```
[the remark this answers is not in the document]
clemire: Also, do we want this to be on the chain config? I'm not sure why we would need to specify this in the protocol - It feels more like it should be configuration on the node.

core/node/track_streams/multi_sync_runner.go
```diff
@@ -42,6 +43,8 @@ type streamSyncInitRecord struct {
 	// Set to false if we wish to send only events starting from the current sync position. This should
 	// be true, as we will restart the sync with the latest sync position.
 	applyHistoricalStreamContents bool
+	// Set applyHistoricalStreamContentFromMiniblockHash
```
clemire, line 46: Can you add some text that explains what this setting does? Similar to the comment for the above field.

core/node/track_streams/streams_tracker.go
```diff
@@ -27,7 +27,11 @@ import (
 // must provide the logic for determining which streams to track, and for constructing application-specific
 // tracked stream views.
 type StreamFilter interface {
-	TrackStream(streamID shared.StreamId) bool
+	// TrackStream
+	// inInit is true when the stream_tracker is initialized (usually when the process starts), otherwise it's false
+	// updateType is nil when initializing; otherwise is set the appropriate type
```
clemire, line 32: updateType is not a parameter here

core/node/track_streams/multi_sync_runner.go
clemire: Also, let's unify this with the above field and make it a subtype that describes if historical content is desired and, if so, whether or not a miniblock hash is specified.

core/node/track_streams/multi_sync_runner.go
```diff
@@ -203,10 +207,13 @@ func (ssr *syncSessionRunner) applyUpdateToStream(
...
 			}
 		}
+		// apply blocks from the first block that matches applyHistoricalStreamContentFromMiniblockHash
+		// and all consecutive blocks
+		applyBlocks = applyBlocks || bytes.Equal(block.Header.GetHash(), record.applyHistoricalStreamContentFromMiniblockHash)
 		// If the stream was just allocated, process the miniblocks and events for notifications.
 		// If not, ignore them because there were already notifications sent for the stream, and possibly
 		// for these miniblocks and events.
-		if record.applyHistoricalStreamContents {
+		if record.applyHistoricalStreamContents || applyBlocks {
```
clemire, line 216: Let's move the reference to `record.applyHistoricalStreamContents` up to line 200:

`applyBlocks := record.applyHistoricalStreamContents`

then here you can just say, `if applyBlocks {`

core/node/track_streams/streams_tracker.go
```diff
@@ -44,7 +48,14 @@ type StreamsTracker interface {
 
 	// A stream that does not meet criteria for tracking at the time it is created can later be added via
 	// AddStream. An error will be returned if the stream could not be successfully added to the sync runner.
-	AddStream(streamId shared.StreamId) error
+	// set applyHistoricalStreamContents to true if you want to apply the stream from inception/last snapshot.
+	// alternatively, you can set applyHistoricalStreamContentFromMiniblockHash to the exact miniblock hash
+	// that should be applied, and all of the consecutive miniblocks
```
clemire, line 53: Yes, I think these two should be one object - they are coupled it it's easier to reason about.

core/node/track_streams/streams_tracker.go
```diff
@@ -186,71 +197,77 @@ func (tracker *StreamsTrackerImpl) Run(ctx context.Context) error {
...
 	}
 }
 
-func (tracker *StreamsTrackerImpl) AddStream(streamId shared.StreamId) error {
+func (tracker *StreamsTrackerImpl) AddStream(
+	streamId shared.StreamId,
+	applyHistoricalStreamContents bool,
+	applyHistoricalStreamContentFromMiniblockHash []byte,
+) error {
+	// Check if the stream is already tracked, so we won't call the expensice GetStream.
```
clemire, line 216: nice catch!

core/node/crypto/config.go
```diff
@@ -155,6 +157,10 @@ type OnChainSettings struct {
...
+	// Options: 1 means enabled, 0 means disabled.
+	NotificationsColdStreamsEnabled uint64 `mapstructure:"notifications.coldStreamsEnabled"`
```
sergekh2, line 162: here and above, it's better to have key lowercased (smart contract needs it to be lowercased, then it's not confusing).

core/node/rpc/notification_test.go
```diff
@@ -51,6 +51,250 @@ func authenticateNS[T any](
...
+		}
+
+		// Check that we have exactly 1 event hash
+		if len(nc.WebPushNotifications) != 1 {
+			return false
+		}
+
+		_, hasHash1 := nc.WebPushNotifications[eventHash1]
+		return hasHash1
+	}, notificationDeliveryDelay, 2500*time.Millisecond, "Didn't receive expected notifications for stream %s", test.dmStreamID)
```
sergekh2, line 106: On CI timeout this low can lead to flakes: sometimes it's just slow (many tests run in parallel for utilization reasons, which in turn makes timing less consistent)

sergekh2 (approved):

core/node/crypto/config.go
jontra: @sergekh2 I assume you mean the `notifications.coldStreamsEnabled` should all be lowercased, and not the variable name, correct?
The rest of the keys are camel cased, except of two.
I'll change it.

core/node/track_streams/multi_sync_runner.go
jontra: unified

core/node/track_streams/streams_tracker.go
jontra: unified

core/node/track_streams/streams_tracker.go
jontra: Thanks!

core/node/crypto/config.go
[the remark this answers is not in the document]
jontra: I wanted it to be dynamic configuration, so I won't have to make a PR and deploy to change this configuration.

core/node/track_streams/streams_tracker.go
```diff
@@ -163,7 +171,7 @@ func (tracker *StreamsTrackerImpl) Run(ctx context.Context) error {
 			_, loaded := tracker.tracked.LoadOrStore(stream.StreamId(), struct{}{})
 			if !loaded {
 				// start tracking the stream, until the root ctx expires.
-				tracker.multiSyncRunner.AddStream(stream, false)
+				tracker.multiSyncRunner.AddStream(stream, ApplyHistoricalContent{Enabled: false})
```
clemire, line 174: FYI, you can just pass `ApplyHistoricalContent{}` because the zero value of Enabled is false.

core/node/rpc/notification_test.go
sergekh2: I misread the code, though it's 2.5 seconds. 
All good!

jontra:
> @jontra can you also look at the tests in sync_runner_norace_test.go and add some testing to validate this? Doesn't have to be as heavy-weight, even a couple of unit tests are fine.

Yep, added some tests there

core/node/rpc/notification_test.go
```diff
@@ -61,6 +305,9 @@ func TestNotifications(t *testing.T) {
 		ApnPushNotifications: make(map[common.Hash]map[common.Address]int),
 	}
 
+	// enable cold streams, since this should be the default ASAP
+	tester.btc.SetConfigValue(t, ctx, crypto.NotificationsColdStreamsEnabledConfigKey, crypto.ABIEncodeUint64(1))
```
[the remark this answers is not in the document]
jontra: > Perhaps it makes sense to set the config value here as 1 during the initialization of the blockchain test context and allow some tests to explicitly disable it, since we plan for this to be the default?

It is set to 1 here :-)
The tests here create the channels after the initialization of the notification service, so they are not affected by this change. I've set it to 1 here just to be 100% sure (and since it's going to be the default soon)

> But I really do think this makes more sense as a config value, unless you think it's valuable to change it in production without re-deploying, which is actually pretty easy.
Yes, that's why I made it a dynamic config

clemire (approved):

PR #3628, opened 2025-07-15
Allow web workers to perform unpacking

miguel-nascimento:
This PR moves the unpacking to a isolated file, allowing us to have different implementations. 
One of the possible implementation is using `workerpool` and delegating the unpacking process to workers.

I kept the unpacking functions in `sign.ts`, but marked as deprecated, so we slowly migrate. This is not a breaking change.

Without Workers: https://share.firefox.dev/40TynOg
With Workers: https://share.firefox.dev/46I1PdQ

### Why it's not default?

`workerpool` support both browser and node, so we could ship this as a default.
However, we need to register the Worker if its on browser, and this changes if we're using vite / next / other framework.

Due to this, I decided to keep allow the developer to pass the module (`createUnpacker` vs `createWorkerUnpacker`) instead of doing this internally in the SDK.


packages/playground/src/config/wagmi.ts
```diff
@@ -7,9 +7,9 @@ import { injected } from 'wagmi/connectors'
 export const wagmiConfig = createConfig({
     chains: [base, baseSepolia, foundry],
     transports: {
-        [base.id]: http(),
-        [baseSepolia.id]: http(),
-        [foundry.id]: http(),
+        [base.id]: http('https://base-rpc.publicnode.com'),
+        [baseSepolia.id]: http('https://base-sepolia-rpc.publicnode.com'),
+        [foundry.id]: http('http://127.0.0.1:8545'),
```
texuf, line 12: oops

PR #3666, opened 2025-07-17
Stop scrolling back on all channels by default

texuf:
in the spirit of only sync the channel you’re currently looking at, stop scrolling back on channels in the sdk
we have duplciate scrollback logic in the app
cursory testing doesn’t turn over any problems


miguel-nascimento (approved):

texuf:
I think this is going to negatively impact the threads panel.

ttillberg (approved):
nice!

texuf:
> Can we make this change under a `clientOption`? Then we can ship this in a feature flag and see the reaction.

i've been testing this locally and it seems fine. this needs to go, and we should build a better way to ensure that we eventually get all the missing blocks.

PR #3671, opened 2025-07-18
Flaky test: Allow more than one fulfillment message

jontra:
### Description

any online client can respond to a key solicitation request, and it's possible to get multiple fulfillments for a single key solicitation message.
Fix the test to allow more than a single fulfillment message

### DIFF
--- a/packages/sdk/src/tests/multi_ne/ephemeralEvents.test.ts
+++ b/packages/sdk/src/tests/multi_ne/ephemeralEvents.test.ts
@@ -120,9 +120,9 @@ describe('ephemeralEvents', () => {
             ephemeralSolicitations.push(ephemeral ?? false)
         })
 
-        const ephemeralFulfillments: boolean[] = []
-        alice.on('ephemeralKeyFulfillment', (_) => {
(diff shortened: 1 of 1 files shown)


packages/sdk/src/tests/multi_ne/ephemeralEvents.test.ts
```diff
@@ -139,7 +139,12 @@ describe('ephemeralEvents', () => {
...
 
         expect(ephemeralSolicitations).toEqual([true])
-        await waitFor(() => expect(ephemeralFulfillments.length).toEqual(1))
-        expect(ephemeralFulfillments).toEqual([true])
+        
+        // Wait for at least one ephemeral fulfillment
+        await waitFor(() => expect(ephemeralFulfillments.length).toBeGreaterThanOrEqual(1))
+        
+        // Verify all fulfillments are from different clients
+        const uniqueSenders = new Set(ephemeralFulfillments)
```
texuf, line 147: sets in typescript will compare objects by reference, so this will always be true, you need to convert the user address to a string and make a set of that.

PR #3677, opened 2025-07-19
fix(contract): prevent point farming in `executeSwapWithPermit`

shuhuiluo:
### Description

Fixes a medium-severity audit finding where any member could submit another member's swap using `executeSwapWithPermit` and receive the airdrop points intended for the permit signer. This created an incentive for users to monitor and front-run others' swaps to farm points.

The fix ensures points are always minted to the permit owner (`permit.owner`) instead of the transaction submitter (`msg.sender`) in permit-based swaps, while maintaining backward compatibility for regular swaps.

### Changes

- Modified `_afterSwap` function to accept a `payer` parameter instead of hardcoding `msg.sender`
- Updated `executeSwap` to pass `msg.sender` as payer (maintains existing behavior)
- Updated `executeSwapWithPermit` to pass `permit.owner` as payer (fixes the vulnerability)
- Enhanced `test_executeSwapWithPermit_swapTokenToEth` with fuzzed `caller` parameter to verify points go to permit owner
- Enhanced `test_executeSwapWithPermit` with fuzzed `caller` parameter to verify no points are minted in ERC20-to-ERC20 swaps
- Added comprehensive assertions to prevent regression of the point farming vulnerability

### Checklist

### DIFF
--- a/packages/contracts/src/spaces/facets/swap/SwapFacet.sol
+++ b/packages/contracts/src/spaces/facets/swap/SwapFacet.sol
@@ -101,7 +101,7 @@ contract SwapFacet is ISwapFacet, ReentrancyGuardTransient, Entitled, PointsBase
         );
 
         // post-swap processing (points minting and events)
-        _afterSwap(params, amountOut, protocolFee, poster);
+        _afterSwap(params, amountOut, protocolFee, poster, msg.sender);
 
         // handle refunds of unconsumed input tokens
         _handleRefunds(params.tokenIn, tokenInBalanceBefore);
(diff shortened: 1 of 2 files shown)


packages/contracts/src/spaces/facets/swap/SwapFacet.sol
```diff
@@ -177,11 +177,13 @@ contract SwapFacet is ISwapFacet, ReentrancyGuardTransient, Entitled, PointsBase
     /// @param amountOut The amount of output tokens received
     /// @param protocolFee The protocol fee collected for points calculation
     /// @param poster The swap poster address
+    /// @param payer The address that should receive the points
     function _afterSwap(
```
giuseppecrj, line 181: Why is this called afterSwap and not just mintPoints? Is the intention to make this a virtual function?

shuhuiluo: The intention is to encapsulate shared logic after swap. Why spend time on formality?

PR #3678, opened 2025-07-19
include high prio syncCookies in the initial syncStreams call

erikolsson:
This PR adds the high prio sync cookies to the initial `syncStreams` call.
By including them directly instead of waiting to add them using `modifySync`, we're able to shave one call before the high prio stream is added to sync.

No need to sync prioritized spaces at this point, they will be added automatically by the `tick()`.


packages/sdk/src/syncedStreamsLoop.ts
```diff
@@ -335,6 +335,25 @@ export class SyncedStreamsLoop {
                     const syncCookies: SyncCookie[] = []
                     if (this.streamOpts?.useModifySync == true) {
                         this.pendingSyncCookies.push(...Array.from(this.streams.keys()))
+                        // if the stream is a channel, dm, or gdm, add the sync cookie to the initial sync cookies
+                        // prioritized spaces will be added later during the calls to tick()
+                        for (const id of this.highPriorityIds) {
+                            if (
+                                isChannelStreamId(id) ||
+                                isDMChannelStreamId(id) ||
+                                isGDMChannelStreamId(id)
```
texuf, line 344: if (isHighPriorityStreamForSync(id))

texuf (approved):

PR #3679, opened 2025-07-19
process responses before calling ModifySync

erikolsson:
We're currently performing async `ModifySync` logic _before_ we process any received responses. On startup, this potentially delays the processing of received responses while we're making calls to `ModifySync`.
Let's flip the order here.

The `syncId ===` check is already performed inside `onUpdate`, no need to do that here.


packages/sdk/src/syncedStreamsLoop.ts
```diff
@@ -548,6 +548,11 @@ export class SyncedStreamsLoop {
     }
 
     private async tick() {
+        const item = this.responsesQueue.shift()
+        if (item) {
+            await this.onUpdate(item)
+        }
```
texuf, line 554: There is a throw in this function that can trigger now. Can you remove it? image

texuf (changes requested):

PR #3681, opened 2025-07-19
[stream-metadata] public endpoint to fetch user spaces

miguel-nascimento:
This PR adds a endpoint to fetch the spaces that a user is in.

`/user/{userId}/spaces`

It returns a list of space id, address, name, image, description, motto.

Should close TOWNS-29457

### DIFF
--- a/packages/stream-metadata/src/routes/userSpaces.ts
+++ b/packages/stream-metadata/src/routes/userSpaces.ts
@@ -1,0 +1,113 @@
+import { FastifyReply, FastifyRequest } from 'fastify'
+import { StreamPrefix, StreamStateView, makeStreamId, isSpaceStreamId } from '@towns-protocol/sdk'
+import { MembershipOp } from '@towns-protocol/proto'
+import { z } from 'zod'
+import { SpaceAddressFromSpaceId, type ISpaceOwnerBase } from '@towns-protocol/web3'
+
+import { getStream } from '../riverStreamRpcClient'
+import { isValidEthereumAddress } from '../validators'
+import { config } from '../environment'
+import { spaceDapp } from '../contract-utils'
+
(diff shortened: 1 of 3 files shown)


packages/stream-metadata/src/routes/userSpaces.ts
```diff
@@ -0,0 +1,113 @@
...
+	}
+
+	const spaceInfoMap = new Map<string, ISpaceOwnerBase.SpaceStructOutput>()
+	await Promise.all(
+		spaceIds.map(async (spaceId) => {
+			const spaceAddress = SpaceAddressFromSpaceId(spaceId)
+			const spaceInfo = await spaceDapp.spaceOwner.getSpaceInfo(spaceAddress)
+			spaceInfoMap.set(spaceAddress, spaceInfo)
+		}),
+	)
```
miguel-nascimento, line 72: Wonder if we should add a batch limit here. Probably wont be a issue right now, but it can trigger a lot of requests.
(iirc, the spaceDapp isnt using multicall, so each one is triggering a single request to the rpc?)

texuf: Also - multi call?

texuf: seems like this could be problematic, we'll get rate limited by alchemy

miguel-nascimento: Yeah that’s what I’m doing with images. Just retruning the URL here so in the client we can pass to a img html component

miguel-nascimento: Multicall https://viem.sh/docs/contract/multicall.html

texuf (approved):

PR #3689, opened 2025-07-20
separate `tick()` into `tickResponses()` and `tickModifications()`

erikolsson:
This PR separates the syncStreams `tick()` into two parallel queues: `tickResponses()` and `tickModifications()` — this way they can do work in parallel. Specifically the calls to `ModifySync` were previously blocking processing of responses.

### DIFF
--- a/packages/sdk/src/syncedStreamsLoop.ts
+++ b/packages/sdk/src/syncedStreamsLoop.ts
@@ -563,6 +563,8 @@ export class SyncedStreamsLoop {
         queueMicrotask(() => {
             tick.catch((e) => this.logError('ProcessResponseTick Error', e)).finally(() => {
                 this.inProgressResponseTick = undefined
+                // Tick both queues, not just the response queue. Handled responses affect the ModifySync flow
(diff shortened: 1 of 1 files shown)


packages/sdk/src/syncedStreamsLoop.ts
```diff
@@ -521,38 +522,60 @@ export class SyncedStreamsLoop {
...
-        this.inProgressTick = tick
+        const tick = this.tickResponses()
+        this.inProgressResponseTick = tick
         queueMicrotask(() => {
-            tick.catch((e) => this.logError('ProcessTick Error', e)).finally(() => {
-                this.inProgressTick = undefined
+            tick.catch((e) => this.logError('ProcessResponseTick Error', e)).finally(() => {
+                this.inProgressResponseTick = undefined
                 setTimeout(() => this.checkStartTicking())
             })
```
texuf, line 567: Can you leave a comment why this ticks both?

texuf (approved):

PR #3694, opened 2025-07-21
Lock syncer per address in syncer set

rhbntt:
- Lock syncer per address
- Add unit tests

### DIFF
--- a/core/node/events/stream_reconcile_task.go
+++ b/core/node/events/stream_reconcile_task.go
@@ -178,9 +178,8 @@ func (s *StreamCache) reconciliationTask(
 
 					return nil, xsync.DeleteOp
 				})
+			return
 		}
-
-		return
 	}
 
 	schedule := false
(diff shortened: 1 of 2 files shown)


bas-vk:
This does more than just updating the implementation of `getOrCreateSyncer`. Please drop changes that are non-related to this change.

core/node/rpc/sync/client/syncer_set.go
```diff
@@ -617,62 +618,66 @@ func (ss *SyncerSet) selectNodeForStream(ctx context.Context, cookie *SyncCookie
...
-			ss.streamID2Syncer.Delete,
-			ss.otelTracer,
-		)
-	} else {
-		client, err := ss.nodeRegistry.GetStreamServiceClientForAddress(nodeAddress)
-		if err != nil {
-			return nil, AsRiverError(err).Tag("remoteSyncerAddr", nodeAddress)
-		}
+func (ss *SyncerSet) getOrCreateSyncer(nodeAddress common.Address) (resultSyncer StreamsSyncer, resultErr error) {
+	ss.syncers.Compute(
```
bas-vk, line 622: `Compute` locks an entire hash bucket. So this approach doesn't guarantee parallel creation/retrieval of syncers.

core/node/rpc/sync/client/syncer_set.go
```diff
@@ -617,18 +619,35 @@ func (ss *SyncerSet) selectNodeForStream(ctx context.Context, cookie *SyncCookie
...
 
-	if syncer, found := ss.syncers[nodeAddress]; found {
-		return syncer, nil
+		// Initialization failed, return error
+		return nil, RiverError(Err_INTERNAL, "Syncer initialization failed")
 	}
 
-	var syncer StreamsSyncer
+	// Lock the given syncer until it is initialized
+	syncer.Lock()
```
bas-vk, line 646: - 2 go-routines call `getOrCreateSyncer` at the same time and request a syncer that isn't initialised.
- They end up with the same syncer instance on line 630.
- Both try to grab the lock on line 646.
- One succeeds and initialises the syncer and releases the lock on leaving the function.
- The second go-routine now grabs the lock and initialises the syncer again overwriting the previous syncer.

PR #3698, opened 2025-07-21
IMembershipShim: change listenForMembershipToken to use getLogs

salzbrenner:
### Description

Our SpaceDapp sets the pollingInterval on the provider to 250ms - which is probably unnecessary, but we can change that in a different PR
```
        if ('pollingInterval' in provider && typeof provider.pollingInterval === 'number') {
            provider.pollingInterval = 250
        }
```
The membership listener was using the ethers/typechain filter listener, which used this pollingInterval, which can result in extraneous rpc requests.

This PR changes that behavior such that:

- it uses getLogs w/ an interval of 1_000 ms to reduce rpc calls
- listenForMembershipToken returns a function. Call the returned function to start polling from the block number saved in the call to listenForMembershipToken (see example comment)

### DIFF
--- a/packages/web3/src/space/IMembershipShim.ts
+++ b/packages/web3/src/space/IMembershipShim.ts
@@ -47,6 +47,10 @@ export class IMembershipShim extends BaseContractShim<typeof connect> {
 
     // If the caller doesn't provide an abort controller, create one and set a timeout
     // to abort the call after 20 seconds.
+    // exmample:
+    // const startListener = space.Membership.listenForMembershipToken(recipient)
+    // const issuedListener = startListener()
+    // const result = await issuedListener()
     listenForMembershipToken(
         receiver: string,
         providedAbortController?: AbortController,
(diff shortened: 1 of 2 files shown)


packages/web3/src/space/IMembershipShim.ts
```diff
@@ -47,66 +47,117 @@ export class IMembershipShim extends BaseContractShim<typeof connect> {
...
+            const abortController = providedAbortController ?? new AbortController()
+            const abortTimeout = providedAbortController
+                ? undefined
+                : setTimeout(() => {
+                      logger.error('joinSpace timeout')
+                      abortController.abort()
+                  }, 20_000)
+
+            const pollingInterval = 1_000
+            let lastCheckedBlock = await provider.getBlockNumber()
```
texuf, line 72: i think you need to pass in a block number here... you need to start with a block number <= to the block you submitted your transaction

salzbrenner: good call

texuf (approved):

PR #3702, opened 2025-07-22
Fix accidentally reverted test fix

texuf:
sorry yoni


jontra:
> sorry yoni

I'll make you pay for it :-)

PR #3710, opened 2025-07-22
Add validation for reactions

salzbrenner:
This PR adds validation in nodes for reaction permissions, and client-side filtering for malformed reaction events.

**Changes**
Nodes
- Added permission checks for REACTION and REPLY message interaction types in nodes

SDK
- Added tags field to TimelineEvent interface and isReactionInteraction helper
- Added filtering logic to ignore malformed reaction events
- Added comprehensive integration tests

TODO
- the client sdk [already does checks](https://github.com/towns-protocol/towns/blob/main/packages/sdk/src/client.ts#L1810) around writes and reactions - which were in place b/c there was nothing in node. We can remove these after this rolls out.

Test Plan
  - Added integration tests for malformed reaction filtering

### DIFF
--- a/core/node/rules/rule_builder_ae.go
+++ b/core/node/rules/rule_builder_ae.go
@@ -107,20 +107,22 @@ func (re *ruleBuilderAEImpl) requireChannelMessagePermissions(params *aeParams) 
 		interactionType := params.parsedEvent.Event.Tags.MessageInteractionType
 
 		switch interactionType {
+		// reactions - should be allowed if write or react
+		// redactions - should be allowed if write or react (users can always redact their own reactions)
+		// tips - if you can react in a channel, you can tip
+		// trades - if you can read/react in a channel, you should be able to make trades regardless of write
 		case MessageInteractionType_MESSAGE_INTERACTION_TYPE_REACTION,
+			MessageInteractionType_MESSAGE_INTERACTION_TYPE_TIP,
+			MessageInteractionType_MESSAGE_INTERACTION_TYPE_TRADE,
 			MessageInteractionType_MESSAGE_INTERACTION_TYPE_REDACTION:
-			// reactions - should be allowed if write or react
-			// redactions - should be allowed if write or react (users can always redact their own reactions)
 			return re.requireOneOfChainAuths(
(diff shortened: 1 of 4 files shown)


core/node/rules/rule_builder_ae.go
```diff
@@ -101,6 +102,33 @@ func (re *ruleBuilderAEImpl) fail(err error) ruleBuilderAE {
...
+		interactionType := params.parsedEvent.Event.Tags.MessageInteractionType
+
+		switch interactionType {
+		case MessageInteractionType_MESSAGE_INTERACTION_TYPE_REACTION,
+			MessageInteractionType_MESSAGE_INTERACTION_TYPE_REDACTION:
+			// reactions - should be allowed if write or react
+			// redactions - should be allowed if write or react (users can always redact their own reactions)
+			return re.requireOneOfChainAuths(
+				params.channelEntitlements(auth.PermissionReact),
+				params.channelEntitlements(auth.PermissionWrite),
```
texuf, line 116: 1. can you flip these? I expect that write will be way more common than react and we should check that one first (assuming they're not checked in parallel). Also it keeps it the same as it was before.
2. REPLY should be up here right? Isn't "React/Reply" the setting in the ui?
3. TIP and TRADE should definitely be up here
4. EDIT needs to be up here if REPLY is up here

packages/sdk/src/views/streams/timelinesModel.ts
```diff
@@ -347,6 +348,21 @@ export function makeTimelinesViewInterface(
...
         updatingEventId?: string,
     ) {
+        // if it was tagged as a reaction but the content is not a reaction, just ignore
+        // but don't filter out encrypted events as they may decrypt to valid reactions
+        if (
+            isReactionInteraction(event.tags) &&
+            event.content?.kind !== RiverTimelineEvent.Reaction &&
+            event.content?.kind !== RiverTimelineEvent.ChannelMessageEncrypted &&
+            event.content?.kind !== RiverTimelineEvent.ChannelMessageEncryptedWithRef
+        ) {
```
texuf, line 358: i think we may need to flip this around, if it's a top level POST that's not tagged as a POST then it needs to be ignored..

PR #3718, opened 2025-07-23
Snapshot migration 0005

erikolsson:
this PR adds an aggressive snapshot migration + removes empty key solicitations

Largely based on migration 0004 with a few tweaks

### DIFF
--- a/core/node/events/migrations/snapshot_migration_0005.go
+++ b/core/node/events/migrations/snapshot_migration_0005.go
@@ -2,6 +2,7 @@ package migrations
 
 import (
 	"encoding/hex"
+	"slices"
 
 	. "github.com/towns-protocol/towns/core/node/protocol"
 )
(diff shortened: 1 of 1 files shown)


core/node/events/migrations/snapshot_migration_0005.go
```diff
@@ -0,0 +1,92 @@
...
+					if !lostSessionIdsSet[sessionId] {
+						newSessionIds = append(newSessionIds, sessionId)
+					}
+				}
+				solicitation.SessionIds = newSessionIds
+			}
+
+			// Remove empty solicitations
+			var filteredSolicitations []*MemberPayload_KeySolicitation
+			for _, solicitation := range member.Solicitations {
```
bas-vk, line 67: You can use here https://pkg.go.dev/slices#DeleteFunc

bas-vk, line 67: You can use here https://pkg.go.dev/slices#DeleteFunc

PR #3734, opened 2025-07-24
fix snapshot hash checks

erikolsson:
ok this one wasn't easy to track down — we're sometimes seeing this:
image
sometimes when syncing, there's a miniblock header with a valid `snapshotHash`, but without an accompanying snapshot counterpart — that may be a separate bug which I will also look into.


texuf (commented):
This doesn't look right...there should be a snapshot if either of those conditions are true... is there a case where the hashes dont match?

packages/sdk/src/syncedStream.ts
```diff
@@ -198,7 +198,9 @@ export class SyncedStream extends Stream implements ISyncedStream {
         const minipoolEvents = Array.from(this.view.minipoolEvents.values())
 
         const lastSnapshotMiniblockNum =
-            miniblock.header.snapshot !== undefined || miniblock.header.snapshotHash !== undefined
+            miniblock.header.snapshot !== undefined ||
+            (miniblock.header.snapshotHash !== undefined &&
+                bin_equal(miniblock.header.snapshotHash, snapshot?.hash))
```
texuf, line 203: please change this to check explicitly for the condition where miniblock.header.snapshotHash is defined and does not match snapshot?.hash and log it.

texuf: actually this is good

PR #3756, opened 2025-07-26
Added logging to selectNodeForStream in shared syncer set & fixed bug when adding stream to sync

rhbntt:
- Added logging to the shared syncer
- Fixed the following bug: the subscription identifies the incoming stream as syncing when adding it if the first attempt to do so is failed.

### DIFF
--- a/core/node/rpc/sync/client/syncer_set.go
+++ b/core/node/rpc/sync/client/syncer_set.go
@@ -8,18 +8,16 @@ import (
 	"sync/atomic"
 	"time"
 
-	"go.opentelemetry.io/otel/codes"
-
-	"github.com/towns-protocol/towns/core/node/logging"
-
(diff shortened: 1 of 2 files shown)


core/node/rpc/sync/client/syncer_set.go
```diff
@@ -737,6 +763,9 @@ func (ss *SyncerSet) selectNodeForStream(ctx context.Context, cookie *SyncCookie
 		for range remotes {
 			if _, err = ss.getOrCreateSyncer(ctx, selectedNode); err == nil {
 				return selectedNode, true
+			} else {
+				logging.FromCtx(ss.globalCtx).Errorw("Failed to get or create syncer for remote node",
```
bas-vk, line 767: It is an idea to start a sub span for each remote. That way we can get details information what remote is slow or has issues.

PR #3769, opened 2025-07-28
Handle streams in parallel when modifying sync in shared syncer

rhbntt:
- Handling streams in the modify sync request in parallel. It adds a network load but reduces complexity of the code and potential unexpected issues such as deadlocks. 
- Added more unit tests.

### DIFF
--- a/core/node/rpc/sync/client/syncer_set.go
+++ b/core/node/rpc/sync/client/syncer_set.go
@@ -175,6 +175,9 @@ func (ss *SyncerSet) Modify(ctx context.Context, req ModifyRequest) error {
 				if st = ss.processAddingStream(ctx, req.SyncID, cookie, true); st != nil {
 					req.AddingFailureHandler(st)
 				}
+			} else {
+				// Call the failure handler for other error types (e.g., Err_UNAVAILABLE)
+				req.AddingFailureHandler(st)
 			}
 		}(cookie)
 	}


core/node/rpc/sync/client/syncer_set.go
```diff
@@ -37,12 +35,6 @@ type (
 		DebugDropStream(ctx context.Context, streamID StreamId) (bool, error)
 	}
 
-	// StreamCache represents a behavior of the stream cache
-	StreamCache interface {
-		GetStreamWaitForLocal(ctx context.Context, streamId StreamId) (*events.Stream, error)
-		GetStreamNoWait(ctx context.Context, streamId StreamId) (*events.Stream, error)
-	}
-
 	ModifyRequest struct {
```
bas-vk, line 38: Add documentation to all callback in `ModifyRequest` that they must not lock a stream. If they lock a stream there is an opportunity for a deadlock.

core/node/rpc/sync/client/syncer_set.go
```diff
@@ -211,473 +144,261 @@ func (ss *SyncerSet) unlockStream(streamID StreamId) {
...
-			ToRemove:                  req.ToRemove,
-			BackfillingFailureHandler: backfillingFailuresHandler,
-			AddingFailureHandler:      addingFailuresHandler,
-			RemovingFailureHandler:    removingFailuresHandler,
-		}, false); err != nil {
-			return AsRiverError(err).Func("SyncerSet.Modify.second")
-		}
+	if _, found := ss.streamID2Syncer.Load(streamID); found {
+		unlock()
+		return ss.processBackfillingStream(ctx, syncID, syncID, cookie)
```
bas-vk, line 234: After the unlock it is assumed that this stream is already part of the sync session. Is it possible that after unlock is called a remove stream requests comes in that drops the stream from syncing before the backfill request is handled?

bas-vk (approved):

PR #3789, opened 2025-07-30
Add tracing to notification service

rhbntt:


core/node/track_streams/multi_sync_runner.go
```diff
@@ -120,6 +121,14 @@ func (ssr *syncSessionRunner) AddStream(
 	ctx context.Context,
 	record streamSyncInitRecord,
 ) error {
+	if ssr.otelTracer != nil {
```
jontra, line 124: it would be much better to create a common function for the boilerplate code. so we can easily use it in other places
you can use go runtime functions to get the caller name with `runtime.Caller(1)`
and extract the function name with `runtime.FuncForPC(....)`
then add any attributes passed (or if you make it a part of ssr, you can use `ssr.GetSyncId()` to always add the sync id)

PR #3791, opened 2025-07-30
Define Sync V3 interfaces

rhbntt:


core/node/rpc/service_sync_streams.go
```diff
@@ -14,6 +14,10 @@ import (
 	"github.com/towns-protocol/towns/core/node/utils"
 )
 
+func useSyncV3(req connect.AnyRequest) bool {
+	return req.Header().Get(UseSharedSyncHeaderName) == "true"
```
bas-vk, line 18: Shall we use a different header for v3?

begmaroman: Ideally we should but the current one is already integrated with the client and properly adjusted in LD setup. I would just use this one tbh.

PR #3795, opened 2025-07-30
feat: stream sync monitor service

clemire:
## Summary
- Adds a new stream sync monitor service to detect lagging streams on monitored nodes
- Updates StreamFilter interface to support stream modification

## Description

This PR introduces a stream sync monitor service designed to detect when streams are lagging behind on specific monitored nodes. The monitor compares the sync state with the on-chain registry state and captures stack traces when lag exceeds configured thresholds.

### Key Changes:

1. **New StreamFilter Interface**: Updated to accept and return `StreamWithId`, allowing filters to modify which nodes the MultiSyncRunner sees
2. **Stream Sync Monitor Service**: Tracks streams only on configured nodes and monitors for lag
3. **Stack Trace Capture**: When lag exceeds threshold (after grace period), captures stack traces from the syncing node
4. **Periodic Status Reporting**: Reports tracked/lagging stream counts every 15 seconds

### Implementation Details:

- Filters nodes at the StreamFilter level before MultiSyncRunner sees them
- Adjusts replication factor when filtering nodes to prevent slice bounds errors
- Includes a grace period to avoid false positive lag detection
- Tracks placement information (sync ID, node address, minipool gen) for debugging

## Test Plan

- [x] Unit tests pass
- [x] Service starts and tracks streams on monitored nodes
- [ ] Stack trace capture on lag detection (not yet tested - no lag was observed locally)

## Notes

This is a work in progress; however I'd still like to capture it in the repository for later use. The stack trace capture functionality has not been tested yet as no lag has been observed during local testing. Additional testing in an environment with actual stream lag is needed.

## Run instructions

```
./core/env/omega/run.sh sync-monitor -c monitor-river-node-streams.yaml | yarn exec pino-pretty
```

where `monitor-river-node-streams.yaml` contains the following:

```
streamSyncMonitor:
  monitoredNodeAddresses:
    - 0xC5FB45BDb448766135ce634bD1Da34D483ADd382
    - 0xB85dd7d21Bc093DAe6f1b56a20a45ec1770D11dD
    - 0xc69ceBB1e84Ca8dda9a9Ad68D345Fbe4Ac21Fd54
  lagThreshold: 10
  stackTraceOutputDir: "~/repos/monitor-streams"
  streamTracking:
    streamsPerSyncSession: 1000
log:
  level: "debug"
```

### DIFF
--- a/core/node/infra/metrics_factory.go
+++ b/core/node/infra/metrics_factory.go
@@ -41,6 +41,10 @@ type MetricsFactory interface {
 	NewStatusCounterVecEx(name string, help string, labels ...string) *StatusCounterVec
 
 	Registry() *prometheus.Registry
+}
+
+type DebugMetricsFactory interface {
+	MetricsFactory
 
 	// GetMetricsAsMap returns all metrics as a structured map suitable for JSON logging.
 	// This is intended for debugging purposes only.
--- a/core/node/stream_sync_monitor/service.go
+++ b/core/node/stream_sync_monitor/service.go
@@ -6,7 +6,6 @@ import (
 	"net"
 	"net/http"
 	"net/http/pprof"
-	"runtime"
 	"time"
 
 	"github.com/prometheus/client_golang/prometheus"
@@ -33,7 +32,7 @@ func NewService(
 	onChainConfig crypto.OnChainConfiguration,
 	riverRegistry *registries.RiverRegistryContract,
 	nodeRegistries []nodes.NodeRegistry,
-	metricsFactory infra.MetricsFactory,
+	metricsFactory infra.DebugMetricsFactory,
 ) (*Service, error) {
 	monitor, err := sync.NewStreamSyncMonitor(
 		ctx,
@@ -47,8 +46,6 @@ func NewService(
 	if err != nil {
 		return nil, err
 	}
-
-	runtime.SetBlockProfileRate(1)
 
 	// Extract the prometheus registry from the metrics factory
 	var metricsRegistry *prometheus.Registry
@@ -86,7 +83,7 @@ func (s *Service) Run(ctx context.Context) error {
 
 	// Shutdown debug server if it was started
 	if s.debugServer != nil {
-		shutdownCtx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+		shutdownCtx, cancel := context.WithTimeout(ctx, 5*time.Second)
 		defer cancel()
 		if shutdownErr := s.debugServer.Shutdown(shutdownCtx); shutdownErr != nil {
 			log.Errorw("Failed to shutdown debug server", "error", shutdownErr)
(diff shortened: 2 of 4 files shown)


core/node/app_registry/sync/streams_tracker.go
```diff
@@ -72,17 +73,16 @@ func NewAppRegistryStreamsTracker(
...
 
 	return tracker, nil
 }
 
-func (tracker *AppRegistryStreamsTracker) TrackStream(streamId shared.StreamId, _ bool) bool {
-	streamType := streamId.Type()
-
-	return streamType == shared.STREAM_CHANNEL_BIN
+func (tracker *AppRegistryStreamsTracker) TrackStream(stream *river.StreamWithId, _ bool) (bool, *river.StreamWithId) {
+	return stream.StreamId().Type() == shared.STREAM_CHANNEL_BIN, stream
```
bas-vk, line 85: Not sure what always returning the original stream brings?

core/node/infra/metrics_factory.go
```diff
@@ -39,6 +41,10 @@ type MetricsFactory interface {
 	NewStatusCounterVecEx(name string, help string, labels ...string) *StatusCounterVec
 
 	Registry() *prometheus.Registry
+
+	// GetMetricsAsMap returns all metrics as a structured map suitable for JSON logging.
+	// This is intended for debugging purposes only.
+	GetMetricsAsMap() (map[string]interface{}, error)
```
bas-vk, line 47: > This is a work in progress; however I'd still like to capture it in the repository for later use.

Do you want to merge this at some point? This function is never used.

What I sometimes do with functions that intended for debugging/testing is to put them in a separate interface. So in this case:
```
type (
   MetricsFactory interface { ... }
   MetricsFactoryDebug interface { ... }
)
```

That prevents putting functions that are only meant for testing/debugging in production interfaces and makes it clear that these functions are not for use in production flow.

core/node/stream_sync_monitor/service.go
```diff
@@ -0,0 +1,167 @@
...
+		onChainConfig,
+		nodeRegistries,
+		metricsFactory,
+		nil, // otelTracer - can be added later
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	runtime.SetBlockProfileRate(1)
```
bas-vk, line 51: What is the purpose of this?

clemire: I put this in here so I could test some profiling settings on the sync monitor. It's my bad - I'll take it out.

core/node/app_registry/sync/streams_tracker.go
bas-vk: Lets not extend the function signature with returning a modified version of the stream record. The name `TrackStream` doesn't suggest that this function mutates the stream record and its specific to the sync monitor while this interface is also implemented by other services.

Suggestion:
- leave `TrackStream` as is
- for prepping the stream records:
   - implement `StreamSyncMonitor::normalizeStreamRecord` that filters the node address/fixes the repl factor and pass in the normalized record to `TrackStream`. `TrackStream` only checks that the node list contains at least 1 node following the current convention of only checking and returning a bool if it needs to be tracked.
   - or `TrackStream` already accepts the stream record by pointer, so it can already modify the record without returning it. This is not intuitive for a caller but it reduces the non-intuitiviness to just the `StreamSyncMonitor`. I would prepare option 1 but am okay if you follow this approach if it makes more sense for you.

PR #3804, opened 2025-08-01
Design proposal for using queue for non-blocking syncer

sergekh2:
### Description

Design proposal for using queue for non-blocking syncer.

### DIFF
--- a/core/node/rpc/sync/v3/eventbus/eventbus.go
+++ b/core/node/rpc/sync/v3/eventbus/eventbus.go
@@ -1,0 +1,172 @@
+package eventbus
+
+import (
+	"context"
+	"slices"
+
+	. "github.com/towns-protocol/towns/core/node/protocol"
+	. "github.com/towns-protocol/towns/core/node/shared"
+	"github.com/towns-protocol/towns/core/node/utils/dynmsgbuf"
+)
+
+type (
+	// StreamSubscriber subscribes to stream updates.
+	StreamSubscriber interface {
+		// OnUpdate is called when there is a stream update available.
+		// Each subscriber receives its own copy of the update.
+		//
+		// Subscribers MUST NOT block when processing the given update.
+		//
+		// When update.SyncOp is SyncOp_SYNC_DOWN the subscriber is automatically unsubscribed
+		// from future updates on the stream. It is expected that this update is sent to the client
+		// and that the client will resubscribe with a backoff mechanism.
+		OnUpdate(update *SyncStreamsResponse)
+	}
+
+	// StreamSubscriptionManager allows subscribers to subscribe and unsubscribe on/from stream updates.
+	StreamSubscriptionManager interface {
+		// Subscribe on stream events.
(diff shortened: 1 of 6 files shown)


core/docs/features/sync-queue-design.md
```diff
@@ -0,0 +1,168 @@
...
+`SyncStreamHandlerRegistry` is used to look up `SyncStreamHandler` by a sync id to dispatch `ModifySync` and `CancelSync` calls.
+
+`SyncStreamHandler` sends messages to `EventBus` to subscribe or to unsubscribe from the streams.
+
+`EventBus` streams all potential updates for the requested streams to the `SyncStreamHandler` queue. When `Subscribe` message is received,
+`EventBus` requests backfill for the given stream from the `LocalSyncer` or `RemoteSyncer`. `SyncStreamHandler` can receive updates
+before relevant backfill is complete and should filter such updates out and not send them to the client.
+
+`SyncStreamHandler` and `EventBus` communicate only through message passing. `SyncStreamHandler` provides `OnHandlerUpdate` interface which
+provides relevant update methods, implementations of these methods save updates to the `SyncStreamHandler` internal queue.
```
bas-vk, line 92: > `SyncStreamHandler` provides `OnHandlerUpdate` interface which provides relevant update methods

To be clear, there will be just one `OnHandlerUpdate` function? I'm a bit confused by the `relevant update methodS`? Want to make sure that we keep ordering of updates by design and not by implementation.

core/docs/features/sync-queue-design.md
```diff
@@ -0,0 +1,168 @@
...
+
+`RemoteSyncer` is responsible for streaming updates from a single remote stream into the `EventBus` queue.
+
+`RemoteSyncer` uses `SyncStreams` call to the remote node to request updates. It creates a new `SyncStreams` call for
+each requested stream. I.e., `RemoteSyncer` tracks a single stream and uses a single `SyncStreams` exclusively for this stream.
+
+Backfills can be requested while connection to the remote node is being established. In such case `RemoteSyncer`
+should save backfill requests and send them to the remote node once `SyncStreams` is ready and sync id is received.
+
+Lifecycle of a single `RemoteSyncer` is INIT->STREAM->SHUTDOWN. On shutdown, `SYNC_DOWN` is sent to the `EventBus`
```
bas-vk, line 125: > `INIT->STREAM->SHUTDOWN`

What happens if the remote node drops off, so the transition `INIT->STREAM->REMOTE_GONE` (or is `SHUTDOWN` the state after the remote is gone and not when the local node makes a planned shutdown)?

The `RemoteSyncer` will emit the `SYNC_DOWN` update to the `EventBus` and this update is propagated to all `SyncHandlers` and eventually the clients. And once a client resubscribes to the stream a new `SyncStreamHandler` instance for that stream is started because all sync handler unsubscribed from it automatically when the stream was dropped from the lookup map in the `EventBus`?

PR #3808, opened 2025-08-01
ci: add Go code formatting check

clemire:
## Summary
- Add a new CI step that checks Go code formatting using the fmt.sh script
- The step runs after linting and will fail if any Go files are not properly formatted
- This ensures consistent code style across the Go codebase

## Test plan
- This PR should fail CI if there are unformatted Go files in the core directory
- Once all Go files are properly formatted (by running ./fmt.sh in the core directory), CI should pass

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -152,6 +152,44 @@ jobs:
                     exit 1
                   fi
 
+            - name: Check Go Formatting
+              if: ${{ !cancelled() && steps.setup_success.outcome == 'success' }}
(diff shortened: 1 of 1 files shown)


clemire:
> It would be nice to locally format the code automatically (maybe git hook), so it won't fail on CI

I think that one is up to the individual to set up in their editor of choice. But I do wonder if this fmt.sh script should be updated; possibly it's using outdated tools.

PR #3817, opened 2025-08-03
Added blocklisted streams filter to notification service

begmaroman:


core/node/notifications/sync/streams_tracker.go
```diff
@@ -84,7 +84,13 @@ func (tracker *NotificationsStreamsTracker) coldStreamsEnabled() bool {
 
 // TrackStream returns true if the given streamID must be tracked for notifications.
 func (tracker *NotificationsStreamsTracker) TrackStream(streamID shared.StreamId, isInit bool) bool {
```
bas-vk, line 86: There is also `AppRegistryStreamsTracker#TrackStreams`

begmaroman: Done. Moved this logic to track_streams as an extra filter.

PR #3828, opened 2025-08-04
feat(contract): add no points check to drop facet

giuseppecrj:
- Introduced a new address and key in the DropFacetTest contract.
- Added a test case for account claim with no points, verifying expected balance outcomes.
- Updated claim logic to remove unnecessary default values for claim points.
- Enhanced the Merkle tree construction to include new data for claims.
- Separate burn points logic to internal function


shuhuiluo:
The title should be modified.

shuhuiluo (approved):
Lgtm.

PR #3831, opened 2025-08-05
Add abort controller to modifySync call in syncStreamsLoop

texuf:


texuf:
> > My one comment would be that perhaps there's an easier world where there's only a single abortController, since there's only a single syncStreams at a time.. but maybe not possible?
> 
> Nvm I'm a moron, ignore...

The modify sync calls theoretically overlap within the same syncId, but shouldn't overlap between syncIds... we might be able to share it. I'm not going to land this today since it's not causing a huge problem right now.

packages/sdk/src/syncedStreamsLoop.ts
```diff
@@ -217,6 +218,9 @@ export class SyncedStreamsLoop {
                 this.log('stopSync syncState', syncState)
                 this.log('stopSync syncLoop', syncLoop)
                 this.log('stopSync syncId', syncId)
+                if (syncId) {
+                    this.abortControllers[syncId]?.forEach((x) => x.abort())
```
bas-vk, line 222: I'm not familiar with react and don't understand the relation between abortController and cancelSync. But would it make sense to delete `abortController[syncId]` always after this?

PR #3890, opened 2025-08-15
Remove app name from notifications subscriptions primary key and unsubscribe message

jontra:
We do not need the app name field in the primary key, since the device token and key_auth/key_p256dh pair are unique per app.
Also, there is no need to specify the app when unsubscribing, we just need to key (token/key_auth/key_p256dh)


core/node/storage/notification_migrations/000004_update_primary_keys.up.sql
```diff
@@ -0,0 +1,10 @@
+-- Update primary keys to remove app_name field from composite keys
+-- The app_name column remains in the tables for historical data
+
+-- Update webpushsubscriptions primary key
+ALTER TABLE webpushsubscriptions DROP CONSTRAINT webpushsubscriptions_pkey;
+ALTER TABLE webpushsubscriptions ADD PRIMARY KEY (key_auth, key_p256dh);
+
+-- Update apnpushsubscriptions primary key  
+ALTER TABLE apnpushsubscriptions DROP CONSTRAINT apnpushsubscriptions_pkey;
+ALTER TABLE apnpushsubscriptions ADD PRIMARY KEY (device_token);
```
sergekh2, line 10: This can take a bit on a db with data, any estimate/concern on how long is it going to take?

protocol/notifications.proto
```diff
@@ -247,8 +247,6 @@ message SubscribeWebPushResponse {}
 message UnsubscribeWebPushRequest {
   // subscription holds the subscribe data that is returned by the browser when subscribing
   WebPushSubscriptionObject subscription = 1;
-  // app identifies which app is making the unsubscription request
```
sergekh2, line 250: please always use reserved keyword to list already used ids, otherwise there can be a conflict in the future.

protocol/notifications.proto
```diff
@@ -283,8 +281,6 @@ message SubscribeAPNResponse {}
 message UnsubscribeAPNRequest {
   // APN device token
   bytes device_token = 1;
-  // app identifies which app is making the unsubscription request
-  string app = 2;
```
sergekh2, line 287: same

PR #3932, opened 2025-08-22
fix(generated): fix npm artifact download in Docker builds

shuhuiluo:
- Fix `tar` command working directory issue in `downloadArtifactsFromNpm()`
- Add error logging to diagnose download failures
- Move contracts directory check before Foundry installation to fail fast

The tar command was failing in Docker because it ran from wrong directory. npm pack creates .tgz in tempDir but tar was running from package root. Now both commands use `cwd: tempDir` for consistency.

This resolves Docker build failures for services that don't include contracts package.


shuhuiluo:
@coderabbitai review

PR #3954, opened 2025-08-25
feat(sdk): add `app_address` in stream view

miguel-nascimento:
So we can add support to check if a user is a not in the SDK-level


packages/sdk/src/streamStateView_UserInbox.ts
miguel-nascimento, line 1: Not sure if I need to add this field everywhere...

texuf (approved):

PR #3965, opened 2025-08-27
feat(media): allow dynamic resizing

ttillberg:
wipped this together with claude, works well as a POC but need an extra eye on optimisation, cache control and security

achieved features:
- allow image dynamic resizing by suffixing ?size=100x ?size=100x100 ?size=x100
- resize animated gifs
- transform all formats to modern webp

### Changes

- added `sharp` package
- implemented resizing for space, profile and media routes

### Checklist


texuf:
@Tee-S can you find our cache policy for cloudfront? I'm curious about how we handle query params.

packages/stream-metadata/src/routes/media.ts
```diff
@@ -23,10 +24,12 @@ const querySchema = z.object({
 		.string()
 		.transform((value) => bin_fromHexString(value))
 		.optional(),
+	size: z.string().optional(),
 })
 
 const CACHE_CONTROL = {
 	200: 'public, max-age=31536000, immutable',
+	'200-resized': 'public, max-age=86400, s-maxage=31536000',
```
texuf, line 32: this can also be immutable... why not?

packages/stream-metadata/src/routes/media.ts
```diff
@@ -53,8 +56,19 @@ export async function fetchMedia(request: FastifyRequest, reply: FastifyReply) {
...
+
+	// Parse and validate size parameter
+	const sizeOptions = parseSizeParam(size)
+	if (!sizeOptions.isValid) {
+		logger.info({ size }, 'Invalid size parameter format')
+		return reply
+			.code(400)
+			.header('Cache-Control', CACHE_CONTROL['4xx'])
+			.send({ error: 'Bad Request', message: 'Size must be in format WxH, xH, or Wx' })
+	}
```
texuf, line 69: do we require a size now?

packages/stream-metadata/src/routes/profileImage.ts
```diff
@@ -85,30 +117,94 @@ export async function fetchUserProfileImage(request: FastifyRequest, reply: Fast
...
-		}?key=${bin_toHexString(key)}&iv=${bin_toHexString(iv)}`
-
-		return (
-			reply
-				// client should cache the image for 30 seconds, and the CDN for 5 minutes
-				// after 30 seconds, the client will check the CDN for a new image
-				.header('Cache-Control', CACHE_CONTROL[307])
-				.redirect(redirectUrl, 307)
-		)
+
```
texuf, line 120: please remove this and just redirect with the size params.

packages/stream-metadata/src/routes/spaceImage.ts
```diff
@@ -19,30 +18,68 @@ const paramsSchema = z.object({
...
 		logger.info(errorMessage)
 		return reply
 			.code(400)
 			.header('Cache-Control', CACHE_CONTROL[400])
 			.send({ error: 'Bad Request', message: errorMessage })
 	}
 
-	const { spaceAddress, eventId } = parseResult.data
-	logger.info({ spaceAddress, eventId }, 'Fetching space image')
+	if (!queryResult.success) {
```
texuf, line 51: do we require size params now? seems like we should keep this backwards compatible

texuf (changes requested):
please remove the duplicated code and just stick with the redirects

packages/stream-metadata/src/routes/media.ts
ttillberg: an `undefined` size queryParam is a valid size queryParam  🤷 

```ts
export function parseSizeParam(size?: string): ParsedSizeParam {
	if (!size) {
		return { isValid: true } // No size param is valid (original size)
	}
    ...
```

ttillberg:
> @ttillberg Looking good, @Tee-S needs to update our query string cache policy - can we have a set number of sizes so that we can't have infinite number of possible cached images in our cloudfront?

Having the flexibility to customise the presets client side removes a lot of friction.  It's pretty standard for image resizing services. 

My only concern would be if someone want to exploit it to keep the service unreachable but I guess this could be tackled from another angle.

Tee-S:
> > @ttillberg Looking good, @Tee-S needs to update our query string cache policy - can we have a set number of sizes so that we can't have infinite number of possible cached images in our cloudfront?
> 
> Having the flexibility to customise the presets client side removes a lot of friction. It's pretty standard for image resizing services.
> 
> My only concern would be if someone want to exploit it to keep the service unreachable but I guess this could be tackled from another angle.

I've added query strings to the cache key for `alpha` in cloudfront. I've set it to `all` query strings atm, but we can change this to only use `size` if that makes more sense.

PR #3973, opened 2025-08-28
Rahul/towns 30678 media stream event storage offloading

rahul-ramesh:
### Description
Currently, the media stream data is stored directly in the database, causing huge sizes on disk.
Use s3 buckets to store and retrieve media stream data.

### Changes
Additional table in the DB to track media stream external storage.
Interface for file management on s3.
RPC changes to offload media payload

### Checklist

### DIFF
--- a/core/node/storage/pg_ephemeral_store.go
+++ b/core/node/storage/pg_ephemeral_store.go
@@ -442,7 +442,7 @@ func (s *PostgresStreamStore) WriteExternalMediaStreamPartUploadInfo(
 	ctx context.Context,
 	streamId StreamId,
 	miniblock int64,
-	etag string,
+	etag Etag,
 	length int,
 ) error {
 	ctx, cancel := context.WithTimeout(ctx, 5*time.Second)
@@ -465,7 +465,7 @@ func (s *PostgresStreamStore) writeExternalMediaStreamPartUploadInfoTx(
 	tx pgx.Tx,
 	streamId StreamId,
 	miniblock int64,
-	etag string,
+	etag Etag,
 	length int,
 ) error {
 	// First, update the external_media_uploads table with the new etag
@@ -476,14 +476,8 @@ func (s *PostgresStreamStore) writeExternalMediaStreamPartUploadInfoTx(
 		DO UPDATE SET etags = external_media_uploads.etags || $2
 	`
 
-	// Add the new etag to the JSONB array
-	etagObj := Etag{
-		Miniblock: int(miniblock),
-		Etag:      etag,
-	}
-
 	// Marshal the struct to JSON
-	etagJSONBytes, err := json.Marshal(etagObj)
+	etagJSONBytes, err := json.Marshal(etag)
 	if err != nil {
 		return err
 	}
--- a/core/node/rpc/external_media_stream_test.go
+++ b/core/node/rpc/external_media_stream_test.go
@@ -1,0 +1,313 @@
+package rpc
+
+import (
+	"context"
+	"fmt"
+	"sort"
+	"testing"
+
+	"connectrpc.com/connect"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/stretchr/testify/require"
+
+	"github.com/towns-protocol/towns/core/config"
+	"github.com/towns-protocol/towns/core/node/events"
+	"github.com/towns-protocol/towns/core/node/protocol"
+	. "github.com/towns-protocol/towns/core/node/shared"
+	"github.com/towns-protocol/towns/core/node/storage"
+	"github.com/towns-protocol/towns/core/node/testutils"
+)
+
+// MockExternalMediaStorage is a mock implementation of ExternalMediaStorage for testing
+type MockExternalMediaStorage struct {
+	createExternalMediaStreamCalls []StreamId
+	uploadPartCalls                []UploadPartCall
+	completeUploadCalls            []CompleteUploadCall
+	abortUploadCalls               []AbortUploadCall
+	createExternalMediaStreamError error
+	uploadPartError                error
+	completeUploadError            error
+	abortUploadError               error
+}
+
+// Helper methods for test verification
+func (m *MockExternalMediaStorage) Reset() {
+	m.createExternalMediaStreamCalls = nil
+	m.uploadPartCalls = nil
+	m.completeUploadCalls = nil
+	m.abortUploadCalls = nil
+	m.createExternalMediaStreamError = nil
+	m.uploadPartError = nil
+	m.completeUploadError = nil
+	m.abortUploadError = nil
+}
+
+func (m *MockExternalMediaStorage) GetCreateExternalMediaStreamCalls() []StreamId {
+	return m.createExternalMediaStreamCalls
+}
+
+func (m *MockExternalMediaStorage) GetUploadPartCalls() []UploadPartCall {
+	return m.uploadPartCalls
+}
+
+func (m *MockExternalMediaStorage) GetCompleteUploadCalls() []CompleteUploadCall {
+	return m.completeUploadCalls
+}
+
+func (m *MockExternalMediaStorage) GetAbortUploadCalls() []AbortUploadCall {
+	return m.abortUploadCalls
+}
+
+func (m *MockExternalMediaStorage) SetCreateExternalMediaStreamError(err error) {
+	m.createExternalMediaStreamError = err
+}
+
+func (m *MockExternalMediaStorage) SetUploadPartError(err error) {
+	m.uploadPartError = err
+}
+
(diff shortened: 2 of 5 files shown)


core/node/storage/migrations/000009_external_media.up.sql
```diff
@@ -0,0 +1,6 @@
+CREATE TABLE external_media_streams (
+    stream_id BIGINT PRIMARY KEY,                    -- Stream ID as primary key
+    upload_id TEXT NOT NULL DEFAULT '',              -- upload ID for multipart uploads
```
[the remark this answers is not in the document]
rahul-ramesh: good idea!

core/node/events/stream.go
```diff
@@ -671,6 +672,54 @@ func (s *Stream) GetMiniblocks(
...
+		errChan := make(chan error, len(blocks))
+
+		for i, block := range blocks {
+			wg.Add(1)
+			go func(idx int, blk *storage.MiniblockDescriptor) {
+				defer wg.Done()
+
+				rangeHeader, err := s.params.Storage.GetExternalMediaStreamChunkRangeByMiniblock(ctx, blk.Number)
+				if err != nil {
+					errChan <- fmt.Errorf("failed to get range header for block %d: %w", blk.Number, err)
```
sergekh2, line 697: For all errors please use RiverError with appropriate error code - it makes it way easier to deal with errors when they sent through RPC or logged

core/node/events/stream.go
```diff
@@ -671,6 +672,54 @@ func (s *Stream) GetMiniblocks(
...
+		return nil, false, err
+	}
+	// TODO check if location matches the external storage bucket
+	if location != "" {
+		client, err := storage.CreateExternalClient()
+		if err != nil {
+			return nil, false, err
+		}
+		// for each block, get the data from external storage in parallel
+		var wg sync.WaitGroup
```
sergekh2, line 687: I think there is no need for parallel read: blocks should be continues, to it should be a single range read.

core/node/storage/migrations/000009_external_media.up.sql
```diff
@@ -0,0 +1,18 @@
+-- Table for per-stream data
+CREATE TABLE external_media_uploads (
+    stream_id CHAR(64) PRIMARY KEY,                  -- Stream ID as primary key (matches other tables)
+    upload_id TEXT NOT NULL DEFAULT '',              -- upload ID for multipart uploads
+    next_part INT NOT NULL DEFAULT 0,                -- next part number to use for multipart upload
+    etags JSONB NOT NULL DEFAULT '[]',               -- etags for each part in the multipart upload
+);
+
+-- Table for per-chunk data
+CREATE TABLE external_media_markers (
```
sergekh2, line 10: lets partition this table in the same way other tables are partitioned, see https://github.com/towns-protocol/towns/blob/main/core/node/storage/migrations/000006_fixed_stream_partitions.up.sql

core/node/storage/migrations/000009_external_media.up.sql
```diff
@@ -0,0 +1,18 @@
+-- Table for per-stream data
+CREATE TABLE external_media_uploads (
```
sergekh2, line 2: should have primary key stream_id

core/node/storage/migrations/000009_external_media.up.sql
```diff
@@ -0,0 +1,18 @@
...
+    etags JSONB NOT NULL DEFAULT '[]',               -- etags for each part in the multipart upload
+);
+
+-- Table for per-chunk data
+CREATE TABLE external_media_markers (
+    stream_id CHAR(64) NOT NULL,                     -- Stream ID (foreign key)
+    miniblock INT NOT NULL,                          -- Miniblock number
+    start_bytes BIGINT NOT NULL,                     -- Start byte position for this chunk
+    end_bytes BIGINT NOT NULL,                       -- End byte position for this chunk
+    UNIQUE(miniblock),
```
sergekh2, line 15: instead of UNIQUE should have primary key stream_id, miniblock

core/node/storage/pg_ephemeral_store.go
```diff
@@ -352,3 +356,317 @@ func (s *PostgresStreamStore) IsStreamEphemeral(ctx context.Context, streamId St
...
+		pgx.ReadWrite,
+		func(ctx context.Context, tx pgx.Tx) error {
+			return s.WriteExternalMediaStreamPartInfoTx(ctx, tx, streamId, miniblock, partNumber, etag, length)
+		},
+		nil,
+		"streamId", streamId,
+	)
+}
+
+func (s *PostgresStreamStore) WriteExternalMediaStreamPartInfoTx(
```
sergekh2, line 460:
```suggestion
func (s *PostgresStreamStore) writeExternalMediaStreamPartInfoTx(
```

core/node/storage/pg_ephemeral_store.go
```diff
@@ -352,3 +356,317 @@ func (s *PostgresStreamStore) IsStreamEphemeral(ctx context.Context, streamId St
...
+	ctx context.Context,
+	tx pgx.Tx,
+	streamId StreamId,
+	miniblock int64,
+	partNumber int,
+	etag string,
+	length int,
+) error {
+	// First, update the external_media_uploads table with the new etag
```
sergekh2, line 469: lets lock stream record in es table in each transaction (see other transactions for example), with "read committed" transactions this ensures consistency in updates.

core/node/storage/migrations/000009_external_media.up.sql
```diff
@@ -0,0 +1,18 @@
+-- Table for per-stream data
+CREATE TABLE external_media_uploads (
+    stream_id CHAR(64) PRIMARY KEY,                  -- Stream ID as primary key (matches other tables)
+    upload_id TEXT NOT NULL DEFAULT '',              -- upload ID for multipart uploads
+    next_part INT NOT NULL DEFAULT 0,                -- next part number to use for multipart upload
```
sergekh2, line 5: I think next_part can be consistently derived from external_media_markers table, not sure if there is a need for a separate counter?

core/node/storage/migrations/000009_external_media.up.sql
```diff
@@ -0,0 +1,18 @@
...
+    etags JSONB NOT NULL DEFAULT '[]',               -- etags for each part in the multipart upload
```
sergekh2, line 6: should etags go into external_media_markers? there is etag per miniblock

core/node/storage/migrations/000009_external_media.up.sql
rahul-ramesh: sounds good! I'll make that change

core/node/storage/migrations/000009_external_media.up.sql
rahul-ramesh: done!

core/node/rpc/node2node_ephemeral.go
```diff
@@ -104,7 +125,40 @@ func (s *Service) saveEphemeralMiniblock(ctx context.Context, req *SaveEphemeral
...
+		if s.storage.WriteExternalMediaStreamPartInfo(
+			ctx,
+			streamId,
+			storageMb.Number,
+			partNum,
+			etag,
+			len(storageMb.Data),
+		) != nil {
+			return err
+		}
```
begmaroman, line 159: This is not going to work as expected since `err` was not assigned to the given call result.

core/node/events/stream.go
```diff
@@ -671,6 +672,54 @@ func (s *Stream) GetMiniblocks(
 		return nil, false, err
 	}
 
+	// if stream is external, we need to read from external storage
+	location, err := s.params.Storage.GetMediaStreamLocation(ctx, s.streamId)
```
bas-vk, line 676: Why is the logic to fetch block data put here and not in the storage layer that makes how data is stored transparent for the rest of the node logic?

For snapshots the caller passes the `omitSnapshot` bool as indication if it wants snapshot data or not. For media streams I _think_ that ReadMiniblocks will always want to return block data?

rahul-ramesh: The storage layer only handle the psql specific operations (like read miniblocks), this requires access to the external media stream interface (download chunk from s3 or aws), which is outside of the storage layer. The choice was to not contaminate local storage logic with external API code. I guess the most elegant way would be to create another middleware layer to handle this, so it's transparent to this code here, and can call both local and remote storage functions.

core/node/storage/migrations/000009_external_media.up.sql
rahul-ramesh: I now derive the part number from the miniblock number, given the assumption that they are sequential and consecutive per stream. It simplified the design a lot and removed this race condition altogether. I removed "next_part" from the table

PR #3986, opened 2025-08-29
Add version header to clients

rahul-ramesh:
### Description
Add a version header to the sdk client, to track on the node side

### Changes
Include version header to the sdk client, from the package json file

### Checklist


packages/sdk/src/makeAuthenticationRpcClient.ts
```diff
@@ -31,6 +32,7 @@ export function makeAuthenticationRpcClient(
         baseUrl: url,
         interceptors: [
             ...(opts?.interceptors ?? []),
+            setHeaderInterceptor({ Version: packageJson.version }),
```
texuf, line 35: this looks great. Now that i see this we probably want a second header which is client_id, where the different apps can optionally identify themselves for metrics purposes. that is a different task though.

PR #3995, opened 2025-09-01
subgraph: denormalized tipping/swap analytics events

erikolsson:
This PR replaces #3989 and #3988. 
This PR adds a new table, `analytics_events` in the subgraph, as well as a few new properties on `space`.
- swapVolume, tipVolume, joinVolume, memberCount, all properties follow the same pattern: total, 24h, 7d, 30d.
Among other things, these new props unlock the following:
- "trending" spaces (sort by most joins in 24h, 7d, 30d)
- "top grossing" spaces
- "top tipping" spaces
- "top swapping" spaces
- activity feed of all activity across all spaces

### DIFF
--- a/packages/subgraph/ponder.schema.ts
+++ b/packages/subgraph/ponder.schema.ts
@@ -88,7 +88,7 @@ export const analyticsEvent = onchainTable(
         ethAmount: t.bigint().default(0n),
 
         // Event-specific data stored as typed JSON
-        eventData: t.json().notNull(),
+        eventData: t.json().$type<AnalyticsEventData>().notNull(),
     }),
     (table) => ({
         pk: primaryKey({ columns: [table.txHash, table.logIndex] }),


packages/subgraph/src/utils.ts
```diff
@@ -106,4 +106,116 @@ export function decodePermissions(permissions: readonly string[]): Permission[]
...
+    let joinVolume24h = 0n
+    let joinVolume7d = 0n
+    let joinVolume30d = 0n
+    let joinVolume = 0n
+    let memberCount24h = 0n
+    let memberCount7d = 0n
+    let memberCount30d = 0n
+    let memberCount = 0n
+
+    for (const event of allEvents) {
```
jterzis, line 152: I believe the api we want here is for this side effect to kick off on the event type that has fired in the analytic table, ie tip, swap, etc, rather than recalculating all metrics for every event that is added. Can we update this function to take an event type and only update metrics for spaceId, eventType ?

packages/subgraph/src/index.ts
```diff
@@ -193,23 +196,46 @@ ponder.on('Space:SwapExecuted', async ({ event, context }) => {
...
+            blockTimestamp: blockTimestamp,
+            createdAt: blockNumber,
         })
-        if (!existing) {
-            await context.db.insert(schema.swap).values({
-                txHash: transactionHash,
-                spaceId: spaceId,
-                recipient: event.args.recipient,
+
+        await context.db.insert(schema.analyticsEvent).values({
```
jterzis, line 221: Though this is in effect an "append-only" table and design, we are dealing with postgres and if a pk constraint fires here, it will possibly crash the indexer process so we should perform some conflict resolution before inserting.

packages/subgraph/src/index.ts
```diff
@@ -193,23 +196,46 @@ ponder.on('Space:SwapExecuted', async ({ event, context }) => {
     }
 
     try {
-        const existing = await context.db.sql.query.swap.findFirst({
```
jterzis, line 196: Why are we removing this ?

packages/subgraph/src/index.ts
```diff
@@ -193,23 +196,46 @@ ponder.on('Space:SwapExecuted', async ({ event, context }) => {
...
+            recipient: event.args.recipient,
+            tokenIn: event.args.tokenIn,
+            tokenOut: event.args.tokenOut,
+            amountIn: event.args.amountIn,
+            amountOut: event.args.amountOut,
+            poster: event.args.poster,
+            blockTimestamp: blockTimestamp,
+            createdAt: blockNumber,
         })
-        if (!existing) {
```
jterzis, line 199: Before stopping insertion to swap table, we need to ensure all clients consuming data from this table have migrated to the new format or are not yet consuming this table. Maybe we deprecate the swap table in a subsequent pr after figuring that out.

packages/subgraph/ponder.schema.ts
```diff
@@ -26,6 +42,24 @@ export const swap = onchainTable('swaps', (t) => ({
...
+    txHash: t.hex(),
+
+    // ETH value for the event (calculated field for sorting/aggregation)
+    ethAmount: t.bigint().default(0n),
+
+    // Event-specific data stored as JSON
+    // For swap: { tokenIn, tokenOut, amountIn, amountOut, recipient, poster }
+    // For tip: { sender, receiver, currency, amount, tokenId, messageId, channelId }
+    // for join: { recipient, tokenId }
+    eventData: t.json(),
```
jterzis, line 60: Maybe we should type this a bit further that way we can easily add new events with some type safety in the future.

```
  type SwapEventData = {
      type: 'swap'
      tokenIn: string
      tokenOut: string
      amountIn: string
      amountOut: string
      recipient: string
      poster: string
  }

  type TipEventData = {
      type: 'tip'
      sender: string
      receiver: string
      currency: string
      amount: string
      tokenId: string
      messageId: string
      channelId: string
  }

  type JoinEventData = {
      type: 'join'
      recipient: string
      tokenId: string
  }

  type AnalyticsEventData = SwapEventData | TipEventData | JoinEventData

  // In the table:
  eventData: t.json<AnalyticsEventData>().notNull(),
```

packages/subgraph/ponder.schema.ts
```diff
@@ -26,6 +42,24 @@ export const swap = onchainTable('swaps', (t) => ({
     createdAt: t.bigint(),
 }))
 
+// Denormalized events table for all analytics
+export const analyticsEvent = onchainTable('analytics_events', (t) => ({
+    id: t.text().primaryKey(), // ${txHash}-${logIndex}
+    spaceId: t.hex(),
+    eventType: t.text(), // 'swap', 'tip', etc.
```
jterzis, line 49: I believe we can use an enum here see https://ponder.sh/docs/schema/tables#enums

```
  import { onchainTable, onchainEnum, relations } from 'ponder'

  export const analyticsEventType = onchainEnum('analytics_event_type', [
      'swap',
      'tip',
      'join'
  ])

  eventType: analyticsEventType().notNull(),
```

PR #4003, opened 2025-09-02
PeerNodeRequestsWithRetires: Stop retrying on non-retryable errors

texuf:
If making a request that failed permissions checks, the request would get retried on all nodes. I don’t think this was ever intentional, seems like a waste of resources, and makes test results look weird.


texuf:
this is green now, but it was red before with the same code. Closing so @sergekh2 can work on it later.

PR #4007, opened 2025-09-02
sdk Update isTest helper to pick up vitest

texuf:
this is used to retry some on chain transactions in tests


miguel-nascimento (approved):
neat, thank you!

PR #4019, opened 2025-09-03
Copy app registry address to place where it’s referenced from code

texuf:


texuf:
this would have gotten picked up in the next publish to npm action

PR #4028, opened 2025-09-04
subgraph: add bayesian average rating

erikolsson:
This PR adds a `weightedRating` on spaces. To use the average rating for something concrete, we should use a weightedRating using something like a bayesian average.

- Spaces with 1 review at 5 stars → weightedRating ≈ 3.18
- Spaces with 5 reviews at 5 stars → weightedRating ≈ 3.6
- Spaces with 10 reviews at 5 stars → weightedRating ≈ 4
- Spaces with 50 reviews at 5 stars → weightedRating ≈ 4.67
- Spaces with many reviews converge to their true average

This is not for display purposes — this is to make an educated guess on which spaces are truly rated higher.

this is not truly 100% nitpicky exactly a bayesian average but it's pretty good. for this purpose, it doesn't really matter what the real globalAverage is...

stacked on #4027

### DIFF
--- a/packages/subgraph/src/utils.ts
+++ b/packages/subgraph/src/utils.ts
@@ -290,10 +290,7 @@ export async function updateSpaceReviewMetrics(
         const averageRating = allReviews.length > 0 ? totalRating / allReviews.length : 0
 
         // Calculate weighted rating using Bayesian average
-        const weightedRating = calculateWeightedRating(
-            averageRating,
-            allReviews.length,
-        )
+        const weightedRating = calculateWeightedRating(averageRating, allReviews.length)
 
         // Update space with calculated metrics
         await context.db.sql


packages/subgraph/src/utils.ts
```diff
@@ -258,6 +258,18 @@ export async function updateSpaceCachedMetrics(
...
 }
 
+export function calculateWeightedRating(averageRating: number, reviewCount: number): number {
+    // Bayesian average formula: pulls rating toward global average when few reviews exist
+    // This prevents spaces with 1-2 five-star reviews from ranking highest
+    const globalAverage = 3.0 // Middle rating on 1-5 scale
+    const minReviews = 10 // Threshold for confidence
+
+    const weightedRating =
+        (reviewCount * averageRating + minReviews * globalAverage) / (reviewCount + minReviews)
```
texuf, line 268: idk this is kinda lame if my town has 100 5 star reviews it's going to render as 4.81?

texuf (approved):

PR #4048, opened 2025-09-08
bump go-ethereum to v1.16.3 + disable log indexing in simulated backend

bas-vk:
Bump go-ethereum dependency to 1.16.3.

There is a problem in the simulated backend that was introduced in 1.15.6 when it's run with very short block period. The log indexing service can't keep up causing a deadlock through some of the api calls to fetch log data. The result is that some large unit tests time out. The simulated backend can be configured to run with log indexing disabled. There is no noticeable performance drop in unit tests with log indexing turned of.


core/node/crypto/testutil.go
```diff
@@ -99,7 +101,14 @@ func initSimulated(ctx context.Context, numKeys int) ([]*Wallet, *simulated.Back
 		genesisAlloc[wallets[i].Address] = types.Account{Balance: Eth_100}
 	}
 
-	backend := simulated.NewBackend(genesisAlloc, simulated.WithBlockGasLimit(30_000_000))
+	// With very short block period the eth node log indexing service can't keep up indexing logs.
+	// This can cause a deadlock when retrieving logs through the api. Disabled the indexing service,
+	// this hasn't got a noticeable performance impact in tests.
+	disableLogIndexing := func(nodeConf *node.Config, ethConf *ethconfig.Config) {
```
begmaroman, line 107: Should the PR title be changed to briefly mention another change apart from bumping the package version?

PR #4049, opened 2025-09-08
update bot slash commands using `towns-bot-cli`

miguel-nascimento:
This PR adds a typed API on top of slash commands & a command in `towns-bot-cli` that updates the bot slash commands, by reading commands a in `command.ts` file.

```t
// src/commands.ts
import type { PlainMessage, SlashCommand } from '@towns-protocol/proto'

const commands = [
    {
        name: 'help',
        description: 'Get help with bot commands',
    },
    {
        name: 'time',
        description: 'Get the current time',
    },
] as const satisfies PlainMessage<SlashCommand>[]

export default commands
```
image

If the developer wants to add a new command, they can append a new entry in the `commands` array, hit `towns-bot-cli update-slash-commands ./src/commands.ts <BEARER_TOKEN>`, and the cli will update the app registry server with the new command list.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -14581,7 +14581,7 @@ __metadata:
   languageName: node
   linkType: hard
 
-"dotenv@npm:^16.4.7":
+"dotenv@npm:^16.4.5, dotenv@npm:^16.4.7":
   version: 16.6.1
   resolution: "dotenv@npm:16.6.1"
   checksum: e8bd63c9a37f57934f7938a9cf35de698097fadf980cb6edb61d33b3e424ceccfe4d10f37130b904a973b9038627c2646a3365a904b4406514ea94d7f1816b69


packages/towns-bot-cli/src/index.ts
```diff
@@ -49,29 +61,52 @@ ${yellow('Usage:')}
...
-  ${cyan('# Update dependencies in current project')}
+  ${cyan('# Update dependencies')}
   towns-bot update
+
+  ${cyan('# List slash commands')}
+  towns-bot list-commands src/commands.ts
+  towns-bot list-commands --file src/commands.ts
+
+  ${cyan('# Update slash commands (positional arguments)')}
+  towns-bot update-commands commands.ts token123
```
miguel-nascimento, line 103: Later, we could be fancy and have a `towns-bot auth` that stores the bearer token in keychain, and then we can remove the bearer token argument here

PR #4078, opened 2025-09-10
move bot quickstart to bun

miguel-nascimento:
The wise beaver bot on alpha is running using the same base as here - working flawless until now.
It's easier, as it removes the esbuild and/or tsx dependency
We could in theory even [compile this to a single file executable](https://bun.com/docs/bundler/executables)

It makes dev experience better, watch + run with esbuild is painful


texuf (changes requested):
there are no tests....

PR #4079, opened 2025-09-10
[bot-framework] add `createdAt` in base payload

miguel-nascimento:
It was missing before. We can finally make a ping command


texuf (commented):
i think you should probably pass the whole event... or at least give the user a way to get it

PR #4129, opened 2025-09-16
feat(bots): ban/unban support

miguel-nascimento:


texuf (approved):

packages/bot/src/bot.ts
```diff
@@ -1018,6 +1038,26 @@ const buildBotActions = (client: ClientV2, viemClient: ViemClient, spaceDapp: Sp
...
+
+    /**
+     * Unban a user from a space
+     * Requires Permission.ModifyBanning to execute this action
+     */
+    const unban = async (userId: string, spaceId: string) => {
+        const tx = await spaceDapp.unbanWalletAddress(spaceId, userId, client.wallet)
+        const receipt = await tx.wait()
+        return { txHash: receipt.transactionHash }
+    }
```
miguel-nascimento, line 1059: note that we're using the client.wallet to pay the gas

maybe in a v2 or something we can expose only the tx and allow the user to pass the desired wallet to pay

PR #4137, opened 2025-09-16
Remove one-off account abstraction mapping in contracts

texuf:
i’m having build problems in harmony and trying to debug what’s going on


texuf:
not sure why this fails - perhaps the trailing slash? not going to go deeper.

shuhuiluo:
> not sure why this fails - perhaps the trailing slash? not going to go deeper.

In @erc6900/reference-implementation/interfaces/IValidationHookModule.sol, it uses
```solidity
import {PackedUserOperation} from "@eth-infinitism/account-abstraction/interfaces/PackedUserOperation.sol";
```
In SubscriptionModuleFacet, we have
```solidity
import {PackedUserOperation} from "account-abstraction/interfaces/PackedUserOperation.sol";
```
We can consolidate by only referring to @eth-infinitism/account-abstraction.

PR #4153, opened 2025-09-17
envUtils: Enable passing import.meta.env all the way down from the browser

texuf:
super lame worker code just doesn’t compile so we need to work around it.


texuf:
I'm really unhappy with this, will fix tomorrow after fixing the builds.

PR #4158, opened 2025-09-18
Remove shared syncer from the codebase

begmaroman:
- Removed shared syncer
- Added some unit tests in eventbus package of sync v3


core/node/rpc/syncv3/eventbus/stream_subscribers_test.go
```diff
@@ -0,0 +1,279 @@
+package eventbus
```
bas-vk, line 1: This PR removes the shared syncer, why does it add `syncv3/eventbus` files?

PR #4194, opened 2025-09-23
app-registry: listen app inbox to flush enqueued messages

miguel-nascimento:
Not sure if that's the root issue that was causing bots to be stuck in alpha in some channels / spaces, but locally it fixes it. Would like to give a try and deploy to `alpha`

Claude Code helped me to build this PR.

### DIFF
--- a/core/node/app_registry/sync/tracked_stream.go
+++ b/core/node/app_registry/sync/tracked_stream.go
@@ -19,6 +19,7 @@ type AppRegistryTrackedStreamView struct {
 	TrackedStreamViewImpl
 	listener track_streams.StreamEventListener
 	queue    EncryptedMessageQueue
+	tracker  track_streams.StreamsTracker
 }
 
 func (b *AppRegistryTrackedStreamView) processUserInboxMessage(ctx context.Context, event *ParsedEvent) error {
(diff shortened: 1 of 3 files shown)


core/node/app_registry/sync/tracked_stream.go
```diff
@@ -64,6 +65,7 @@ func (b *AppRegistryTrackedStreamView) onNewEvent(ctx context.Context, view *Str
 		return err
 	}
 	apps := mapset.NewSet[string]()
```
bas-vk, line 67: Streams of type `STREAM_USER_INBOX_BIN` are added to the tracked set of streams in this PR. There is a special check on line 54/55 if the tracked stream is of this type. If so calls `processUserInboxMessage` and returns. Any logic after this check isn't executed.

Is it not enough to just start tracking these user inbox streams?

miguel-nascimento: Thanks the comment. I decided to remove the code that adds `STREAM_USER_INBOX_BIN` to the set of tracked streams, now I'm only adding app user inbox streams to be tracked. Previously, all user inbox streams would be tracked if I got it right

PR #4197, opened 2025-09-24
Merge stream and snapshot trimmers into a single worker

begmaroman:


core/node/storage/pg_stream_trimmer.go
```diff
@@ -38,25 +47,33 @@ type streamTrimmer struct {
...
-		miniblocksToKeep:  miniblocksToKeep,
+		config:            config,
 		trimmingBatchSize: trimmingBatchSize,
 		workerPool:        workerPool,
 		pendingTasks:      make(map[StreamId]struct{}),
 		stop:              make(chan struct{}),
+		taskDuration: metrics.NewHistogramEx(
+			"stream_trimmer_task_duration_seconds",
+			"Stream trimmer task duration in seconds",
+			[]float64{.05, .1, .5, 1.0, 1.5, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0, 8.0},
```
bas-vk, line 75: I think its fine to make this less granular, something like `[0.5,1,2,4,6,8]`.

PR #4235, opened 2025-09-30
Make default image load default env config based on TOWNS_ENV var name

sergekh2:
### Description

If TOWNS_ENV is not set, default to `omega`.
Then load contract addresses, chain ids and public rpc provider from matching config.
Value set by env vars are going to overwrite these values from default config.
In particular, RPC providers should be set.


texuf:
Pretty cool. We can test by removing our argo cd params

PR #4241, opened 2025-10-01
chore: fix phantom dependencies and upgrade packages

shuhuiluo:
Add previously undeclared dependencies that were being used transitively, and upgrade several packages to their latest compatible versions.


texuf (approved):

packages/rpc-connector/package.json
```diff
@@ -57,14 +59,6 @@
...
         "@connectrpc/connect-web": ">=2.0.0"
     },
-    "peerDependenciesMeta": {
-        "@connectrpc/connect-node": {
-            "optional": true
-        },
-        "@connectrpc/connect-web": {
-            "optional": true
-        }
-    },
```
miguel-nascimento, line 67: Can you keep this? 
If we remove, then pnpm users will have a warning talking about missing peer dependency 

e.g: if they already have `@connectrpc/connect-web` installed, it will warn about missing `@connectrpc/connect-node`, and we want users to install the connectrpc library to match their env

shuhuiluo: Yes.

sergekh2 (approved):

PR #4250, opened 2025-10-02
Add beta env to towns

Tee-S:
### Description

Adding support for a new testnet environment called `beta`

### Changes

- Add boilerplate files needed to support the new Beta env (such as contract addresses and URLs, etc.)
- update the github workflows to handle the new beta env


core/env/beta/config.yaml
```diff
@@ -0,0 +1,17 @@
...
+    blockTimeMs: 2000
+riverChain:
+    chainId: 6524490
+    networkUrl: 'https://testnet.rpc.towns.com'
+    blockTimeMs: 2000
+
+# Base contracts
+architectContract:
+    address: '0x5A077e686545c541476D3a99711dE42339357467'
+
```
texuf, line 14: We need two more contract addresses here i'm pretty sure

.github/workflows/Bytecode_diff_report.yml
shuhuiluo, line 1: This workflow has been down. It would be great to turn it back on.
https://gamma-reports.river.build/

core/env/beta/config.yaml
Tee-S: yeah good catch

PR #4256, opened 2025-10-03
TOWNS-31164 Bot username validation - do not allow space

jontra:
### Description

Added a bunch of username validation for bots:
- Must start with a letter or number
- Can contain letters (a-z, A-Z), numbers (0-9), underscores (_), and hyphens (-)
- Must be between 1 and 255 characters
- No spaces allowed

### Checklist

### DIFF
--- a/core/node/app_registry/types/app_metadata.go
+++ b/core/node/app_registry/types/app_metadata.go
@@ -257,7 +257,10 @@ func validateSlashCommandName(name string) error {
 	// Check if name contains only letters, numbers, and underscores
 	for i, ch := range name {
 		if !((ch >= 'a' && ch <= 'z') || (ch >= 'A' && ch <= 'Z') || (ch >= '0' && ch <= '9') || ch == '_') {
-			return RiverError(protocol.Err_INVALID_ARGUMENT, "command name must contain only letters, numbers, and underscores").
+			return RiverError(
+				protocol.Err_INVALID_ARGUMENT,
+				"command name must contain only letters, numbers, and underscores",
+			).
 				Tag("name", name).
 				Tag("invalidCharAt", i)
 		}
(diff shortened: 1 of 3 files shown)


core/node/app_registry/types/app_metadata.go
```diff
@@ -160,8 +176,10 @@ func ValidateAppMetadata(metadata *protocol.AppMetadata) error {
 		return base.RiverError(protocol.Err_INVALID_ARGUMENT, "metadata is required")
 	}
 
-	if metadata.GetUsername() == "" {
-		return base.RiverError(protocol.Err_INVALID_ARGUMENT, "metadata username is required")
+	// Validate username format
+	if err := ValidateBotUsername(metadata.GetUsername()); err != nil {
+		return base.RiverErrorWithBase(protocol.Err_INVALID_ARGUMENT, "metadata username validation failed", err).
```
bas-vk, line 181: Because the `base` package is such a common package that we use throughout the code base its usually imported with a dot so it won't require prefixing its logic with `base.` Same for the protocol package.

core/node/app_registry/types/app_metadata.go
```diff
@@ -12,6 +13,21 @@ const MAX_RESOURCE_URL_LENGTH = 8192
 
 const MAX_SLASH_COMMANDS = 25
 
+// usernameRegex validates bot usernames:
+// - Must start with a letter or number
+// - Can contain letters (a-z, A-Z), numbers (0-9), underscores (_), and hyphens (-)
+// - Must be between 1 and 255 characters
+// - No spaces allowed
+var usernameRegex = regexp.MustCompile(`^[a-zA-Z0-9][a-zA-Z0-9_-]{0,255}$`)
```
bas-vk, line 21: I _think_ this allows usernames that are 256 characters long?
Can you add a test that checks the length with a username of 255 and 256 chars long?

jontra: you are correct, my comment is a mistake.

jontra: added a test

PR #4267, opened 2025-10-07
node: add support for S3/GCP storage to upload media stream chunks

bas-vk:
### Introduction

This is a first PR to add support for storing media stream chunks in S3 compatible storage.

## Summary

  - Add S3 configuration helpers and wire up AWS SDK dependencies so media miniblocks
  can be offloaded to S3-compatible storage.
  - Extend protocol and storage descriptors with chunk metadata required for multipart
  uploads, including JSON persistence for stream metadata.
  - Introduce an S3 stream store with integration tests covering single PUT and
  multipart upload flows against AWS or GCP credentials.

## Key Changes

  - core/config/config.go: define S3Config plus env-based helpers for AWS/GCP test
  credentials.
  - core/go.mod, core/go.sum: pull in aws-sdk-go-v2 modules needed for S3 operations.
  - protocol/protocol.proto: add chunk_count and chunk_size to MediaPayload.Chunk.
  - core/node/storage/storage.go & pg_ephemeral_store.go: record media miniblock
  metadata, implement JSON (un)marshalling, and expose helpers for upload completion.
  - New core/node/storage/s3_stream_store.go & _test.go: implement S3 write/read/delete
  logic and integration tests for both single-part and multipart uploads.

## Testing

  1. go test ./core/node/storage -run TestS3Integration (requires S3-compatible
  credentials via env vars)


core/node/storage/s3_stream_store.go
```diff
@@ -0,0 +1,321 @@
...
+	}
+
+	if miniblock.MediaStream.ChunkSize <= 0 {
+		// backwards compatability, a client didn't provide required information to store miniblocks in S3.
+		return false, nil
+	}
+
+	// the last chunk can have a different size than the given chunk_size.
+	if isFinalChunk := (miniblock.MediaStream.ChunkIndex + 1) == miniblock.MediaStream.ChunkCount; isFinalChunk {
+		return miniblock.MediaStream.ChunkSize < (2 * S3MultiPartMinimumPartSize), nil
```
jontra, line 65: I don't understand the `2 * S3MultiPartMinimumPartSize` part. why does the chunk size needs to be less than twice the minimum part size?
Won't we want the data size to be exactly chunk size except for the final block?

jontra: Nitpick: spaces on `2 * ...` and below without spaces

jontra: does it make sense to add some validation to `len(miniblock.Data)` also for the final block?

core/node/storage/s3_stream_store.go
```diff
@@ -0,0 +1,321 @@
...
+		Err_BAD_BLOCK, "Media miniblock data size does not match chunk size",
+		"stream", streamID,
+		"chunkSize", miniblock.MediaStream.ChunkSize,
+		"blockData", len(miniblock.Data),
+		"miniblock", miniblock.Number).
+		Func("mustStoreEphemeralMiniblockInS3")
+}
+
+// S3DeleteMediaMiniblock deletes the given miniblock from S3.
+// It's intended use if for tests and therefore accepts t.
```
jontra, line 84: s/use if for/use is for/

core/node/storage/s3_stream_store.go
```diff
@@ -0,0 +1,321 @@
...
+func S3DeleteMediaMiniblock(
+	t *testing.T,
```
jontra, line 86: this creates a dependency to the testing library in production code. maybe we can find better way to do it

core/node/storage/s3_stream_store.go
```diff
@@ -0,0 +1,321 @@
...
+	client *s3.Client,
+	bucket string,
+	streamID StreamId,
+	node common.Address,
+	streamMD *StreamMetaData,
+) ([]*MiniblockDescriptor, error) {
+	if streamID.Type() != STREAM_MEDIA_BIN {
+		return nil, RiverError(Err_BAD_STREAM_ID, "Unable to read miniblocks from non media stream").
+			Tag("stream", streamID).
+			Func("ReadMiniblocksFromS3Object")
```
jontra, line 260: wrong function name

core/node/storage/s3_stream_store.go
```diff
@@ -0,0 +1,321 @@
...
+
+	objectKey := externalStorageObjectKey(node, streamID)
+	getObjectResult, err := client.GetObject(ctx, &s3.GetObjectInput{
+		Bucket: &bucket,
+		Key:    &objectKey,
+	})
+	if err != nil {
+		return nil, RiverErrorWithBase(Err_DOWNSTREAM_NETWORK_ERROR, "Unable to read S3 object", err).
+			Tags("stream", streamID).
+			Func("readMiniblocksFromS3/GetObject")
```
jontra, line 283: wrong function name

core/node/storage/s3_stream_store_test.go
```diff
@@ -0,0 +1,316 @@
...
+	"github.com/towns-protocol/towns/core/config/builder"
+
+	. "github.com/towns-protocol/towns/core/node/base"
+	. "github.com/towns-protocol/towns/core/node/protocol"
+	. "github.com/towns-protocol/towns/core/node/shared"
+	. "github.com/towns-protocol/towns/core/node/storage"
+	"github.com/towns-protocol/towns/core/node/testutils"
+)
+
+func TestS3Integration(t *testing.T) {
```
jontra, line 24: I think using real aws service for testing is not a good practice.
We need at the very least mock the service, to allow the test to run even without aws configuration, and also emulate some s3 errors and failures (which I don't see here).

We can also use LocalStack (or other s3 local implementations). I think this is the best option in the long term

PR #4277, opened 2025-10-09
Stream trimming during reconciliation 

begmaroman:
Stream trimming during reconciliation

  - Reworked reconciliation to respect history windows, reject unexpected gaps, and retain snapshots with tighter assertions.
  - Added comprehensive tests that verify trimmed ranges, snapshot behavior, and untouched streams in failure scenarios.

  Stream trimming during node runtime

  - Introduced activation-factor scheduling with modulo snapshot counters to throttle trimming jobs safely.
  - Hardened storage logic (MinRetention clamp, MaxInt64 guard) and added targeted unit tests for the scheduler.
  - Trimming now runs every StreamTrimActivationFactor snapshots per stream, with counters reset after scheduling and skipped when history/retention policies disable trimming.

### DIFF
--- a/packages/contracts/scripts/deployments/facets/DeployTipping.s.sol
+++ b/packages/contracts/scripts/deployments/facets/DeployTipping.s.sol
@@ -7,15 +7,26 @@ import {ITipping} from "src/spaces/facets/tipping/ITipping.sol";
 
 // libraries
 import {LibDeploy} from "@towns-protocol/diamond/src/utils/LibDeploy.sol";
+import {DynamicArrayLib} from "solady/utils/DynamicArrayLib.sol";
 
 library DeployTipping {
+    using DynamicArrayLib for DynamicArrayLib.DynamicArray;
+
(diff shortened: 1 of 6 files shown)


core/node/storage/pg_stream_store.go
```diff
@@ -2635,6 +2621,89 @@ func (s *PostgresStreamStore) getMiniblockNumberRangesTx(
...
+			"startMbExclusively must be non-negative",
+			"streamId", streamId,
+			"startMbExclusively", startMbExclusively,
+		).Func("PostgresStreamStore.trimStreamTx")
+	}
+
+	if startMbExclusively > 1 {
+		query := s.sqlForStream(
+			`DELETE FROM {{miniblocks}}
+		WHERE stream_id = $1 AND seq_num >= 1 AND seq_num < $2`,
```
sergekh2: why not to trim 0?

sergekh2 (approved):

PR #4301, opened 2025-10-14
node: add support to store media stream miniblock externally

bas-vk:
Summary

  - Add external media storage configuration and wiring in core/config/config.go:88 and core/node/rpc/server.go:647, enabling either AWS S3 or GCS.
  - Introduce the external storage layer plus schema support for offloading media miniblocks, including migrations and helpers in core/node/storage/pg_external_storage.go:18 and core/node/storage/migrations/000009_media_miniblock_ext_storage_tables.up.sql.
  - Integrate migration and read-path logic so normalized media streams move to external buckets and hydrate back transparently (core/node/storage/pg_stream_store.go:90, core/node/storage/pg_ephemeral_store_monitor.go:18, core/node/storage/
    pg_stream_trimmer.go:201).
  - Add end-to-end AWS/GCS tests and update test helpers to inject external storage clients (core/node/storage/
    pg_stream_store_external_storage_test.go:29, core/node/storage/test_pg_stream_store.go:18, core/node/events/
    stream_storage_test.go:21).

Workflow

  - Normalizing an ephemeral media stream verifies all chunks, flips the es row to non-ephemeral, and
    calls esm.onSealed, which drops the ID from the live map and enqueues it for migration (core/node/
    storage/pg_ephemeral_store.go:307, core/node/storage/pg_ephemeral_store.go:324, core/node/storage/
    pg_ephemeral_store_monitor.go:83).
  - The ephemeral monitor runs a dedicated migration loop whenever external storage is configured, polling the queued
    IDs every few seconds and retrying failures up to five times (core/node/storage/pg_ephemeral_store_monitor.go:63,
    core/node/storage/pg_ephemeral_store_monitor.go:91, core/node/storage/pg_ephemeral_store_monitor.go:99).
  - Each queued stream invokes MigrateMiniblocksToExternalStorage, which first confirms the stream is fully
    normalized, then streams all miniblock payloads into a combined buffer while recording byte offsets (core/
    node/storage/pg_ephemeral_store_monitor.go:122, core/node/storage/pg_external_storage.go:97, core/node/storage/
    pg_external_storage.go:127).
  - The combined object is written to the configured bucket (S3 or GCS), metadata about each miniblock
    segment is copied into miniblocks_ext_storage_*, and the local blockdata column is nulled while marking
    es.blockdata_ext with the backend code (core/node/storage/pg_external_storage.go:151, core/node/storage/
    pg_external_storage.go:182, core/node/storage/pg_external_storage.go:235).
  - Subsequent reads detect the external storage flag, load the part metadata, fetch the object from the bucket, and
    rebuild the miniblock descriptors before returning them to callers (core/node/storage/pg_stream_store.go:1407,
    core/node/storage/pg_stream_store.go:1477, core/node/storage/pg_stream_store.go:1501).

Follow up PR includes support to migrate miniblock data for existing media streams to external storage.

### DIFF
--- a/core/node/storage/pg_ephemeral_store_monitor.go
+++ b/core/node/storage/pg_ephemeral_store_monitor.go
@@ -247,6 +247,7 @@ func (m *ephemeralStreamMonitor) handleStream(ctx context.Context, streamId Stre
 									 DELETE from {{minipools}} WHERE stream_id = $1;
 									 DELETE FROM es WHERE stream_id = $1`,
 					streamId,
+					MiniblockDataStorageLocationDB,
 				),
 				streamId,
 			)
--- a/core/node/storage/pg_external_storage.go
+++ b/core/node/storage/pg_external_storage.go
@@ -782,10 +782,10 @@ func (s *PostgresStreamStore) writeMediaStreamExternalStoragePartsTx(
 			Func("writeMediaStreamExternalStoragePartsTx")
 	}
 
-	// drop miniblock data from DB now the miniblock data is uploaded to external storage
-	q := s.sqlForStream(`UPDATE {{miniblocks}} SET blockdata = NULL WHERE stream_id = $1;`, streamID)
+	// drop miniblock records from miniblock regular now the miniblock data is uploaded to external storage
+	q := s.sqlForStream(`DELETE FROM {{miniblocks}} WHERE stream_id = $1;`, streamID, location)
 	// update the stream record with the new external storage location
-	q += s.sqlForStream(`UPDATE es SET blockdata_ext = $2 WHERE stream_id = $1;`, streamID)
+	q += s.sqlForStream(`UPDATE es SET blockdata_ext = $2 WHERE stream_id = $1;`, streamID, location)
 
 	if _, err := tx.Exec(ctx, q, streamID, extStorageLoc); err != nil {
 		return RiverErrorWithBase(
@@ -797,10 +797,10 @@ func (s *PostgresStreamStore) writeMediaStreamExternalStoragePartsTx(
 			Func("writeMediaStreamExternalStoragePartsTx")
 	}
 
-	// import parts in DB
+	// import parts in miniblocks external storage table
 	if _, err := tx.CopyFrom(
 		ctx,
-		pgx.Identifier{s.sqlForStream("{{miniblocks_ext_storage}}", streamID)},
+		pgx.Identifier{s.sqlForStream("{{miniblocks}}", streamID, extStorageLoc)},
 		[]string{"stream_id", "seq_num", "start_byte", "size"},
 		pgx.CopyFromSlice(len(parts), func(i int) ([]any, error) {
 			part := parts[i]
@@ -821,10 +821,12 @@ func (s *PostgresStreamStore) readMediaStreamExternalStoragePartsTx(
 	ctx context.Context,
 	tx pgx.Tx,
 	streamID StreamId,
+	miniblockLocation MiniblockDataStorageLocation,
 ) ([]ExternallyStoredMiniblockDescriptor, error) {
 	query := s.sqlForStream(
-		`SELECT seq_num, start_byte, size FROM {{miniblocks_ext_storage}} WHERE stream_id = $1 ORDER BY seq_num;`,
+		`SELECT seq_num, start_byte, size FROM {{miniblocks}} WHERE stream_id = $1 ORDER BY seq_num;`,
 		streamID,
+		miniblockLocation,
 	)
 
 	partsRows, err := tx.Query(ctx, query, streamID)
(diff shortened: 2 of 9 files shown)


core/node/storage/pg_ephemeral_store_monitor.go
```diff
@@ -20,6 +20,10 @@ type ephemeralStreamMonitor struct {
 	// streams is a map of ephemeral stream IDs to the creation time.
 	// This is used by the monitor to detect "dead" ephemeral streams and delete them.
 	streams *xsync.Map[StreamId, time.Time]
+	// streamsToMigrateToExternalStorage keeps a collection of ephemeral streams that
+	// need their miniblocks to be migrated to external storage.
+	streamsToMigrateToExternalStorage *xsync.Map[StreamId, int]
```
sergekh2: please add comment what int is for

core/node/storage/migrations/000009_media_miniblock_ext_storage_tables.up.sql
```diff
@@ -0,0 +1,34 @@
...
+            -- make the media stream blockdata column optional
+            EXECUTE 'ALTER TABLE miniblocks_m' || SUFFIX || ' ALTER COLUMN blockdata DROP NOT NULL;';
+
+            -- Tables that contain records when media stream miniblock data is stored in an external storage.
+            EXECUTE 'CREATE TABLE IF NOT EXISTS miniblocks_ext_storage_m' || suffix || ' (
+                stream_id CHAR(64) NOT NULL,
+                seq_num BIGINT NOT NULL,             -- miniblock number
+                start_byte BIGINT NOT NULL,          -- offset where miniblock starts in combined object
+                size BIGINT NOT NULL,                -- how big the blockdata is in bytes
+                bucket TEXT NOT NULL,                -- bucket name
```
begmaroman: How the DB structure is affected if we will introduce an alternative storage (something different from AWS S3)?

core/node/storage/pg_external_storage.go
```diff
@@ -0,0 +1,316 @@
...
+	// combine miniblocks into a single object and keep track of parts to decode individual miniblocks
+	// from single object later
+	var (
+		combinedMiniblocksData = new(bytes.Buffer)
+		miniblocksTotalLength  = uint64(0)
+		parts                  []externallyStoredMiniblockDescriptor
+	)
+
+	if err := s.ReadMiniblocksByStream(ctx, streamID, true, func(blockdata []byte, seqNum int64, snapshot []byte) error {
+		_, err := combinedMiniblocksData.Write(blockdata)
```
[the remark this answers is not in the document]
bas-vk: I have now introduced a 5MiB upload chunk to satisfy S3 constraints (interface ExternalStorageWriter implemented for gcs and s3).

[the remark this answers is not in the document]
bas-vk: With the switch to plain put operations the miniblocks are now streams to external storage.

core/node/storage/migrations/000009_media_miniblock_ext_storage_tables.up.sql
```diff
@@ -0,0 +1,33 @@
...
+
+    numPartitions INT;
+
+BEGIN
+SELECT num_partitions from settings where single_row_key=true into numPartitions;
+    FOR i IN 0.. numPartitions LOOP
+            suffix = LPAD(TO_HEX(i), 2, '0');
+
+            -- make the media stream blockdata column optional
+            EXECUTE 'ALTER TABLE miniblocks_m' || SUFFIX || ' ALTER COLUMN blockdata DROP NOT NULL;';
```
sergekh2: is there a need to do this? not clear why it's needed.

core/node/storage/pg_ephemeral_store_monitor.go
```diff
@@ -68,9 +94,96 @@ func (m *ephemeralStreamMonitor) onCreated(streamId StreamId) {
...
+
+// runStreamMigrationToExternalStorage migrates miniblocks of normalized ephemeral stream
+// to external storage.
+func (m *ephemeralStreamMonitor) runStreamMigrationToExternalStorage(ctx context.Context) {
+	log := logging.FromCtx(ctx)
+
+	log.Info("Enable media stream miniblock data in external storage",
+		"migrateExistingStreams", m.migrateMediaStreamsToExternalStorage)
+
+	go func() {
```
sergekh2: please make it a named method - it's long enough and code will be more readable

core/node/storage/pg_external_storage.go
```diff
@@ -0,0 +1,1045 @@
...
+	tx pgx.Tx,
+	streamID StreamId,
+	extStorageLoc MiniblockDataStorageLocation,
+	parts []ExternallyStoredMiniblockDescriptor,
+) error {
+	if _, _, err := s.lockStream(ctx, tx, streamID, true); err != nil {
+		return err
+	}
+
+	// delete any previous parts for this stream
```
sergekh2: before updates it should check that storage is not external already. (i.e. if function is called twice second time it should fail)

core/node/storage/pg_external_storage.go
```diff
@@ -0,0 +1,1045 @@
...
+	q := s.sqlForStream(`DELETE FROM {{miniblocks_ext_storage}} WHERE stream_id = $1;`, streamID)
```
sergekh2: it doesn't seem to be needed: it's transactional, so there should be no records to delete.

PR #4315, opened 2025-10-16
Removed enableNewSnapshotFormat feature flag from config

begmaroman:
The given feature flag is deprecated since the new snapshot format has been enabled for few month already. The given config can be safely removed from the codebase.


texuf (approved):
Is there a typescript sdk component for this one?

PR #4354, opened 2025-10-21
Update appService::getSessionKey to allow fetching a session key per stream

texuf:
When sending messages from the bot we want to reuse existing sessions, not create a new one every time. We don't require bots to have local storage, so lets add the ability for them to fetch a key instead of creating it.

### DIFF
--- a/core/node/rpc/app_registry_test.go
+++ b/core/node/rpc/app_registry_test.go
@@ -1717,14 +1717,18 @@ func TestAppRegistry_GetSession(t *testing.T) {
 	require.EventuallyWithT(func(c *assert.CollectT) {
 		req := &connect.Request[protocol.GetSessionRequest]{
 			Msg: &protocol.GetSessionRequest{
-				AppId:     botWallet.Address[:],
-				SessionId: &testSession1,
+				AppId: botWallet.Address[:],
+				Identifier: &protocol.GetSessionRequest_SessionId{
+					SessionId: testSession1,
+				},
 			},
 		}
 		req2 := &connect.Request[protocol.GetSessionRequest]{
 			Msg: &protocol.GetSessionRequest{
-				AppId:    botWallet.Address[:],
-				StreamId: channelId[:],
+				AppId: botWallet.Address[:],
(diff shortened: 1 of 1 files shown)


core/node/app_registry/service.go
```diff
@@ -407,14 +407,28 @@ func (s *Service) GetSession(
...
+	} else {
 		return nil, base.RiverError(
 			Err_INVALID_ARGUMENT,
-			"Invalid session id",
+			"Invalid session id or stream id",
 		).Tag("sessionId", req.Msg.SessionId).Tag("appId", app).Func("GetSession")
 	}
 
-	envelopeBytes, err := s.store.GetSessionKey(ctx, app, req.Msg.SessionId)
 	if err != nil {
```
jontra: This error handling should be removed as well, since you have this error handling on lines 413 and 422

core/node/app_registry/service.go
jontra: Fix `req.Msg.SessionId` is pointer

core/node/storage/pg_app_registry_store.go
```diff
@@ -762,6 +770,30 @@ func (s *PostgresAppRegistryStore) GetSessionKey(
...
 
+func (s *PostgresAppRegistryStore) GetSessionKeyForStream(
+	ctx context.Context,
+	app common.Address,
+	streamId shared.StreamId,
+) (encryptionEnvelope []byte, err error) {
+	err = s.txRunner(
+		ctx,
+		"GetSessionKeys",
+		pgx.ReadWrite,
```
jontra: use `pgx.ReadOnly` for select queries

protocol/apps.proto
```diff
@@ -146,7 +146,8 @@ message GetSessionRequest {
     // public key of the app
     bytes app_id = 1;
 
-    string session_id = 2;    
+    optional string session_id = 2;
```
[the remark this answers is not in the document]
texuf: I also agree, it just seemed like overkill here. If you feel strongly then yes.

jontra (approved):

PR #4374, opened 2025-10-22
Added an index on spaces table on subgraph setup hook

jontra:
### Description

Subgraph has an optimization to create the indexes only after the historic indexing is done.
We are reading from the table while we index, and if this index is not added before indexing starts, then postgres will do a full table scan for every SpaceOwner:Transfer event, which make it run slower the more spaces we index.
 
### Changes

Add the index on startup

### Checklist


erikolsson (approved):
Interesting, I didn't know about the 'setup' hook

PR #4376, opened 2025-10-22
bots: make `viem` a peer dependency

miguel-nascimento:
We depend on viem, but we want our users to use viem as well. In this case, we need to make `viem` a peer dependency.


shuhuiluo (approved):
Are you sure 2.29.3 is the minimum?

PR #4382, opened 2025-10-23
Bots: Change getStream to return streamStateView, fix key sharing

texuf:
Bots: Change getStream to return streamStateView, fix key sharing

### DIFF
--- a/packages/sdk/src/client-v2.ts
+++ b/packages/sdk/src/client-v2.ts
@@ -25,12 +25,12 @@ import {
     makeUserInboxStreamId,
     makeUserMetadataStreamId,
     streamIdAsBytes,
+    streamIdAsString,
     userIdFromAddress,
 } from './id'
 import {
     make_UserInboxPayload_GroupEncryptionSessions,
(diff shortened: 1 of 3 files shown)


packages/sdk/src/client-v2.ts
```diff
@@ -184,13 +185,26 @@ export const createTownsClient = async (
...
+    const getStream = async (streamId: string): Promise<StreamStateView> => {
         const { disableHashValidation, disableSignatureValidation } = client
         const stream = await client.rpc.getStream({ streamId: streamIdAsBytes(streamId) })
-        return unpackStream(stream.stream, {
+        const unpackedResponse = await unpackStream(stream.stream, {
             disableHashValidation,
             disableSignatureValidation,
         })
+        const streamView = new StreamStateView(userId, streamIdAsString(streamId), undefined)
+        streamView.initialize(
```
miguel-nascimento: we're creating a new stream view per `getStream(sameStream)` request
can we reuse it? (if defined)

PR #4403, opened 2025-10-26
fix: resolve Docker build failures with contract artifact generation

shuhuiluo:
### Description

Fixes Docker build failures caused by Foundry installation issues in Alpine Linux containers. The root cause was Alpine's musl libc being incompatible with Foundry's glibc-compiled binaries, preventing `prepare.js` from generating contract artifacts during builds.

This resulted in "generated artifacts mismatch" errors when Docker images were built with stale or missing contract typings.

### Changes

**Dockerfiles (stream-metadata, metrics-discovery, xchain-monitor):**
- Switch builder stage from Alpine to Ubuntu 24.04 (native glibc support for Foundry)
- Add Node.js 20 installation via NodeSource
- Add `yarn cache clean` to all yarn install commands
- Keep Alpine runtime stage for minimal final image size

**prepare.js:**
- Fix Docker detection: when git unavailable but contract source exists, force regeneration
- Fix fallback logic: properly attempt local generation when npm download fails
- Preserves npm package behavior (use existing artifacts when no source available)

**Impact:**
- ✅ Docker builds now successfully generate fresh contract artifacts
- ✅ Artifacts match current contract code (no more hash mismatches)
- ✅ prepare.js works in both local dev (with git) and Docker (without git)

### Checklist

### DIFF
--- a/packages/generated/scripts/prepare.js
+++ b/packages/generated/scripts/prepare.js
@@ -63,11 +63,19 @@ function getContractsHash() {
 // Check if contracts have changed by comparing hashes
 function contractsChanged() {
   const currentHash = getContractsHash();
-  
-  if (!currentHash) return false; // No contracts source, use existing generated files
-  
+
+  if (!currentHash) {
+    // Check if we're in Docker with source but no git (git rev-parse failed)
+    if (existsSync(resolve(contractsDir, 'src'))) {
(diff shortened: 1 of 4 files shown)


shuhuiluo:
@coderabbitai review

PR #4413, opened 2025-10-27
feat(bots): `getSmartAccountFromUserId`

miguel-nascimento:
Only getting modular accounts (new or upgraded) for now.
Not sure how many simple accounts are out there yet. Let me know if its a blocker.

Later, we want to move this to use WalletLink `getDefaultWallet/getActiveWallet` instead of checking linked wallets and finding one that is a modular smart account.

I'm now checking if `userId` contains a smart wallet address, if so, we send the tip to it.

One of the things that this PR enables is bot tipping any message in `onMessage` handlers
```ts
import { getSmartAccountFromUserId } from "@towns-protocol/bot"

bot.onMessage((handler, { channelId, eventId, userId, message }) => {
  const shouldTip = checkIfDeserveTip(message)
  if (shouldTip) {
    await handler.sendTip({
      userId,
      amount: parseEther('0.001')
      messageId: eventId,
      channelId
    })
  }
})
```

image

```ts
botfather.onSlashCommand(
  "getWallet",
  async (handler, { channelId, userId }) => {
    const wallet = await getDeployedSmartAccount(botfather, { userId });
    await handler.sendMessage(channelId, `Your wallet address is: ${wallet}`);
  }
);
```

### DIFF
--- a/packages/bot/src/bot.test.ts
+++ b/packages/bot/src/bot.test.ts
@@ -800,8 +800,7 @@ describe('Bot', { sequential: true }, () => {
 
         bot.onMessage(async (handler, event) => {
             const result = await handler.sendTip({
-                receiver: bob.userId,
-                receiverUserId: bob.userId,
+                userId: bob.userId,
                 amount: ethers.utils.parseUnits('0.005').toBigInt(),
                 messageId: event.eventId,
                 channelId: event.channelId,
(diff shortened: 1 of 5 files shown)


salzbrenner:
There is no definitive plan to make use of getDefaultWallet.

For now as far as towns app is concerned, the smart accounts is always going to be the receiving address for a tip

salzbrenner:
Re. "Not sure how many simple accounts are out there yet. Let me know if its a blocker." - I encountered one today. It's not a blocker per se - I'm not sure how many exist but they are still out there. There are things we could do in the client to force a user to upgrade to modular if they installed a bot. Otherwise I guess they just tip just won't be sent as long as there is a check for an address

salzbrenner:
Being unfamiliar with bot architecture - is there any way the receiver address could be passed in i.e. as something like a getAddress callback or something?
If not, I get it, we're trying to ship. This repo has no knowledge of userops - aside from a couple contracts - we have got to move it eventually, at least partially. So if this is the best way forward that's ok with me, let's just be very careful about the duplicated code returning the same address for a given account type.

texuf (approved):

PR #4457, opened 2025-11-04
stop filtering out your own device when encrypting data

texuf:
we actually want bots to encrypt any new keys they create to themselves so that they can download them later if needed


texuf:
I have a test that makes me think this isn't necessary, i'll investigate tomorrow

erikolsson (approved):

PR #4506, opened 2025-11-11
node: High usage monitor 

jontra:
### Description

Implemented abuse detection for event RPCs:

  - Introduced a configurable CallRateMonitor that tracks per-user call rates across sliding windows via ring buffers, plus helper wiring to instantiate it from config.AbuseDetection.
  - Hooked AddEvent, AddMediaEvent, and CreateMediaStream so every successful call records the creator’s address under the appropriate call type
  - Exposed recent abusers on the /status endpoint via a new abuse field, detailing which accounts exceeded which thresholds over 1-minute and 30-minute windows by default.

### Checklist

### DIFF
--- a/core/node/rpc/highusage/monitor.go
+++ b/core/node/rpc/highusage/monitor.go
@@ -67,11 +67,19 @@ type UsageViolation struct {
 	Limit  uint32
 }
 
+// UsageSnapshot captures the current usage for a specific threshold window.
+type UsageSnapshot struct {
+	Window time.Duration
+	Count  uint32
+	Limit  uint32
+}
+
 // HighUsageInfo represents a single offending account for a specific call type.
 type HighUsageInfo struct {
 	User       common.Address
 	CallType   CallType
 	Violations []UsageViolation
+	Usage      []UsageSnapshot
 	LastSeen   time.Time
 }
 
@@ -200,7 +208,7 @@ func (m *inMemoryCallRateMonitor) RecordCall(userBytes []byte, now time.Time, ca
 		return
 	}
 
-	spec := m.callSpecs[int(callType)]
+	spec := m.callSpecs[callType]
 	if spec == nil {
 		return
 	}
@@ -228,18 +236,19 @@ func (m *inMemoryCallRateMonitor) GetHighUsageInfo(now time.Time) []HighUsageInf
 
 	for addr, stats := range m.users {
 		for ct := CallType(0); ct < callTypeCount; ct++ {
-			cs := stats.perType[int(ct)]
+			cs := stats.perType[ct]
 			if cs == nil {
 				continue
 			}
-			violations := cs.violations(now)
-			if len(violations) == 0 {
+			usageSnapshots, violations := cs.snapshot(now)
+			if len(usageSnapshots) == 0 || !hasUsageCounts(usageSnapshots) {
 				continue
 			}
 			result = append(result, HighUsageInfo{
 				User:       addr,
 				CallType:   ct,
 				Violations: violations,
+				Usage:      usageSnapshots,
 				LastSeen:   stats.lastSeen,
 			})
 			for _, v := range violations {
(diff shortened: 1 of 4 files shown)


core/node/rpc/abuse/monitor.go
```diff
@@ -0,0 +1,434 @@
...
+	}
+
+	w.lastSlotTime = aligned
+}
+
+func (w *window) add(now time.Time, delta uint32) {
+	w.advance(now)
+	w.slots[w.head] += delta
+	w.sum += delta
+}
```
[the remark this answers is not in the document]
jontra: this will never happen and should not be dealt with

core/config/config.go
```diff
@@ -53,6 +53,24 @@ func GetDefaultConfig() *Config {
 		Metrics: MetricsConfig{
 			Enabled: true,
 		},
+		AbuseDetection: AbuseDetectionConfig{
+			Enabled:    false,
+			MaxResults: 10,
+			Thresholds: map[string][]AbuseThreshold{
```
sergekh2: I suspect map like this won't be easily settable from env var or yaml.
Maybe just normal field per type? (i.e. why map is required?)

core/node/rpc/abuse/monitor.go
```diff
@@ -0,0 +1,457 @@
...
+
+// WindowViolation captures the counts that exceeded a specific threshold.
+type WindowViolation struct {
+	Window time.Duration
+	Count  uint32
+	Limit  uint32
+}
+
+// AbuserInfo represents a single offending account for a specific call type.
+type AbuserInfo struct {
```
sergekh2: Lets rename through this PR s/Abuse/HighUsage/. I.e. not impling intent.

core/config/config.go
jontra: I think this should work without issues with yaml config
something like this
```
  HighUsageDetection:
    enabled: true
    maxResults: 10
    thresholds:
      event:
        - window: 1m
          count: 100
        - window: 30m
          count: 1000
      media_event:
        - window: 1m
          count: 50
        - window: 30m
          count: 500
      create_media_stream:
        - window: 1m
          count: 10
        - window: 30m
          count: 100
```

The reason I've used map here is so I will be able to code the monitor as generic as possible, without specific references to rpc calls. 
@sergekh2 Not sure exactly what do you have in mind when saying `normal field per type`. is "type" an rpc call?

core/node/rpc/call_rate_monitor.go
```diff
@@ -0,0 +1,55 @@
...
+
+	return highusage.NewCallRateMonitor(highusage.Config{
+		Enabled:    cfg.Enabled,
+		MaxResults: cfg.MaxResults,
+		Thresholds: thresholds,
+		Logger:     logger,
+	})
+}
+
+func (s *Service) recordCallRate(callType highusage.CallType, creator []byte) {
```
sergekh2: why this function at all? always init monitor to real one or noop and call it.

core/node/rpc/highusage/monitor.go
```diff
@@ -0,0 +1,512 @@
...
+//
+// Memory usage: each active user consumes roughly ~2KB per tracked call type
+// (window slots + metadata), so about 4k users equates to ~8–10MB of heap. The
+// cleanup watermark keeps the set bounded while remaining configurable.
+
+// CallType identifies the RPC operation category being tracked.
+type CallType string
+
+const (
+	CallTypeEvent             CallType = "event"
```
sergekh2: lets make these int and use array in place of map: this data structure is updated on every RPC call, and using map is not speed and memory efficient for constant values.

core/node/rpc/highusage/monitor.go
```diff
@@ -0,0 +1,512 @@
...
+import (
+	"sort"
+	"sync"
+	"time"
+
+	"github.com/ethereum/go-ethereum/common"
+	"go.uber.org/zap"
+)
+
+// Package highusage provides a lightweight in-memory CallRateMonitor that tracks how
```
sergekh2: please move this comment above package so it becomes package comment (go convention)

core/node/rpc/highusage/monitor.go
```diff
@@ -0,0 +1,512 @@
...
+			zap.Int("cleaned_count", cleaned),
+			zap.Duration("cleanup_after", m.cleanupAfter),
+			zap.Duration("since_last_cleanup", now.Sub(m.lastCleanup)),
+		)
+	}
+	m.lastCleanup = now
+}
+
+func (m *inMemoryCallRateMonitor) cleanupLoop() {
+	defer m.cleanupWG.Done()
```
sergekh2: create and close ticker here, no need to have it as a field

core/node/rpc/highusage/monitor.go
```diff
@@ -0,0 +1,512 @@
...
+	before := len(m.users)
+	expireBefore := now.Add(-m.cleanupAfter)
+	for addr, stats := range m.users {
+		if stats.lastSeen.Before(expireBefore) {
+			delete(m.users, addr)
+		}
+	}
+	cleaned := before - len(m.users)
+	if cleaned > 0 {
+		m.logger.Info(
```
sergekh2: this will produce uninformative log statement every 30 seconds?

core/node/rpc/highusage/monitor.go
jontra: moved

core/node/rpc/highusage/monitor.go
```diff
@@ -0,0 +1,574 @@
...
+			m.mu.Lock()
+			m.cleanupLocked(now)
+			m.mu.Unlock()
+		case <-ctx.Done():
+			return
+		}
+	}
+}
+
+func (m *inMemoryCallRateMonitor) Close() {
```
sergekh2: now Close(), closeOnce, caleanupWG and cancel can be removed: it exits cleanupLoop on context done, so there is nothing else left to cleanup. 

(we switched to <-ctx.Done() cleanup pattern through the code, earlier code still uses Close, which is more error prone (easy to forget to call), the pattern is to pass in default server context which gets canceled on exit, or test context, which gets canceled on test exit).

sergekh2 (approved):

PR #4524, opened 2025-11-13
Remove StreamParentId from protocol

texuf:
This was a holdover from before we were deriving channel ids from space ids.

IMPORTANT - don't land until https://github.com/towns-protocol/towns/pull/4523 has been deployed to all nodes


erikolsson:
👍  could you just let me know before merging this? didnt have time to update ios yet

PR #4544, opened 2025-11-17
Un-deprecate hasSpaceMembership

texuf:
why was this deprecated?


salzbrenner:
> why was this deprecated?

I guess it doesn't need to be. When we switched to checking expired memberships, I think my thinking was that `hasSpaceMembership` was no longer semantically correct because you can have a space membership nft forever, but it can be expired. And there were also only a couple places using it if iirc. But I guess it's a shortcut to checking the expiration

PR #4545, opened 2025-11-17
sdk: add app address in member metadata stream view

miguel-nascimento:


texuf (approved):
so sorry that you need to do this :)

PR #4546, opened 2025-11-18
fix(client-v2): lru for crypto store & retry makeEvent

miguel-nascimento:
should close TOWNS-31741


packages/encryption/src/CryptoStoreInMemory.ts
```diff
@@ -8,15 +8,26 @@ import {
 import { InboundGroupSessionData } from './encryptionDevice'
 import { UserDevice } from './olmLib'
 import { CryptoStore, DEFAULT_USER_DEVICE_EXPIRATION_TIME_MS } from './cryptoStore'
+import { LRUCache } from 'lru-cache'
+
+const DEFAULT_MAX_CRYPTO_STORE_ENTRIES = 5_000
```
miguel-nascimento: Thoughts? Currently its a freestyle number, need to tweak it

PR #4553, opened 2025-11-18
node: introduce settings to configure the number of events per miniblock

bas-vk:
## Changes

  ### Configuration (`crypto/config.go`)
  - Added two new on-chain configuration keys:
    - `stream.maxEventsPerMiniblock` - Maximum number of events per miniblock
    - `stream.maxTotalEventsSizePerMiniblock` - Maximum combined size (in bytes) of all protobuf-encoded events
  - Defined protocol-level defaults that also serve as hard limits:
    - `MaxEventsPerMiniblockDefault = 15,000` events
    - `StreamMaxTotalEventsSizePerMiniblockDefault = 10 MB`
  - Added corresponding fields to `OnChainSettings` struct

  ### Event Selection (`minipool.go`)
  - Modified `eventHashes()` and `eventHashesAsBytes()` to enforce configured limits
  - Implemented dual-limit logic:
    - Stops at max event count, OR
    - Stops when cumulative event size exceeds max size
  - Events are selected in order until either limit is reached

  ### Proposal & Validation (`stream_view.go`, `stream.go`)
  - Added `MiniblockEventLimits()` helper function that:
    - Reads limits from on-chain settings
    - Falls back to defaults if not configured
    - Enforces protocol max limits even if on-chain values are higher
  - Enhanced `ProposeNextMiniblock()` to reject proposals exceeding limits
  - Added validation in `tryApplyCandidate()` to reject oversized candidates before applying

  ### Testing (`replication_test.go`)
  - Added `TestReplicatedMbProductionWithTooManyEvents` - validates max event count enforcement
  - Added `TestReplicatedMbProductionWithTooBigEvents` - validates max total size enforcement
  - Both tests verify that miniblocks are properly split when limits are exceeded

  ## Behavior

  **Before**: Miniblocks could contain unlimited events from the minipool, potentially creating very large blocks
  that impact network performance and blockchain transaction costs.

  **After**: Miniblocks are capped at:
  - Default: 15,000 events OR 10 MB total size (whichever is reached first)
  - Configurable via on-chain settings within protocol max limits
  - Events exceeding limits remain in the minipool for subsequent miniblocks

  ## Motivation

  - **Network efficiency**: Prevents oversized miniblocks that strain replication and storage
  - **Cost control**: Limits on-chain registration transaction sizes
  - **Configurability**: Operators can tune limits via on-chain settings without code changes
  - **Safety**: Hard-coded maximum limits prevent misconfiguration

  ## Testing

  - ✅ New tests verify both limit types are enforced correctly
  - ✅ Existing miniblock production tests still pass
  - ✅ Tests confirm events properly overflow into subsequent miniblocks

### DIFF
--- a/core/node/crypto/config.go
+++ b/core/node/crypto/config.go
@@ -70,10 +70,25 @@ const (
 	StreamUserInboxStreamHistoryMiniblocksConfigKey    = "stream.historyminiblocks.a1"
 	StreamUserSettingsStreamHistoryMiniblocksConfigKey = "stream.historyminiblocks.a5"
 
+	// StreamMaxEventsPerMiniblockKey is the maximum number of events in a miniblock.
+	StreamMaxEventsPerMiniblockKey = "stream.maxeventsperminiblock"
+
+	// StreamMaxTotalEventsSizePerMiniblockKey is the maximum size (in bytes) of all protobuf encoded events
+	// combined in a miniblock.
+	StreamMaxTotalEventsSizePerMiniblockKey = "stream.maxtotaleventssizeperminiblock"
+
 	// StreamDistributionExtraCandidatesCountCountKey is the key for many extra nodes on top of
 	// replication factor must be picked as candidates to place a stream on. From these candidates
 	// the best replication factor nodes are picked.
 	StreamDistributionExtraCandidatesCountCountKey = "stream.distribution.extracandidatescount"
+
+	// MaxEventsPerMiniblockDefault defines the default (also the maximum) number of events in a miniblock
+	// if not overwritten by StreamMaxEventsPerMiniblockKey.
+	MaxEventsPerMiniblockDefault = 15_000
+
+	// StreamMaxTotalEventsSizePerMiniblockDefault defines the default (also the maximum) size of all protobuf encoded events
+	// combined in a miniblock if not overwritten by StreamMaxTotalEventsSizePerMiniblockKey.
+	StreamMaxTotalEventsSizePerMiniblockDefault = 10 * 1024 * 1024
 )
 
 var (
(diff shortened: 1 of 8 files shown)


core/node/crypto/config.go
```diff
@@ -70,10 +70,25 @@ const (
 	StreamUserInboxStreamHistoryMiniblocksConfigKey    = "stream.historyminiblocks.a1"
 	StreamUserSettingsStreamHistoryMiniblocksConfigKey = "stream.historyminiblocks.a5"
 
+	// StreamMaxEventsPerMiniblockKey is the maximum number of events in a miniblock.
+	StreamMaxEventsPerMiniblockKey = "stream.maxEventsPerMiniblock"
```
sergekh2: lets use lowercase for keys (then it matches the name in solidity since there is toLower (we need to get rid of this toLower))

core/node/crypto/config.go
```diff
@@ -170,6 +185,12 @@ type OnChainSettings struct {
 	// Number of miniblocks to keep for each type of stream.
 	// 0 means keep all miniblocks.
 	StreamHistoryMiniblocks StreamHistoryMiniblocks `mapstructure:",squash"`
+
+	// StreamMaxEventsPerMiniblock is the maximum number of events that in a miniblock.
+	StreamMaxEventsPerMiniblock uint64 `mapstructure:"stream.maxEventsPerMiniblock"`
```
sergekh2: (same)

core/node/events/stream_view.go
```diff
@@ -339,7 +351,7 @@ func (r *StreamView) ProposeNextMiniblock(
 
 	return &ProposeMiniblockResponse{
 		Proposal: &MiniblockProposal{
-			Hashes:            r.minipool.eventHashesAsBytes(),
+			Hashes:            r.minipool.eventHashesAsBytes(cfg),
```
sergekh2: this limit can't be applied to proposal: consider pathological situation when the limit is set to 1, then if 3 nodes have 3 events in different order, they will never produce proposal that can be voted on (i.e. there will be no event which is common to at least 2 proposals).

PR #4554, opened 2025-11-18
App registry: Sync cookie persistence

jontra:
## Summary

Adds cookie persistence for stream synchronization in the App Registry service, allowing the service to resume from the last processed position after restarts, so it won't miss any new events that were sealed in a miniblock while the service was down

## Key Changes

**Cookie Persistence Infrastructure**
- Added `PostgresStreamCookieStore` - a new storage implementation for persisting sync cookies to PostgreSQL
- Added database migration (`000007_create_stream_cookies`) to create the `stream_sync_cookies` table
- Defined `StreamCookieStore` interface in `stream_cookie_store.go` for cookie storage operations

**Smart Cookie Persistence via TrackedStreamView**
- Added `ShouldPersistCookie(ctx) bool` method to the `TrackedStreamView` interface
- Implemented `shouldPersistCookie` callback in `AppRegistryTrackedStreamView` that only persists cookies for **channel streams** that have bot members (checked via `IsForwardableApp`)
- User inbox streams don't need cookie persistence since session keys are already stored in the database and re-processing is idempotent

## Test plan

- [x] Added `pg_stream_cookie_store_test.go` - Tests for the PostgreSQL cookie store (CRUD operations, multiple streams, edge cases)
- [x] Added `tracked_stream_test.go` - Tests for the `shouldPersistCookie` logic with mock `EncryptedMessageQueue`
- [x] All existing tests pass

### DIFF
--- a/core/node/track_streams/pg_stream_cookie_store.go
+++ b/core/node/track_streams/pg_stream_cookie_store.go
@@ -77,6 +77,11 @@ func (s *PostgresStreamCookieStore) PersistSyncCookie(
 	streamID shared.StreamId,
 	cookie *protocol.SyncCookie,
 ) error {
+	if cookie == nil {
+		return base.RiverError(protocol.Err_INVALID_ARGUMENT, "nil cookie").
+			Tag("streamId", streamID)
+	}
+
 	_, err := s.pool.Exec(
 		ctx,
 		`INSERT INTO `+s.tableName+` (stream_id, minipool_gen, prev_miniblock_hash, updated_at)


core/node/track_streams/pg_stream_cookie_store.go
```diff
@@ -0,0 +1,199 @@
...
+// This is a shared implementation that can be used by any service (App Registry, Notifications, etc.)
+// that needs to persist sync cookies for stream resumption.
+//
+// The table schema is:
+//
+//	CREATE TABLE IF NOT EXISTS stream_sync_cookies (
+//	    stream_id            CHAR(64) PRIMARY KEY NOT NULL,
+//	    minipool_gen         BIGINT NOT NULL,
+//	    prev_miniblock_hash  BYTEA NOT NULL,
+//	    updated_at           TIMESTAMP DEFAULT NOW()
```
texuf: Do we need this comment? Seems like it could easily get out of date

PR #4562, opened 2025-11-19
run generate_indexes script

texuf:


miguel-nascimento (approved):
we should run this in "publish to npm" action PR

PR #4567, opened 2025-11-20
Add support to store and read miniblock data to GCS and S3

bas-vk:
## Summary

  Add support for storing and reading miniblock data to external storage (GCS
  and S3). This enables offloading large media stream miniblocks from the
  PostgreSQL database to cloud object storage, improving database performance
  and reducing storage costs.

  ## Key Changes

  ### External Storage Infrastructure
  - **Storage abstraction layer** (`core/node/storage/external/storage.go`):
  Unified interface for GCS and S3 operations
  - **Upload sessions**: Streaming upload support for both GCS and S3 with
  automatic multipart handling
  - **Download operations**: Efficient byte-range downloads for selective
  miniblock retrieval
  - **Storage location tracking**: New `blockdata_ext` column in `es` table to
  track where miniblock data is stored (DB/GCS/S3)

  ### Migration System
  - **Automatic migration**: Ephemeral streams are migrated to external storage
   upon normalization
  - **Existing stream migration**: Optional migration of existing media streams
   from database to external storage
  - **Retry logic**: Up to 5 retry attempts per stream with exponential backoff
  - **Configurable**: Can enable/disable migration through configuration

  ### Database Schema
  - New `miniblocks_ext_m##` tables (partitioned) for tracking external storage
   metadata
  - Added `blockdata_ext` column to `es` table with constraint checking (D=DB,
  G=GCS, S=S3)
  - Migration script `000009_media_miniblock_ext_storage_tables.up.sql`

  ### Observability & Metrics
  - **Migration metrics**: Success/failure counters and bytes migrated
  - **Network metrics**: Upload/download duration histograms
  - **Queue monitoring**: Gauge for pending migration count

  ### Testing
  - Comprehensive test suite for external storage operations
  (`pg_stream_store_external_storage_test.go`, 786 lines)
  - Upload/download validation with byte-level verification
  - Multi-stream concurrent migration tests
  - **Goleak fix**: Added opencensus worker goroutine to ignore list for GCS/S3
   clients

  ## Technical Details

  ### Storage Format
  Miniblocks are concatenated into a single object per stream in external
  storage:
  - Each miniblock's position tracked by start byte offset and size
  - Enables efficient range-based downloads for specific miniblocks
  - Object key format: `{schema_lock_id}/{stream_id}`

  ### Configuration
  ```go
  ExternalMediaStreamStorageConfig {
      Gcs: GcsConfig              // Google Cloud Storage settings
      AwsS3: AwsS3Config          // AWS S3 settings
      EnableMigrationExistingStreams: bool  // Migrate existing streams
  }

  Performance Considerations

  - Streaming uploads avoid memory pressure for large miniblock sets
  - Partial downloads using HTTP range headers
  - Concurrent migration with configurable batch sizes (default 2500 streams)
  - Non-blocking migration runs in background

  Files Changed

  - 18 files modified, 1653 insertions, 53 deletions
  - Core storage layer (pg_stream_store.go, pg_stream_store_external.go)
  - External storage implementation (external/gcs.go, external/s3.go,
  external/storage.go)
  - Monitor system (pg_ephemeral_store_monitor.go)
  - Comprehensive test coverage (pg_stream_store_external_storage_test.go)

  Test Plan

  - Verify external storage upload/download with GCS
  - Verify external storage upload/download with S3
  - Confirm migration of existing streams when enabled
  - Validate metrics reporting for upload/download operations
  - Test retry logic for failed migrations
  - Ensure backward compatibility for streams without external storage
  - Run integration tests with media streams

  Backward Compatibility

  - Fully backward compatible - defaults to database storage (blockdata_ext =
  'D')
  - Existing streams continue working without migration until explicitly
  enabled
  - Storage location checked at runtime for each miniblock retrieval

### DIFF
--- a/core/node/storage/external/gcs.go
+++ b/core/node/storage/external/gcs.go
@@ -36,7 +36,6 @@ func newGcsUploadSession(
 	bucket string,
 	totalMiniblockDataSize uint64,
 	token *oauth2.Token,
-	httpClient *http.Client,
 ) (*gcsUploadSession, error) {
 	var (
 		reqCtx, reqCancel      = context.WithCancel(ctx)
@@ -74,7 +73,7 @@ func newGcsUploadSession(
 		defer close(respChan)
 		defer reqCancel()
 
-		resp, err := httpClient.Do(req)
+		resp, err := http.DefaultClient.Do(req)
 		if err != nil {
 			_ = bodyWriter.CloseWithError(err)
 			respChan <- struct {
@@ -158,7 +157,7 @@ func (g *gcsUploadSession) Finish(ctx context.Context) (
 }
 
 // Abort the pending upload session
-func (g *gcsUploadSession) Abort(ctx context.Context) {
+func (g *gcsUploadSession) Abort() {
 	g.miniblocks = nil
 	g.reqCancel()
 	_ = g.gcStorage.Close()
--- a/core/node/storage/external/storage.go
+++ b/core/node/storage/external/storage.go
@@ -2,6 +2,7 @@ package external
 
 import (
 	"context"
+	"database/sql/driver"
 	"errors"
 	"fmt"
 	"io"
@@ -9,6 +10,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/ethereum/go-ethereum/common"
 	"golang.org/x/sync/errgroup"
 
 	"github.com/towns-protocol/towns/core/config"
@@ -36,14 +38,14 @@ type (
 		Finish(ctx context.Context) ([]MiniblockDescriptor, MiniblockDataStorageLocation, error)
 		// Abort the upload session.
 		// If Finish was called before this is a no-op.
-		Abort(ctx context.Context)
+		Abort()
 	}
 
 	// Storage defines the interface to upload and download miniblock data to/from external storage.
 	Storage interface {
-		// MigrateExistingStreams returns true if streams that have their miniblock data
+		// MigrateStreams returns true if streams that have their miniblock data
 		// stored in the database must be actively migrated to external storage.
-		MigrateExistingStreams() bool
+		MigrateStreams() bool
 
 		// StartUploadSession starts an upload session for miniblock data to external storage.
 		StartUploadSession(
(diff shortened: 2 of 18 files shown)


core/node/storage/migrations/000009_media_miniblock_ext_storage_tables.down.sql
jontra: Is this file empty on purpose?

core/node/storage/external/gcs.go
```diff
@@ -74,7 +73,7 @@ func newGcsUploadSession(
 		defer close(respChan)
 		defer reqCancel()
 
-		resp, err := httpClient.Do(req)
+		resp, err := http.DefaultClient.Do(req)
```
jontra: why have you removed `httpClient`? 
It's useful if we want to customize the client, inject test client, etc.

Also in two other places (s.3go and storage.go)

bas-vk: It was there in the initial implementation that used the aws/gc SDK to upload/download blobs. Those SDK's used a http connection pool with keep-alive that goleak detects and reports after the tests finishes. Later we switched go plain http calls that allows use to close idle connections in the tests making goleak happy. And therefore passing in a custom http client isn't needed anymore.

core/node/storage/pg_stream_store_external.go
```diff
@@ -0,0 +1,391 @@
...
+
+	uploadSession, err := s.externalStorage.StartUploadSession(ctx, streamID, totalMiniblockDataSizeInDB)
+	if err != nil {
+		s.extStorageMigrationFailure.Inc()
+		return true, err
+	}
+	defer uploadSession.Abort() // this is a no-op after finish was called on success
+
+	// fetch miniblocks in batches of pageSize from DB and write them one by one to external storage.
+	pageSize := int64(10)
```
jontra: Why 10? should this be a configuration or at least a const somewhere?

core/node/storage/pg_stream_store_external.go
```diff
@@ -0,0 +1,391 @@
...
+
+// StreamMiniblocksStoredLocation returns the location where miniblock data is stored.
+func (s *PostgresStreamStore) StreamMiniblocksStoredLocation(
+	ctx context.Context,
+	streamID StreamId,
+) (location external.MiniblockDataStorageLocation, err error) {
+	err = s.txRunner(
+		ctx,
+		"StreamMiniblocksStoredLocation",
+		pgx.ReadWrite,
```
jontra: you only need read here

core/node/storage/pg_stream_store_external.go
bas-vk: Many media streams have <=10 miniblocks resulting in a single query. Larger media streams are chunked to reduce memory usage during the upload. There is no need for operators to configure this.

core/node/storage/external/storage.go
```diff
@@ -223,7 +248,6 @@ func (s *storage) StartUploadSession(
 	totalMiniblockDataSize uint64,
 ) (UploadSession, error) {
 	if s.gcs != nil {
```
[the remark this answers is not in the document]
bas-vk: I agree that would be better and I will address that in a follow up PR.

PR #4569, opened 2025-11-20
feat(bots): dx around mention behavior

miguel-nascimento:
Previously, if you wanted to mention a user, you had to do the following

```ts
bot.sendMessage(channelId, 'hi @bob', {
    mentions: [
        {
            userId: bobUserId,
            displayName: 'bob',
            mentionBehavior: { case: undefined, value: undefined },
        },
    ],
})
```

We're now processing and allowing devs to use

```ts
bot.sendMessage(channelId, 'hi @bob', { mentions: [{ userId: bobUserId, displayName: 'bob' }] })
```

If they want, they can also mention atChannels or atRole with 

```ts
mentions: [{ atChannel: true }]
mentions: [{ roleId: 'roleIdHere' }]
```

### DIFF
--- a/packages/bot/src/bot.test.ts
+++ b/packages/bot/src/bot.test.ts
@@ -784,11 +784,19 @@ describe('Bot', { sequential: true }, () => {
         )
         const channelMessage = bobDefaultChannel.timeline.events.value.find(
             (x) => x.eventId === messageId,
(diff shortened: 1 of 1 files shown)


texuf (approved):
approved when you fix build  ;)

PR #4602, opened 2025-11-28
feat: allow bot owners update their bot profile image in stream

miguel-nascimento:
Currently, only you can update your user metadata stream.
However, with bots, we found this to be kinda limiting, since only the bot can update their own profile image (which causes bad ux, bot owner can't update their bots image)

This change adds a new chain auth kind rule `IsBotOwner`, allowing us to check if event is from bot owner and allowing it to be written.

Node changes was mostly AI written. It make sense to me, but I would appreciate a review for someone with more deep knowledge.

### DIFF
--- a/core/node/rules/can_add_event.go
+++ b/core/node/rules/can_add_event.go
@@ -645,7 +645,7 @@ func (params *aeParams) creatorIsMemberOrBotOwner() (*auth.ChainAuthArgs, error)
 	}
 	if isMember {
 		// skip chain auth - already authorized as member
-		return nil, nil 
+		return nil, nil
 	}
 
 	// not member - check if this is a bot user's stream


core/node/rules/can_add_event.go
```diff
@@ -634,6 +640,62 @@ func (params *aeParams) creatorIsMember() (bool, error) {
...
+func (params *aeParams) getCreatorIsMember() (bool, error) {
+	if params.creatorMembershipCached {
+		return params.creatorIsMemberResult, params.creatorMembershipErr
+	}
+	params.creatorMembershipCached = true
+	params.creatorIsMemberResult, params.creatorMembershipErr = params.streamView.IsMember(
+		params.parsedEvent.Event.CreatorAddress,
+	)
+	return params.creatorIsMemberResult, params.creatorMembershipErr
+}
```
texuf: please remove the caching, just use botOnwerChainAuthIfNotMember - please rename to what i suggested earlier.

PR #4621, opened 2025-12-03
Registry restart sync fixes

jontra:
### Description

When the app registry or notification service restarts:
1. It starts syncing streams from scratch (requesting a reset)
2. The server returns a reset response starting from the **last snapshot**
3. The service starts handling events from the minipool, skipping all miniblocks (assuming they are already processed)
4. Events between the last processed miniblock and the current last miniblock are **not handled**

This is problematic for bots that need to receive all messages in channels they're members of.

## Solution

Persist the last `minipoolGen` (last processed miniblock + 1). On restart, always request a fresh sync (which triggers a reset), then detect gaps by comparing the persisted position with the server's snapshot position. If there's a gap, fetch missing miniblocks via `GetMiniblocks` RPC.
Send notifications for all events from the persisted minipoolGen to the last miniblock + all events in the minipool.

### Checklist

### DIFF
--- a/core/node/track_streams/multi_sync_runner.go
+++ b/core/node/track_streams/multi_sync_runner.go
@@ -204,7 +204,7 @@ func (ssr *syncSessionRunner) handleHistoricalContent(
 		return
 	}
 
-	// FromMiniblockNum is before our fist miniblock num, we need to fetch miniblocks from the remote
+	// FromMiniblockNum is before our first miniblock num, we need to fetch miniblocks from the remote
 	if fromMiniblockNum < serverSnapshotMb {
 		log.Infow("handleHistoricalContent: fetching missing miniblocks",
 			"streamId", streamId,
@@ -294,7 +294,7 @@ func (ssr *syncSessionRunner) maybePersistCookie(
 		return
 	}
 
-	if err := ssr.cookieStore.PersistSyncCookie(ssr.rootCtx, streamId, cookie); err != nil {
+	if err := ssr.cookieStore.WriteSyncCookie(ssr.rootCtx, streamId, cookie); err != nil {
 		logging.FromCtx(ssr.syncCtx).Errorw("Failed to persist sync cookie", "streamId", streamId, "error", err)
 		return
 	}
--- a/core/node/track_streams/sync_helpers.go
+++ b/core/node/track_streams/sync_helpers.go
@@ -89,6 +89,8 @@ func fetchMiniblocks(
 }
 
 // fetchMiniblocksFromNode fetches miniblocks from a specific node.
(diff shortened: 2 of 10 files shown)


core/node/track_streams/sync_helpers.go
```diff
@@ -0,0 +1,66 @@
...
+
+	req := connect.NewRequest(&protocol.GetMiniblocksRequest{
+		StreamId:      streamId[:],
+		FromInclusive: fromInclusive,
+		ToExclusive:   toExclusive,
+	})
+	req.Header().Set(headers.RiverNoForwardHeader, headers.RiverHeaderTrueValue)
+
+	resp, err := client.GetMiniblocks(ctx, req)
+	if err != nil {
```
jontra: If this fails, we can try to fetch from replicated nodes

core/node/track_streams/multi_sync_runner.go
```diff
@@ -176,6 +180,156 @@ func (ssr *syncSessionRunner) AddStream(
...
+	serverSnapshotMb, err := getFirstMiniblockNumber(miniblocks)
+	if err != nil {
+		// Log but continue - don't block sync for parse errors
+		log.Errorw("handleHistoricalContent: Failed to parse first miniblock", "error", err, "streamId", streamId)
+		return
+	}
+
+	// FromMiniblockNum is before our fist miniblock num, we need to fetch miniblocks from the remote
+	if fromMiniblockNum < serverSnapshotMb {
+		log.Warnw("handleHistoricalContent: fetching missing miniblocks",
```
sergekh2: Should this be a warning? Seems like expected situation?

core/node/track_streams/sync_helpers.go
```diff
@@ -0,0 +1,123 @@
...
+		return nil, base.AsRiverError(err).
+			Message("Failed to fetch miniblocks for gap recovery").
+			Tag("streamId", streamId).
+			Tag("nodeAddr", nodeAddr).
+			Tag("fromInclusive", fromInclusive).
+			Tag("toExclusive", toExclusive)
+	}
+
+	return resp.Msg.GetMiniblocks(), nil
```
bas-vk: `client.GetMiniblocks` has a server side limit (on-chain setting) on the number of miniblocks it can return. If the request range of miniblocks is larger than this setting the server only returns `limit` number of miniblocks, so `response.Miniblocks` can be truncated. In that case the client must fetch the next batch of miniblocks until it received them all.

jontra: good catch
fixed

bas-vk (approved):

PR #4659, opened 2025-12-08
Initial stream validation when adding to sync

begmaroman:
Make sure stream exists in cache before adding it to sync


core/node/rpc/syncv3/handler/handler_test.go
```diff
@@ -82,20 +82,75 @@ func TestSyncStreamHandlerImpl_SyncID(t *testing.T) {
 func TestSyncStreamHandlerImpl_Modify(t *testing.T) {
 	t.Run("validation failure", func(t *testing.T) {
 		env := newHandlerTestEnv(t)
-		_, err := env.handler.Modify(&protocol.ModifySyncRequest{})
+		_, err := env.handler.Modify(context.Background(), &protocol.ModifySyncRequest{})
```
sergekh2: here and everywhere: in test use t.Context() which gets properly canceled when test is over

sergekh2 (approved):

PR #4670, opened 2025-12-09
fix: use wallet link in is bot owner auth check

miguel-nascimento:
Previously it was only checking the user wallet.
This PR changes it so it checks the linked wallets instead.

Should close TOWNS-31860


miguel-nascimento:
Tested locally and it works

PR #4673, opened 2025-12-10
[TOWNS-31826] Track messages sent on mixpanel

jontra:
### Description

We want to know how many messages bot are sending vs total messages by users.
- Added mixpanel sdk
- injected mixpanel client to app registry
- track messages as they arrive

### Checklist

### DIFF
--- a/core/node/infra/analytics/analytics.go
+++ b/core/node/infra/analytics/analytics.go
@@ -2,12 +2,15 @@ package analytics
 
 import (
 	"context"
+	"time"
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/mixpanel/mixpanel-go"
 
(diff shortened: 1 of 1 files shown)


core/node/app_registry/event_processor.go
```diff
@@ -158,6 +162,11 @@ func (p *MessageToAppProcessor) OnMessageEvent(
 		return
 	}
 
+	// Track message posted event
+	p.analytics.Track(ctx, creator.Hex(), "posted message", map[string]any{
```
bas-vk: Instead of passing hex why not pass the address and let the `Track` function handle the formatting? That ensures it is consistent everywhere bacause addresses in hex format can be lower, upper and a mix of lower/upper case if it's checksummed.

